PR #721, opened 2017-08-09
Add a loading screen for browser and desktop IDE

### LINKED ISSUE #688: As a user I want a loading screen so that I understand that IDE is starting
### Rationale

XOD IDE is heavy and may take up to dozen seconds (for the first visit) to load and start. Currently, all a user sees is a blank white screen. It distracts UX.

We should make some starting screen that shows up immediately so that all other preparations would be done behind it.

### Acceptance criteria

- [x] Once I launch browser IDE with browser cache cleared, I see full width and height Loading… screen
- [x] When loading is done the screen switches to the IDE interface instantly

### Out of scope

- Progress bar
- Desktop IDE screen

Although, they would be nice if could be implemented trivially.

evgenykochetkov:
Closes #688


package.json
```diff
@@ -7,7 +7,7 @@
     "bootstrap": "lerna bootstrap",
     "build": "node ./tools/lerna-run-sorted build",
     "clean": "yarn run clean:dist && yarn run clean:node_modules",
-    "clean:dist": "rimraf packages/*/dist",
+    "clean:dist": "rimraf packages/*/dist packages/xod-cli/bin packages/xod-client-electron/src-babel",
```
nkrkv: Oops :+1:

packages/xod-client-browser/webpack.config.js
```diff
@@ -89,7 +89,7 @@ module.exports = {
   plugins: [
     new webpack.NoErrorsPlugin(),
     new CopyWebpackPlugin([
-      { from: pkgpath('src/index.html') },
+      { from: pkgpath('../xod-client/src/core/assets/index.html') },
```
nkrkv: The `..` could lead to bad things if we’ll reorganize packages some day. Better use:

```javascript
pkgpath('node_modules/xod-client/src/core/assets/index.html')
```

packages/xod-client-electron/webpack.config.js
```diff
@@ -103,7 +103,7 @@ const options = {
   plugins: [
     new webpack.NoErrorsPlugin(),
     new CopyWebpackPlugin([
-      { from: pkgpath('src/index.html') },
+      { from: pkgpath('../xod-client/src/core/assets/index.html') },
```
nkrkv: Ditto

packages/xod-client-browser/webpack.config.js
evgenykochetkov: done 👍🏻

PR #728, opened 2017-08-16
Introduce hm-def type hints

### LINKED ISSUE #716: Optimize type checking
### Rationale

Now `hm-def` checking is very slow. It is because it recursively validates any object shape on each call. We can make a shortcut to get an order of magnitude boost. The idea is to check objects’ constructor rather than their shape. See https://gist.github.com/nkrkv/b7779ead228c1d98334356edaa2f8bb7 for the idea.

**xod-project**

- [ ] Project
- [ ] Patch
- [ ] Node
- [ ] Link

A quick experiment on converting `Project` to the new scheme showed a 2× gain in test performance. Converting these four entities validation should give even better results (hope 10×)

nkrkv:
Closes #716 :tada: 

Test speed up 4×. Overall CI gain is 2×.

Also:

- Make `xod-fs` tests more managable
- Add end-to-end `xod-fs` test
- Use a JSON serializer with stable key sorting to avoid false diffs on FS
- Make `xod-project` a bit more DRY


packages/xod-func-tools/src/types.js
```diff
@@ -88,13 +88,22 @@ export const EnumType = R.curry(
...
+
 // Model :: String -> String -> String -> StrMap Type -> Type
 export const Model = R.curry(
   (packageName, docUrl, typeName, schema) => NullaryType(
     packageName,
     docUrl,
     typeName,
-    hasType($.RecordType(schema))
+    R.ifElse(
+      hasTypeHint,
```
evgenykochetkov: So, if we're in default(non-strict) mode, and a structure doesn't have a type hint, we just silently fall back to the slow check? Maybe, we should give some kind of warning in this case?

nkrkv: This is by-design and totally OK, I think. Small structures like `Position` or ad-hoc `TPatch` and friends don’t worth check hinting.

PR #742, opened 2017-08-23
Switch from slots to grid

### LINKED ISSUE #635: Desktop IDE has blurred fonts on Windows
### Steps to reproduce

- Compare desktop and browser IDE side by side

### Expected behavior

Fonts are crisp in both versions

### Actual behavior

Fonts in desktop version are blurred

xod-scrn

### LINKED ISSUE #730: Switch from slots to a grid
### Rationale

Current layout with the big slots is compelling but impractical when one needs to place many single-pin nodes like terminals or constants. It leads to pointless fans and terminal nodes’ puzzles.

fan-problem

However, the slots scheme is liked by many. We’re going to implement a compromise solution where things are still aligned to a grid, the grid is sparse enough, which at the same time lets arrange single-pin nodes effectively.

Another flaw in the current scheme is wrong accents in the nodes. Pin labels are more visible than a node label, links are drawn over labels and nodes.

We’re going to do this:

Effectively we’re going to:

- Change the slot size to somewhat smaller
- Change slot background pattern to a grid pattern
- Adjust the rendering of nodes and its parts
- Change layer ordering

Few tweaks to the mockup shown are necessary:

- Make the grid exactly two times less granular in X-direction so that distance between neighbor pins would be a single cell
- Make the grid exactly three times less granular in Y-direction so that a single node fits a single cell vertically
- Always place pins on grid lines intersection. This would require decentering pin rows at ½ of grid cell to the left in cases when one row (e.g. inputs) has an odd number of pins and another (e.g. outputs) has an even number of pins.
- Do not draw any text for nodes other than its label and pin labels

### Acceptance criteria

- [ ] The patch board looks like in mock-up provided
- [ ] Pins are always placed on grid line intersections regardless of number of pins
- [ ] A single-pin node size is exactly one cell of the grid (although ½ offset by X to hit grid lines intersection)
- [ ] A zero-pin node is rendered the same way as a single-pin node does
- [ ] Links are lowered behind nodes, pins, and labels
- [ ] Hovered and selected links are drawn in front, i.e. above nodes, pins, and labels
- [ ] Pin labels have a semi-transparent background to contrast with links
- [ ] Nodes have a semi-transparent background so that a link behind it is slightly visible

### Out of scope

- Terminal nodes, they’re in the separate issue #411

evgenykochetkov:
Closes #730
Closes #635


nkrkv:
A little :bug: 

### Steps to reproduce

1. Add a comment
2. Hold the resize handle and slowly drag down to increase the height

[comment-bug.mp4.zip](https://github.com/xodio/xod/files/1245773/comment-bug.mp4.zip)

### Expected behavior

- The cursor is double arrow for the whole duration of the resize process
- The blue shadow is rounded up (i.e. the height projection triples) once I cross an invisible line which is drawn in the middle of DMZ

### Actual behavior

- The cursor blinks
- The height projection is rounded up only after I resize over triple of the initial height

evgenykochetkov:
@nkrkv done and added a couple of other tweaks

packages/xod-client-electron/src/app/main.js
```diff
@@ -53,6 +53,8 @@ function createWindow() {
     minHeight: 600,
     title: 'XOD IDE',
     show: false,
+    // this is required for subpixel antialiasing to work
+    backgroundColor: '#FFF',
```
nkrkv: In the PR description, please, refer to #635 (we’re going to close it) so that we’d have audit trail

PR #744, opened 2017-08-24
Add suggester with quick search for nodes to insert them easily

### LINKED ISSUE #235: As a xoder I want to quick-search for nodes available in libs so that I can choose node type to insert fast
Our current combo box with all node types should be replaced with a list view dialog to be more accessible.

Each list item should contain the title (label), node type with full path and first paragraph of the description.

The dialog should provide a search box to quickly filter out node types by search term specified. An item should be included in the result if it _matches_ each _word_ of the search query.

An item matches a word if it contains it as a substring somewhere among title, full description, fully-qualified type.
### Acceptance criteria
- [ ] I can choose “Insert node” item in the menu or press a shortcut (i) and would be presented with a dialog box with a search input
- [ ] I can type a search query and search results get refreshed
- [ ] The results list is sorted by relevance
- [ ] Visible matches are highlighted in bold
- [ ] I can select a list item and press enter or double click to insert the node
- [ ] I can select node with arrow keys
### Out of scope
- Search among not connected libraries

### LINKED ISSUE #740: Add help in quick search popup
### Rationale

A help article shown next to a node selected in quick search results would provide richer UX while discovering XOD nodes.

### Prerequisites

#34 

### Acceptance criteria

- [ ] Once I change selected item during quick search the help pane of the Suggester updates to show an article about the node selected
- [ ] The help article is rendered exactly the same way as in #34

### LINKED ISSUE #741: As a xoder I want to refine my quick search with a library filter so that I can discover a single library
### Rationale

A common case is to discover all nodes in a particular library. To assist that and several another cases we can support special query syntax like "lib:xod/core".

Then we can provide search shortcuts for ready-to-use queries that filter by all libraries available out of the box.

### Prerequisites

#235 

### Acceptance criteria

- [ ] On query "lib:xod/core" I see all nodes from xod/core sorted alphabetically
- [ ] On query "lib:xod/patch-nodes input-" I see only input terminal nodes from the patch-nodes lib
- [ ] Once I open quick search I’m presented with shortcuts for ready-to-use queries that filter by all libraries available out of the box.

### Out of scope

Fuzzy search on partial lib specifications like "lib:xod" or "lib:xod/co". They may lead to empty search results. They *could* match xod/core, xod/common-hardware if search engine makes it simple.

### LINKED ISSUE #759: Typing `h` in an input toggles Helpbar
### Steps to reproduce

1. Open any patch
2. In the description field type “hula hoop”

### Expected behavior

Just “hula hoop” gets entered, the Helpbar does not appear.

### Actual behavior

- The help bar toggles on every `h` press.
- The same occurs if I edit a comment
- The same occurs if I bind a value

brusherru:
It will close #235, close #740, close #741 and fix #759.

Actually, it's ready, but we discussed that I have to refactor tests and improve some cases, like this one: search for "ult" should find "xod/common-hardware/hc-sr04-ultrasonic-range" at the first position, but it's on second, on the first one — "xod/core/multiply". I hope I can solve it by changing weights, but I failed to do quickly.

In any case, you can already see pull request and give me a bunch of comments :)


packages/xod-patch-search/test/fixtures/indexData.json
```diff
@@ -0,0 +1,107 @@
...
+    "fullDescription": ""
+  },
+  {
+    "path": "xod/utils/m-to-cm",
+    "keywords": ["m", "cm"],
+    "lib": "xod/utils",
+    "description": "Maps meters to centimeters",
+    "fullDescription": ""
+  }
+]
```
nkrkv: This fixture will be hard to maintain. It’s better to generate the index just in time from the real data in `./workspace`

packages/xod-client-browser/src/containers/App.jsx
```diff
@@ -174,6 +174,7 @@ class App extends client.App {
           onClick(items.redo, this.props.actions.redoCurrentPatch),
           items.separator,
           onClick(items.insertComment, this.props.actions.addComment),
+          onClick(items.insertNode, this.props.actions.showSuggester),
```
nkrkv: I suggest placing “Insert Node” above “Insert Comment”

packages/xod-client-electron/src/view/containers/App.jsx
```diff
@@ -336,6 +336,7 @@ class App extends client.App {
           items.paste,
           items.separator,
           onClick(items.insertComment, this.props.actions.addComment),
+          onClick(items.insertNode, this.props.actions.showSuggester),
```
nkrkv: The same here… And it looks like duplication. Maybe it’s a good idea to implement a menu items factory somewhere in `xod-client`? Something (roughly) like:

`createEditMenu :: [Action] -> Submenu`

packages/xod-client/src/editor/components/Suggester.jsx
```diff
@@ -0,0 +1,184 @@
...
+  }
+
+  getSuggestions(value) {
+    const { index } = this.props;
+    const inputValue = value.trim().toLowerCase();
+
+    if (inputValue.length === 0) { return []; }
+
+    const result = index.search(inputValue);
+    const filtered = result.filter(res => res.score < 0.2);
```
nkrkv: Isn’t `0.2` is a view-level override of the index `threshold` option?

In any case, I’d abstract out the `search` function and place it in `xod-patch-search`. So that if we’d like to hack or even change the search engine, its consumers will not be affected.

packages/xod-client/src/editor/components/Suggester.jsx
```diff
@@ -0,0 +1,184 @@
...
+  }
+
+  storeInputReference(autosuggest) {
+    if (autosuggest !== null) {
+      this.input = autosuggest.input;
+    }
+  }
+
+  // eslint-disable-next-line class-methods-use-this
+  renderItem({ item }, { isHighlighted }) {
```
nkrkv: I guess you can really make the method static and not suppress ESLint

packages/xod-client/src/editor/components/Suggester.jsx
```diff
@@ -0,0 +1,184 @@
...
+          className="description"
+          searchWords={[this.state.value]}
+          textToHighlight={item.description}
+        />
+      </div>
+    );
+  }
+
+  // eslint-disable-next-line class-methods-use-this
+  renderSuggestionsContainer({ containerProps, children }) {
```
nkrkv: Ditto

packages/xod-client/src/editor/selectors.js
```diff
@@ -77,3 +77,8 @@ export const getCurrentPatchOffset = createSelector(
   [getCurrentPatchPath, getTabs],
   (currentPatchPath, tabs) => R.path([currentPatchPath, 'offset'], tabs)
 );
+
+export const getSuggesterVisibility = R.pipe(
```
nkrkv: Better name `isSuggesterVisible`?

nkrkv:
You’ve changed `package.json` but forgot to update `yarn.lock`

nkrkv:
:bug:

### Steps to reproduce

1. Double click a slot
2. Enter a search query which contains “i” letter, e.g. `digital`
3. Press Enter to place the node

### Expected behavior

A new node is placed in the slot double-clicked

### Actual behavior

The node is created in default { 50, 50 }

nkrkv:
A little design issue. In the Suggester I expect the cursor to always have default arrow shape when it is over an item. Now it is either arrow or I-Beam. I see no point to allude to select text portions in suggested items.

packages/xod-client/src/editor/containers/Editor.jsx
```diff
@@ -47,26 +63,45 @@ class Editor extends React.Component {
       [COMMAND.SET_MODE_DEFAULT]: this.setModeDefault,
       [COMMAND.UNDO]: () => this.props.actions.undo(this.props.currentPatchPath),
       [COMMAND.REDO]: () => this.props.actions.redo(this.props.currentPatchPath),
+      [COMMAND.INSERT_NODE]: () => this.showSuggester(),
     };
   }
 
+  showSuggester(pos) {
+    this.props.actions.showSuggester(pos);
```
evgenykochetkov: A little bug: the suggester will be shown while editing a comment node and typing `i`.
I suggest checking here if event target is input or textarea.

evgenykochetkov: Actually, this reproduces with all kinds of inputs, not only comment editing.

nkrkv: :open_mouth: nice catch!

nkrkv:
One more thing. You’ve introduced a new package. Don’t forget to add its `node_modules` directory as a cache target in `.travis.yml`

brusherru:
@nkrkv @evgenykochetkov it's done!
There is left one issue with mouse/keyboard highlighting, but I can't find a quick solution and without crazy kludges. :(

Check it out, please. And then I'll squash commits.

packages/xod-client/src/editor/containers/Editor.jsx
```diff
@@ -47,26 +63,48 @@ class Editor extends React.Component {
       [COMMAND.SET_MODE_DEFAULT]: this.setModeDefault,
       [COMMAND.UNDO]: () => this.props.actions.undo(this.props.currentPatchPath),
       [COMMAND.REDO]: () => this.props.actions.redo(this.props.currentPatchPath),
+      [COMMAND.INSERT_NODE]: (event) => {
+        if (R.contains(event.target.type, ['textarea', 'input'])) return;
```
evgenykochetkov: Turns out we already had an util for this https://github.com/xodio/xod/blob/master/packages/xod-client/src/utils/browser.js#L19

brusherru: Thanks! I'll replace this one and couple of lines in the `Patch.jsx` :)

packages/xod-client/src/editor/components/SuggesterContainer.jsx
```diff
@@ -0,0 +1,118 @@
...
+            this.props.disableBlocker();
+          }
+
+          this.moveTimes += 1;
+          this.moveTimeout = setTimeout(() => {
+            this.moveTimes = 0;
+          }, 50);
+        }}
+      />
+    ) : null;
```
nkrkv: This is a trick. Its intent is not obvious. So please leave a detailed comment describing the rationale and implementation.

packages/xod-client/src/editor/components/SuggesterContainer.jsx
```diff
@@ -0,0 +1,118 @@
...
+        const isOutsideUp = (top - height <= scrollPos);
+        const isOutsideDown = (top + height >= (scrollPos + containerHeight) - height);
+
+        if (isOutsideDown) {
+          return (top + height + height) - containerHeight;
+        } else if (isOutsideUp) {
+          return top - height;
+        }
+
+        // If its inside container — don't do anything at all.
```
nkrkv: “If it’s inside container do nothing at all” :)

packages/xod-patch-search/package.json
```diff
@@ -0,0 +1,30 @@
+{
+  "name": "xod-patch-search",
+  "version": "0.0.1",
```
nkrkv: I think it should be `0.12.1`

packages/xod-patch-search/test/index.spec.js
```diff
@@ -0,0 +1,89 @@
...
+    const result = idx.search('path:number');
+    assert.equal(
+      result[0].item.path,
+      'xod/core/format-number'
+    );
+  });
+
+  it('searches by lib correctly', () => {
+    const result = idx.search('lib:xod/core');
+    assert.lengthOf(result, 63);
```
nkrkv: Very fragile assertion. xod/core constantly grows. I think we should count patch count programmatically. The readability would degrade, but the test will be solid.

packages/xod-patch-search/test/index.spec.js
```diff
@@ -0,0 +1,89 @@
...
+  it('searches: "lib:xod/patch-nodes number"', () => {
+    assert.equal(
+      idx.search('lib:xod/patch-nodes number')[0].item.path,
+      'xod/patch-nodes/input-number' // The same as above + filtered by lib
+    );
+  });
+
+  it('searches: "meter"', () => {
+    const results = idx.search('meter');
+    assert.lengthOf(results, 12);
```
nkrkv: Fragile. Better remove.

packages/xod-patch-search/test/index.spec.js
```diff
@@ -0,0 +1,89 @@
...
+  const getProjectPath = projectName => path.resolve(workspace, projectName);
+
+  before(async () => {
+    const project = await loadProject(getProjectPath('welcome-to-xod'));
+    const indexData = createIndexData(listPatches(project));
+    idx = createIndex(indexData);
+  });
+
+  it('searches by path correctly', () => {
+    const result = idx.search('path:number');
```
nkrkv: Wow. If that works, should we really have `lib:` syntax? Maybe just `path:` is enough?

brusherru: Actually, I'm not sure that is working really correct. But for our current dataset, it returns a correct result. I think that will be good to extract tokens like this from a query and filter data using our algorithms.
But for now, it's working good enough. So I propose to do it later when it will be really needed.

brusherru:
@nkrkv fixed! Ready to rebase and squash. Will you approve?

nkrkv:
Flaw

### Steps to reproduce

1. Press `i` to invoke quick search
2. Type `add`
3. Press `Enter`

### Expected behavior

- The `add` node is placed
- The suggester closes

### Actual behavior

- Nothing changes on `Enter`

### How to fix

Select the first item in the search results. Make sure it works OK if no results found.

nkrkv:
:bug:

### Steps to reproduce

- Open the `welcome-to-xod` project
- Press `i`
- Type `add`
- Press `Down` and `Enter` to place the node
- Try to move the new node with mouse drag

### Expected behavior

The node is moving

### Actual behavior

Nothing

### Console output

Lots of repeating

```
Error: <rect> attribute x: Expected length, "NaN".
Error: <rect> attribute y: Expected length, "NaN".
Error: <svg> attribute x: Expected length, "NaN".
Error: <svg> attribute y: Expected length, "NaN".
```

nkrkv:
:bug: 

### Steps to reproduce

1. Press `i` to invoke quick search
2. Type `a` to get a long list of matches
3. Click an item somewhere in the middle of the list with the mouse to select it
4. Move mouse cursor out of the suggester

### Expected behavior

The selection is kept

### Actual behavior

The selection is dropped

nkrkv:
One more flaw. Hover and active states of searched items are indistinguishable. Better use different styles for these two states like the Project Browser does. A good idea is even to reuse the same colors.

brusherru:
@nkrkv fixed 😬

nkrkv:
The “Insert Node” menu item should show “I” as a shortcut for it. Like “Toggle Helpbar” does.

brusherru:
Also added interaction with help bar. So it closes #740 too.

brusherru:
And it already filters data set by `lib:xod/core` (#741).
But I'm not sure how correct it will be when we have some libs with similar names.
@nkrkv Should I close it too?

nkrkv:
:bug: 

### Steps to reproduce

1. Invoke quick search
2. Type `s`
3. Click `xod/core/select`

### Expected behavior

`xod/core/select` is selected and purple

### Actual behavior

The selection stays on the first item (`xod/common-hardware/servo`)

brusherru:
In the conversation, we're decided to remove hotkeys without modifiers from electron menu ("i" and "h"). So I did it and this PR also fixes #759.
Another way to keep hotkeys available and prevent triggering it while typing text in inputs — disabling menu items. Example: [webtorrent-desktop electron app](https://github.com/webtorrent/webtorrent-desktop/blob/4a3ca5459da995c15af36952c07701479e69c472/main/menu.js)

In the rest all fixed.

nkrkv:
:bug: 

### Steps to reproduce

1. Open Helpbar
1. Invoke quick search
1. Press `sh`

### Expected behavior

Help for the selected Sharp infrared range meter is shown

### Actual behavior

Help for `servo` is shown

brusherru:
I hope this is the last one 🙈 
@nkrkv check it out, please

PR #754, opened 2017-08-29
Implement nodes for ST IMUs

nkrkv:
This is out of band PR that mostly a proof that XOD could be used to interact with complex hardware using XOD patches themselves without relying much on native implementations. The targets are accelerometers and gyros based on STMicroelectronic chips.

As a consequence, we have new nodes:

- `xod/core/discretize-*` — a useful helper (split by arity until we have lists)
- `xod/core/nth-number-*` — yet another helper
- `xod/core/i2c-*` — low-level I²C bus interaction
- `xod/common-hardware/st-imu-*` — internal utilities for ST IMUs
- `xod/common-hardware/{lis,l3g}*` — concrete ST IMU drivers


workspace/lib/xod/core/i2c-read/arduino.cpp
```diff
@@ -0,0 +1,25 @@
...
+{{ GENERATED_CODE }}
+
+void evaluate(Context ctx) {
+    if (!isInputDirty<input_READ>(ctx))
+        return;
+
+    auto usLimit = micros() + getValue<input_TOUT>(ctx) * 1e6;
+    while (!Wire.available()) {
+        // TODO: we should not block here,
+        // but setTimeout(ctx, 0) to try again later
```
brusherru: Are there any difficulties to complete "TODO" now?

workspace/lib/xod/core/word-to-number/patch.xodp
```diff
@@ -0,0 +1,38 @@
+{
```
brusherru: Description is missed.

workspace/lib/xod/core/i2c-read/arduino.cpp
nkrkv: Not so many.

The main one is a lack of research. It *could* be a problem if we don’t consume incoming I²C data quickly because of possible hardware overflow. Maybe it is not. The sensors I used for testing actually send the data right away and the timeout handling is not required for them at all.

A research on a module that is actually unstable/slow is required. I have no such right at this moment.

workspace/lib/xod/core/word-to-number/patch.xodp
nkrkv: Done

PR #781, opened 2017-09-13
Cloud compilation in the electron client

brusherru:
There is no issue 😬


packages/xod-client-electron/src/app/arduinoActions.js
```diff
@@ -213,7 +204,100 @@ export const uploadToArduinoHandler = (event, payload) => {
...
+      error: err,
+      message,
+      percentage,
+    })
+  );
+  const sendSuccess = send('success');
+  const sendProgress = send('progress');
+  const sendFailure = send('failure');
+  const convertAndSendError = err => R.compose(
+    msg => sendFailure(msg, 0, errorToPlainObject(err))(),
```
evgenykochetkov: Do we really need to reset progress bar to 0 on error?
Maybe, it's useful to show user at which stage where an error has happened?

packages/xod-client-electron/src/shared/errorFormatter.js
```diff
@@ -1,5 +1,6 @@
 import R from 'ramda';
 import { ERROR_CODES as XFS_EC } from 'xod-fs';
+import { COMPILATION_ERRORS as XD_EC } from 'xod-deploy/dist/constants';
```
evgenykochetkov: Probably, a comment like "we import it like this because we need only the error codes and we don't mess with transpiling async/await" would be nice here :)

packages/xod-deploy/src/cloudCompiler.js
```diff
@@ -0,0 +1,132 @@
...
+        R.T,
+        R.always(null),
+      ],
+    ])(data.type);
+
+    if (errCode) {
+      reject(createCompileError(data.body, errCode));
+    }
+
+    if (data.type === COMPILATION_RESPONSE_TYPES.SUCCEEDED) {
```
evgenykochetkov: Could there be a situation when `errCode` is falsy and `type  !== COMPILATION_RESPONSE_TYPES.SUCCEEDED `?

packages/xod-client-electron/src/shared/errorFormatter.js
brusherru: Good idea 👍

packages/xod-deploy/src/toolManager.js
```diff
@@ -0,0 +1,132 @@
...
+import { tapP } from 'xod-func-tools';
+
+// =============================================================================
+//
+// Utils
+//
+// =============================================================================
+
+// :: Path -> Boolean
+const isDir = (dir) => {
```
evgenykochetkov: are all these moved from `xod-arduino-deploy`?

packages/xod-deploy/src/uploader.js
```diff
@@ -0,0 +1,123 @@
...
+import cpp from 'child-process-promise';
+import fetch from 'node-fetch';
+
+import { flushSerialBuffer, touchPort1200, waitNewPort, listPorts } from './serialport';
+
+// UploadConfig :: { exe: StrMap String, tool: String, toolUrl: StrMap Url, cmdTemplate: String }
+// LocalConfig :: { toolPath: Path, artifactPath: Path, port: String }
+
+// :: Number -> (* -> Promise a b) -> Promise a b
+const delayPromisedFn = (delayMs, fn) => new Promise(
```
evgenykochetkov: alternative solution:
```
const delay = ms => new Promise(resolve => setTimeout(resolve, ms));

delay(500).then(fn())
```

packages/xod-deploy/src/uploader.js
```diff
@@ -0,0 +1,123 @@
...
+const rejectFetchResult = (res, payload) => Promise.reject(
+  Object.assign(new Error(res.statusText), {
+    status: res.status,
+    statusText: res.statusText,
+  }, payload)
+);
+
+// :: String -> Promise UploadConfig Error
+export const getUploadConfig = boardId =>
+  fetch(`https://compile.xod.show/upload/${boardId}`)
```
evgenykochetkov: another hardcoded url

packages/xod-deploy/src/toolManager.js
brusherru: Yep.
I didn't reuse this package, cause `xod-arduino-deploy` should be removed soon.

packages/xod-deploy/src/cloudCompiler.js
brusherru: Yes. Added a comment about it.

PR #783, opened 2017-09-15
Place nodes by dragging from project browser, bulk move/delete nodes and comments

### LINKED ISSUE #364: As a xoder I want to place nodes by dragging them from the project browser to the board so that I get smoother UX
See #236 for reference.

### Prerequisites

- #361
- #363

### Acceptance criteria

- [ ] I can drag an item from the project browser to the board and get the same UX as if I’d drag an existing node

### LINKED ISSUE #419: As a xoder I want perform bulk move/delete so that I can craft patches faster
### Rationale

Now only a single node or link could be selected. There are many cases when it is required to move or delete a whole cluster of nodes.

If we’ll allow selecting multiple entities it would be a big booster

- [ ] I can click an entity with `Ctrl` holden down to add it to the selection or remove if it is already selected, i.e. toggle selection
- [ ] When multiple entities are selected the Inspector just shows “Multiple selection”
- [ ] If I press DEL on multiple selection all selected nodes and links along the links that connected to these nodes get deleted
- [ ] If I drag multiple selections all selected nodes move at once

### How to implement

- Lift few React components that are responsible for slot highlights and dragged nodes rendering so that they can deal with multiple nodes at once

### How to test

- Test mass deletion of inter-dependent things so that we don’t fail if try to delete a thing “twice”
- Test add to selection / remove from selection / select none / reset to single item selection

### Out of scope

- Occupied slots handling

### LINKED ISSUE #768: Untangle Patch container
### Rationale

After introduction of new editor modes(resizing, panning, etc), `Patch` container became quite messy. We have to check for current mode in almost every event handler and keep a lot of state that is used for a single mode.

`Patch` code should be split into chunks that explicitly describe how events should be handled for each editor mode, what should be rendered, and rules for switching to other modes. Something like state machine.

Also, see discussion in #729

`xod-client`

evgenykochetkov:
Closes #364
Closes #419 
Closes #768


packages/xod-client/src/editor/containers/patchModes/linking.jsx
```diff
@@ -0,0 +1,104 @@
+import R from 'ramda';
+import React from 'react';
+import { HotKeys } from 'react-hotkeys';
+
+import { EDITOR_MODE } from '../../constants';
+
+import PatchSVG from '../../../project/components/PatchSVG';
+import * as Layers from '../../../project/components/layers';
+
+import {
```
brusherru: Looks strange, that these utils imported from "selecting" mode.
How about to move it somewhere into "modeUtils.js" or something like this?

packages/xod-client/src/editor/containers/patchModes/moving.jsx
```diff
@@ -0,0 +1,178 @@
...
+    )(api.props.nodes);
+    const draggedComments = R.compose(
+      translatePositions(deltaPosition),
+      R.pick(draggedCommentIds)
+    )(api.props.comments);
+
+    const [draggedLinks, idleLinks] = R.compose(
+      R.over(R.lensIndex(0), updateLinksPositions(draggedNodeIds, deltaPosition)),
+      R.partition(isLinkConnectedToNodeIds(draggedNodeIds)),
+      R.values // why Layers.Links accepts array?
```
brusherru: Maybe it's time to deal with this question? :)

packages/xod-client/src/editor/containers/patchModes/panning.jsx
```diff
@@ -0,0 +1,125 @@
...
+      api.setState({ isPanning: false });
+      api.props.actions.setOffset(offset);
+    }
+
+    if (isMiddleButtonPressed(event)) {
+      api.goToMode(EDITOR_MODE.DEFAULT);
+    }
+  },
+  onKeyUp(api, event) {
+    if (event.key !== ' ') return;
```
brusherru: Isn't better to use event.keyCode?

brusherru:
Awesome! 👍 
Don't forget to squash commits

PR #784, opened 2017-09-20
implement defer-* nodes to allow feedback loops

### LINKED ISSUE #733: As a xoder I want to make patches with feedback loops so that I can implement state machines
### Rationale

Right now the only way to make a program which current state depends on the past is to use composition of very few accumulator nodes like `count` or `flip-flop`. This is very limiting. We’re going to allow feeding output values back to inputs upstream.

Enabling graph loops directly would cause two problems:

1. The graph would fail to sort topologically what’s crucial for the C++ runtime
2. Even if we define an order somehow, we’d potentially get dead locks and board hangs

So we need a trick. Like vvvv does, we could introduce few special nodes `defer-boolean`, `defer-number`, `defer-string`, etc in `xod/core`. They should act as loop breakers which give a chance any transaction to be completed.

They are single-input/single-output nodes which implementation is simple `setTimeout(ctx, 0);` that would allow firing new values not right on input update, but as soon as possible, i.e. they would actually cause a new transaction. It solves the problem (2).

To make toposort possible we should exclude all `xod/core/defer-*` nodes and their links from the sorting algorithm input. It would solve the problem (1).

### Acceptance criteria

- [x] I can make a program that counts seconds up to 10, then resets and counts again. It could be done with `clock`, `count`, `less`, and `text-lcd`: output of the `less` is fed back to `count`’s `RST`

### How to implement

- Implement the new native nodes `defer-*`
- Exclude them from toposort in Arduino transpiler

### How to test

Add a fixture from acceptance criteria to make sure a program with a feedback loop transpiles correctly.

evgenykochetkov:
Closes #733


packages/xod-project/src/patch.js
```diff
@@ -738,6 +738,24 @@ export const upsertComments = def(
...
+const listLinksNotFromDeferredNodes = def(
+  'listLinksNotFromDeferredNodes :: Patch -> [Link]',
+  R.converge(
+    (deferredNodeIds, links) => R.reject(
+      R.pipe(Link.getLinkOutputNodeId, isAmong(deferredNodeIds)),
+      links
+    ),
+    [
+      R.compose(
+        R.map(Node.getNodeId),
```
brusherru: Looks like `getDeferredNodeIds` function, that could be reused in the transpiler.js `getDeferredNodeIds` :)

packages/xod-arduino/src/transpiler.js
```diff
@@ -121,12 +121,17 @@ const formatValueLiteral = def(
...
 const createTConfig = def(
-  'createTConfig :: PatchPath -> Project -> TConfig',
-  (path, project) => R.applySpec({
-    NODE_COUNT: R.compose(getNodeCount, Project.getPatchByPathUnsafe(path)),
-    MAX_OUTPUT_COUNT: getOutputCount,
-    XOD_DEBUG: R.F,
-  })(project)
+  'createTConfig :: PatchPath -> [TNodeId] -> Project -> TConfig',
+  (path, deferredNodeIds, project) => R.compose(
+    R.merge({
```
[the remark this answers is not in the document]
evgenykochetkov: ¯\\_(ツ)_/¯
I thought it's a bit more readable to separate things that are derived from `Project` and things that are pretty much constants.

evgenykochetkov:
Hold on, I think we can get rid of boolean flags in node implementations and go back to original idea of using timeouts. 🤔

evgenykochetkov:
`setTimeout(nid, 0)` is definitely not working if we are toposorting node graph and clearing dirty flags after each transaction.

If we evaluate them first, they will never have a chance to handle dirty inputs.

If we evaluate them last, no nodes will handle values emitted by them.

nkrkv:
Could you confirm it by running a live example?

The intention is (the current platform implementation with changed comments):

```cpp
bool isTimedOut(NodeId nid) {
    // note the strict comparison
    return g_schedule[nid] < transactionTime();
}

void runTransaction() {
    g_transactionTime = millis();

    // ...

    for (NodeId nid = 0; nid < NODE_COUNT; ++nid) {
        if (isNodeDirty(nid)) {
            // It could run setTimeOut(nid, 0) and the schedule will be set
            // to the current transaction time
            evaluateNode(nid);

            // This will NOT clear timeout because of the strict comparison
            if (isTimedOut(nid))
                clearTimeout(nid);
        }
    }

    // Well, it would clear dirtieness of defer-* too, no problem (see below)
    memset(g_dirtyFlags, 0, sizeof(g_dirtyFlags));

    // ...
}

void idle() {
    TimeMs now = millis();
    for (NodeId nid = 0; nid < NODE_COUNT; ++nid) {
        TimeMs t = g_schedule[nid];
        // This will mark defer-* dirty again
        if (t && t < now)
            markNodeDirty(nid);
    }
}
```

If there’s a flaw that stops us from using `setTimeout(nid, 0)`, let’s fix it and make it possible. Yep?

evgenykochetkov:
`setTimeout` works just fine. It's the evaluation order that's the problem.

Here is a "count to 10" patch from acceptance criteria:
screen shot 2017-09-25 at 16 33 18

We have a `count`-`gt`-`defer` loop here.

Let's take a look at the ways we can toposort and evaluate it:

0. `count`
1. `gt`
2. `defer`

will evaluate like this:

- first transaction
  - `count` evaluates, emits value, marks `gt` as dirty
  - `gt` evaluates, emits value, marks `defer` as dirty
  -  `defer` evaluates, sets timeout
  - transaction ends, all dirty flags are reset etc
- `idle` runs, marks `defer` as dirty
- next transaction
  - `count` is not dirty, skip it
  - `gt` is not dirty, skip it
  -  `defer` evaluates, emits value, marks `count` as dirty. But `count` will not evaluate in this transaction. It has nodeId = 0, we already skipped it.
  - transaction ends,  **all dirty flags are reset**, `count` did not get a chance to evaluate and is not dirty anymore.

With other ways pretty much the same thing will happen. The value that needs to get from last node to first will be lost.

nkrkv:
:open_mouth: Now I understand. Yep, it’s a problem. Will try to find a solution that satisfies both restrictions (implementation complexity *and* the evaluation order)

packages/xod-arduino/platform/program.tpl.cpp
```diff
@@ -72,4 +72,8 @@ namespace xod {
         &storage_{{ id }}{{#unless @last }},{{/unless }}
       {{/each}}
     };
+
+    const NodeId g_defer_node_ids[DEFER_NODE_COUNT] PROGMEM = {
```
nkrkv: Please, use `camelCase`. `g_` is from Hungarian notation and stands for “Global”:

```cpp
const NodeId g_deferNodeIds[DEFER_NODE_COUNT] PROGMEM = {
```

packages/xod-arduino/platform/runtime.cpp
```diff
@@ -346,6 +347,19 @@ void runTransaction() {
...
+    // defer-* nodes are always at the very bottom of the graph,
+    // so no one will recieve values emitted by them.
+    // We must evaluate them before everybody else
+    // to give them a chance to emit values.
+    for (NodeId nid : g_defer_node_ids) {
+        if (isNodeDirty(nid)) {
+            evaluateNode(nid);
+            // clear node dirty flag, so it will evaluate
+            // on "regular" pass only if it has a dirty input
+            g_dirtyFlags[nid] &= 0x11111110;
```
nkrkv: If I understand correctly, you can safely do:

```cpp
g_dirtyFlags[nid] = 0;
```

instead.

BTW, if we’d guarantee all `defer-*` nodes are at the tail of the list, you could use a single `memset` instead of the `for` loop.

packages/xod-arduino/src/transpiler.js
```diff
@@ -121,12 +121,17 @@ const formatValueLiteral = def(
...
 // Creates a TConfig object from entry-point path and project
 const createTConfig = def(
-  'createTConfig :: PatchPath -> Project -> TConfig',
-  (path, project) => R.applySpec({
-    NODE_COUNT: R.compose(getNodeCount, Project.getPatchByPathUnsafe(path)),
-    MAX_OUTPUT_COUNT: getOutputCount,
-    XOD_DEBUG: R.F,
-  })(project)
+  'createTConfig :: PatchPath -> [TNodeId] -> Project -> TConfig',
+  (path, deferredNodeIds, project) => R.compose(
```
nkrkv: You haven’t to pass the `deferredNodeIds` if toposort them to the tail. Counting could be done right inside the function implementation.

evgenykochetkov:
@nkrkv done, check it out please

packages/xod-project/src/patch.js
```diff
@@ -746,28 +746,57 @@ export const upsertComments = def(
...
+          R.map(Node.getNodeId),
+          listNodes
+        ),
+        R.compose(
+          R.map(Link.getLinkNodeIds),
+          listLinksNotFromDeferredNodes
+        ),
+      ]
+    )
+  )(patch)
```
nkrkv: The composition now looks quite complex. Perhaps, it is a good idea to split the function into two. Something like:

```
toposortGraph :: Patch -> Either Error [NodeId]
sendDeferNodesToBottom :: Patch -> [NodeId] -> [NodeId]

getTopology = compose(sendDeferNodesToBottom, toposortGraph);
```

PR #790, opened 2017-09-28
Update functional tests in xod-deploy and add support of Rejected response type

brusherru:
There is no issue 😬 

Because of changes in the cloud compilation backend (added some checks and added a new response type `Rejected`), had to tweak the functional tests and add support for the new response type.


brusherru:
@nkrkv It's done! Check it out please and I'll merge it.

packages/xod-deploy/test-func/cloudCompiler.spec.js
```diff
@@ -5,20 +5,22 @@ import { readFile, emptyDir } from 'fs-extra';
 import { COMPILATION_ERRORS } from '../src/constants';
 import { compile, saveCompiledBinary } from '../src/cloudCompiler';
 
-const code = '#include <Arduino.h>\nvoid setup() {}\nvoid loop() {}';
+// const code = '#include <Arduino.h>\nvoid setup() {}\nvoid loop() {}';
```
nkrkv: Trash comment

PR #791, opened 2017-09-28
Add debug tab pane with interactive watch nodes (in debug session)

### LINKED ISSUE #33: As a user, I want to place watch nodes, so that I can view some values in real-time while interaction session is in progress
### Rationale

It would help a lot to understand what’s going on in a XOD program if I can view values on outputs I’m interested in real-time.

To achieve it we’re going to introduce `xod/core/watch` node that has a single (unnamed) input of type string. When a debug session is active the watch shows the value passed as its body, i.e. instead of own label.

### Prerequisites

- #459 

### Acceptance criteria

- [x] I can place a `xod/core/watch` like a regular node on my patch
- [x] The watch node has outstanding look
- [x] I can link watch’s input like any other pin
- [x] I can Deploy → Upload → :heavy_check_mark: Debug → Upload to start monitoring the value incoming in real time
- [x] I can simply Deploy → Upload and it would not have any differences in memory, etc regardless of whether I have watches or not
- [x] On Deploy → Debug a new tab named “Debug” is opened and shows the entry point patch
- [x] Visually a patch chosen as entry point doesn’t get offset on debug session start
- [ ] The serial log is embedded to the Debug tab pane
- [ ] When a debug session ends the Debug tab gets closed automatically and I’m switched back to the entry point patch tab or the patch that changed

### How to implement

- Remove watch nodes on transpilation if it is targeted toward regular upload, i.e. not for debug session
- Record an ID map from long XOD node IDs to short C++ `NodeId`’s
- Implement `xod/core/watch` like a regular node that `Serial.println(...)` in it’s `evaluate`. A `ctx` argument is passed to `evaluate` it’s a node ID which we could use to correctly bind the data to a watch.
- Use some special *textual* format to send watched value updates. E.g. `+XOD:42=76.9859` plus line break, where 42 is a node ID and `76.9859` is input value (text)
- Monitor incoming serial port data for the lines matching debug format, decode data, map ID back and update the view of corresponding watch node
- Handle wrong IDs gracefully: do not crash, do not update watches, just dump the line as is to the debug log

brusherru:
It closes #33.

There are a couple of imperfections:
1. Slow redux-devtools in the electron version when debugger gets a lot of messages on high speed. It should be turned on by clicking a special item in the menu.
2. Debugger Log has a native scroll. There are some difficulties to integrate `CustomScroll` component. 😬


packages/xod-arduino/src/transpiler.js
```diff
@@ -121,11 +127,12 @@ const formatValueLiteral = def(
...
 const createTConfig = def(
-  'createTConfig :: PatchPath -> Project -> TConfig',
-  (path, project) => R.applySpec({
+  'createTConfig :: TranspilationOptions -> PatchPath -> Project -> TConfig',
+  (opts, path, project) => R.applySpec({
     NODE_COUNT: R.compose(getNodeCount, Project.getPatchByPathUnsafe(path)),
     MAX_OUTPUT_COUNT: getOutputCount,
-    XOD_DEBUG: R.F,
+    XOD_DEBUG: () => (opts.debug),
+    XOD_DEBUG_RUNTIME: () => (opts.debugRuntime),
```
nkrkv: To KISS it would be much better to have only `debug` field in the options. `debugRuntime` is quite pointless. It shouldn’t be mentioned in JS. We can keep `//#define XOD_DEBUG_ENABLE_TRACE` commented out right in the template, unconditionally.

packages/xod-arduino/src/transpiler.js
```diff
@@ -91,10 +91,16 @@ const getPatchByNodeId = def(
 );
 
 const toposortProject = def(
-  'toposortProject :: PatchPath -> Project -> Either Error Project',
+  'toposortProject :: PatchPath -> Project -> Either Error ProjectWithNodeIdsMap',
```
nkrkv: The `ProjectWithNodeIdsMap` looks wild a bit to me. Evolution would lead to `ProjectWithNodeIdsMapLinksExpandedAndDuplicationsSquashed`.

Wouldn’t it be more clear to have a simple additional field like `originalId` right inside any `TNode`?

packages/xod-arduino/src/transpiler.js
```diff
@@ -341,42 +348,52 @@ const createTNodes = def(
...
 
 export default def(
-  'transpile :: Project -> PatchPath -> Either Error String',
+  'transpile :: Project -> PatchPath -> TranspilationOptions -> Either Error TranspilationResult',
   R.compose(
-    R.map(renderProject),
+    R.map(R.applySpec({
+      code: R.pipe(R.prop('project'), renderProject),
+      nodeIdsMap: R.prop('nodeIdsMap'),
+    })),
```
nkrkv: Instead of this ad-hoc object we could export the `transformProject` and `getOriginalIdMap :: TProject -> StrMap NodeId`

packages/xod-arduino/src/types.js
```diff
@@ -30,6 +30,11 @@ const TNodeId = AliasType('TNodeId', $.Number);
 const TPinKey = OneOfType('TPinKey', [PinKey, PinLabel]);
 const DataValue = NullaryType('DataValue', R.complement(R.isNil));
 
+export const TranspilationOptions = Model('TranspilationOptions', {
+  debug: $.Boolean,
+  debugRuntime: $.Boolean,
```
nkrkv: `debugRuntime` seems unnecessary now.

packages/xod-client-electron/src/app/arduinoActions.js
```diff
@@ -134,9 +134,13 @@ export const checkPort = port => listPorts()
  * @param {PatchPath} patchPath Entry-point patch path
  * @returns {Promise<String, Error>} Promise with transpiled code or Error
  */
-export const doTranspileForArduino = ({ project, patchPath }) =>
+export const doTranspileForArduino = ({ project, patchPath, options = {} }) =>
```
nkrkv: Monsterification detected. Instead of the current

```js
export const doTranspileForArduino = ({ project, patchPath, options = {} }) => { ... }
```

could it be:

```js
// simpler name, no param destruction
export const transpileForArduino = (project, patchPath, options = {}) => { ... }
```

packages/xod-client-electron/src/app/arduinoActions.js
```diff
@@ -333,6 +370,7 @@ export const uploadToArduinoHandler = (event, payload) => {
         error: err,
         message,
         percentage,
+        additionalData,
```
nkrkv: Can’t get the idea why we have to carry the IDs map through everything in the `arduinoActions.js` again and again. Sometimes with such implicit stuff?

packages/xod-client/src/debugger/actionTypes.js
```diff
@@ -7,3 +7,5 @@ export const DEBUGGER_LOG_CLEAR = 'DEBUGGER_LOG_CLEAR';
 
 export const DEBUG_SESSION_STARTED = 'DEBUG_SESSION_STARTED';
 export const DEBUG_SESSION_STOPPED = 'DEBUG_SESSION_STOPPED';
+
+export const SAVE_DEBUGGER_NODE_IDS_MAP = 'SAVE_DEBUGGER_NODE_IDS_MAP';
```
nkrkv: Actions are facts. Looks like a Redux misuse. The effect should be applied as a reaction to something like `DEBUG_SESSION_STARTED`.

packages/xod-client/src/debugger/containers/Log.jsx
```diff
@@ -0,0 +1,54 @@
...
+import { getLog } from '../selectors';
+import SystemMessage from '../components/SystemMessage';
+import ErrorMessage from '../components/ErrorMessage';
+import LogMessage from '../components/LogMessage';
+import XodMessage from '../components/XodMessage';
+
+const renderLogMessage = (messageData, idx) => {
+  let Renderer = LogMessage;
+
+  switch (messageData.type) {
```
nkrkv: `R.cond` or even plain object lookup would be simpler.

packages/xod-project/src/project.js
```diff
@@ -581,6 +581,22 @@ export const rebasePatch = def(
       )
 );
 
+/**
+ * Updates patch inside Project and returns Either Error Project.
+ * It's a safe version of `R.over(lensPatch('some-patch'), someFn, project);`
+ */
+export const updatePatch = def(
```
evgenykochetkov: 👍🏻

packages/xod-client/stories/Debugger.jsx
```diff
@@ -0,0 +1,118 @@
...
+};
+
+// =============================================================================
+//
+// Run stories
+//
+// =============================================================================
+
+idle();
+running();
```
evgenykochetkov: 👍🏻

packages/xod-client/src/editor/actions.js
```diff
@@ -31,19 +31,6 @@ import {
 
 import { LINK_ERRORS } from '../messages/constants';
 
-export const setMode = mode => (dispatch, getState) => {
```
evgenykochetkov: Oooh, nice.

brusherru: Sound like irony. Do you think that's bad solution?

evgenykochetkov: Nope. In #783, I got rig of calls to `setMode`s in other actions(see [here](https://github.com/xodio/xod/pull/783/files#diff-717715e380a7256fcfd3949a5309d803L63) an onwards) to be able to "localize" `mode` in the future.

packages/xod-client-electron/src/view/containers/App.jsx
```diff
@@ -186,7 +209,17 @@ class App extends client.App {
       if (payload.success) {
         proc.success(payload.message);
         if (debug) {
-          debuggerIPC.sendStartDebuggerSession(ipcRenderer, port);
+          foldEither(
+            (error) => {
+              console.error(error); // eslint-disable-line no-console
+              this.props.actions.addError(error.message);
```
nkrkv: Why duplicate to the `console`?

packages/xod-client-electron/src/view/containers/App.jsx
```diff
@@ -186,7 +209,17 @@ class App extends client.App {
...
+            },
+            (nodeIdsMap) => {
+              this.props.actions.saveDebuggerNodeIdsMap(nodeIdsMap);
```
nkrkv: Oh no. It’s still here. Can we rephrase the action to be a fact?

packages/xod-client/src/core/containers/App.jsx
```diff
@@ -22,7 +27,19 @@ import * as actions from '../actions';
...
 export default class App extends React.Component {
   onShowCodeArduino() {
-    this.transpile(transpileForArduino);
+    const tProject = this.transformProjectForTranspiler();
+    const eitherCode = transpile(tProject);
+
+    return foldEither(
+      (error) => {
+        console.error(error); // eslint-disable-line no-console
+        this.props.actions.addError(error.message);
```
nkrkv: Why `console.error` duplication?

nkrkv:
Also, `packages/xod-client/yarn.lock` update should be included in the changeset

brusherru:
@nkrkv check it out, please!

brusherru:
@nkrkv It's ready to be merged! Check it out!

nkrkv:
Mmmmonster kill :+1:

PR #792, opened 2017-10-02
cut/copy/paste

### LINKED ISSUE #38: As a user, I want to cut/copy/paste nodes and links, so that I can create repetitive patch clusters quickly
We should have an intuitive cut/copy/paste behavior like most editors do.

### Acceptance criteria

- [ ] I can copy then paste on the same patch. The cluster pasted would appear as zombie with selection preserved so that I can immediately drag it to a new place
- [ ] I can copy then paste on another patch or even another app instance
- [ ] If clipboard content belongs to XOD but pasting it would lead to an error (e.g. integrity fail), a error message is shown
- [ ] If clipboard content doesn’t belong to XOD, pasting results in nothing
- [ ] I can copy/paste individual nodes, multiple nodes or multiple nodes with links between them
- [ ] I can cut rather than copy which behaves like “copy then delete”

### How to implement

- On paste regenerate items IDs to avoid conflicts with existing items
- Use HTML5 features to make clipboard work with web version

### What to test

- Successful scenarios of copy-paste (unit test)
- Failing scenarios of copy-paste (unit test)

Don’t fake clipboard contents, use implemented functionality to forge the data required.

Don’t involve real clipboard of the OS in tests.

evgenykochetkov:
Closes #38


nkrkv:
Regarding the code :+1: 

I’ve tried to test the feature and all aspects seem to be solid and stable. Good job! Although I found few minor flaws that could be worked on now or later (in a dedicated tweak issue). It’s up to you to decide whether they could be tweaked in few hours or not.

1. If the clipboard has improper data (wouldn’t be pasted) than the paste operation should not deselect all. Try it a text editor.
2. If I copy then paste on the same patch I expect the pasted cluster to be at the same position as original with a little offset to the left and bottom (X ~ 1 pin width, Y ~ half of the node height), not viewport origin as currently
3. If I copy-paste some nodes, I expect the cluster to automatically include all links touching these nodes

packages/xod-client/src/editor/utils.js
```diff
@@ -27,3 +34,116 @@ export const getSelectedEntityIdsOfType = R.curry((entityType, selection) => R.c
...
+      resetPosition
+    )),
+    comments: R.map(R.over(
+      R.lens(XP.getCommentPosition, XP.setCommentPosition),
+      resetPosition
+    )),
+  })(entities);
+};
+
+// TODO: move to xod-func-tools?
```
brusherru: Good idea :)

evgenykochetkov:
PR is ready 🕺

@nkrkv @brusherru check it out

- added Edge compatibility
- implemented "If I copy-paste some nodes, I expect the cluster to automatically include all links touching these nodes"

"If I copy then paste on the same patch I expect the pasted cluster to be at the same position as original with a little offset to the left and bottom (X ~ 1 pin width, Y ~ half of the node height), not viewport origin as currently" will be done as a separate tweak issue

PR #796, opened 2017-10-03
Add ‘Select All’ command to desktop IDE

evgenykochetkov:
Because we could not select all text in an input before :(


evgenykochetkov:
Regarding shortcuts documentation(#812): should we really add an entry for "select all text in an input"?

nkrkv:
@evgenykochetkov oh, no. I thought it’s about selecting all entities on the patch board.

I can’t quite understand how it works. You add a command declaration, but not a handler. Is it `role` magic?

evgenykochetkov:
Yep, it's just `role`.

PR #798, opened 2017-10-03
Brand new tutorial

nkrkv:
Whew. The tutorial we have spoken about for the latest two months.


docs/tutorial/01-hello/README.md
```diff
@@ -0,0 +1,47 @@
...
+You can move nodes by dragging them. Position on the patch board does not
+affect your program, but can make it more readable.
+
+You can see a comment as well. Even though it is there, it does not affect the
+patch, because it is invisible to a machine. It’s only visible to you, a
+clever human. You can drag it to move too.
+
+## Project browser
+
+On the left side you will find a list with available projects. The list is
```
brusherru: "list with available projects" — it's not really true, cause we see only current project and projects in the library folder. Is this fact deliberately left out to simplify the learning?

nkrkv: Well, yes. It’s like “numbers start with 1” at the first class in a school.

This is the first contact and we have to say something. “On the left side you will find a list with available… wut”?

Do you have an idea how to improve the sentence and not to complicate things at the same time? That would be nice!

brusherru: Maybe we could tell that user see his opened project and projects available from library. And tell more about libraries, terminal nodes and etc in the next tutorials. Like we do with pulse type, we introduce it and promise to tell more later :)

nkrkv: OK. Changed it this way.

```diff
-On the left side you will find a list with available projects. The list is
-called a *Project Browser*. When you launch XOD for the first time the first
-item in the list is `welcome-to-xod`. This is a special tutorial project you
-can follow to learn XOD right inside the XOD IDE.
+On the left side you will find a list of patches grouped by a project or
+library name. The list is called a *Project Browser*. When you launch XOD for
+the first time the first item in the list is `welcome-to-xod`. This is a
+special tutorial project you can follow to learn XOD right inside the XOD IDE.
```

PR #801, opened 2017-10-04
Voodoo fix of delay in the runtime

brusherru:
Voodoo dance with working delay value in the `setup()` function.


packages/xod-arduino/platform/runtime.cpp
```diff
@@ -417,7 +417,7 @@ bool isTimedOut(NodeId nid) {
 //----------------------------------------------------------------------------
 void setup() {
     // FIXME: looks like there is a rounding bug. Waiting for 1 second fights it
```
evgenykochetkov: It's not 1 second anymore 🤔

brusherru: Oops 😬

PR #807, opened 2017-10-06
Drill down in debug tab

### LINKED ISSUE #780: As a xoder, I want to drill down into patch nodes while debugging so that I can see deeply nested watches
### Rationale

As of #33 I can use watches to monitor values on my entry patch. But it is often desirable to view some watches placed on inner patch nodes.

A conflict exists. Consider I have a patch `main` with two `button`’s in it. Where the `button` is a patch node composed of `digital-input`, its watch, and other nodes. If I upload & debug the `main` and then open a `button`, which watch I would see? To solve the problem we can introduce breadcrumbs in the Debug tab for the time span of a debug session.

In that case, when I choose to debug my `main` patch a breadcrumb line is shown containing a single entry — `main`. Then I can double click a concrete `button` instance to drill down. The breadcrumb line will show “`main` → `button`”. With double-clicking further, I can drill-down to an arbitrary level with an unambiguous path.

### Prerequisites

#33 

### Acceptance criteria

- [x] On Deploy → Debug a new tab named “Debug” is opened with a breadcrumb line containing a single item equal to entry patch
- [x] Visually a patch chosen as entry point doesn’t get offset on debug session start
- [x] In the debug tab I can double click nodes to drill down, breadcrumb depicts the path
- [x] Breadcrumbs show node label with fallback to patch name
- [x] I can click a former breadcrumb to drill-up
- [x] When I drill-up the patch clicked gets open but breadcrumb’s tail is still here so that I can click an item in it to switch back deeper
- [x] When I drill-up *and* double-click to drill down again, the tail gets replaced with the breadcrumb for the patch I’ve double clicked
- [x] When a debug session ends the tab corresponding to the last seen patch is opened

brusherru:
It closes #780.


brusherru:
Travis hanged out 😬

Issue #811, opened 2017-10-09
ESC should cancel linking


nkrkv:
Introduced somewhere in v0.13.0..v0.14.0

[Reported](https://forum.xod.io/t/xod-0-14-0-released/117/9?u=nkrkv) on the forum.

### Steps to reproduce

1. Click an output
2. Move the cursor over the patch board
3. Press ESC key

### Expected behavior

Linking is interrupter, I return to the normal mode

### Actual behavior

Key press is ignored

### Console output

Are there any errors in the console? Copy and paste the output
with all details provided by the browser or application.

PR #812, opened 2017-10-09
Document mouse and keyboard shortcuts

nkrkv:
Ad-hoc PR sourced from a [forum thread](https://forum.xod.io/t/is-there-ui-documentation/109/7).

I intentionally left out some shortcuts because their future does not look clear. E.g. `Ctrl+R` to rename: does this operation ever deserve a shortcut or at least such “tasty” keystroke?! This is yet to be discussed. For now, only stable and vital combinations are included.

Also note that in case of the PR approval, #796 and #811 should have changes in docs as well.

### DIFF
--- a/docs/guide/shortcuts/README.md
+++ b/docs/guide/shortcuts/README.md
@@ -1,0 +1,246 @@
+---
+title: Mouse and Keyboard Shortcuts
+---
+
+# Mouse and Keyboard Shortcuts
+
+For shortcuts the following convention is used:
+
+* <span class="ui label">C</span> — platform-neutral keyboard input;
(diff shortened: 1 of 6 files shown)


docs/guide/shortcuts/README.md
```diff
@@ -0,0 +1,246 @@
...
+        <span class="ui grey label">⌘ O</span>
+      </td>
+      <td>
+        Open project
+      </td>
+    </tr>
+  </tbody>
+</table>
+
+## Patch editing
```
brusherru: We can also add "Double click" on the Node — opens Node as a patch.

nkrkv: Yep. But in a version not yet released. We have to remember to add it right before 0.15.0 release.

PR #823, opened 2017-10-18
Fix crashing of IDE on patch delete

### LINKED ISSUE #818: Deleting a patch crashes IDE
### Steps to reproduce

1. Open `welcome-to-xod`
2. Click trash bin button to delete the `05-pwm` patch
3. Confirm the deletion

### Expected behavior

- The popup disappears
- The patch is deleted

### Actual behavior

- The popup stays intact
- The patch is not deleted

### Console output

```
Uncaught TypeError: Cannot read property 'patchPath' of null
    at reducer.js:273
    at _curryN.js:37
    at _arity.js:6
    at closeTabByPatchPath (reducer.js:284)
    at editorReducer (reducer.js:495)
    at combination (combineReducers.js:133)
    at undoableProject.js:124
    at dispatch (createStore.js:178)
    at middleware.js:13
    at index.js:14
```

### How to test

Add patch deletion test to the tail of functional IDE test

brusherru:
It fixes #818

### DIFF
--- a/packages/xod-client-electron/test-func/suite.spec.js
+++ b/packages/xod-client-electron/test-func/suite.spec.js
@@ -176,5 +176,18 @@ describe('IDE', () => {
         .then(() => page.getCodeboxValue())
         .then(code => assert.strictEqual(code, expectedCpp, 'Actual and expected C++ don’t match'))
     );
+    it('closes show code popup', () =>
+      page.closePopup()
+    );
+  });
+
+  describe('deleting a patch', () => {
+    it('opens patch from xod/core', () =>
+      page.openPatchFromProjectBrowser('clock')
+        .then(() => page.findPatchGroup('welcome-to-xod').click())
+        .then(() => page.assertPatchGroupExpanded('welcome-to-xod'))
+        .then(() => page.deletePatch('my-blink'))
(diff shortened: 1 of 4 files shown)


packages/xod-client-electron/test-func/suite.spec.js
```diff
@@ -177,5 +177,18 @@ describe('IDE', () => {
...
+  });
+
+  describe('deleting a patch', () => {
+    it('opens patch from xod/core', () =>
+      page.openPatchFromProjectBrowser('clock')
+        .then(() => page.findPatchGroup('welcome-to-xod').click())
+        .then(() => page.assertPatchGroupExpanded('welcome-to-xod'))
+        .then(() => page.deletePatch('my-blink'))
+        .then(() => page.assertNodeUnavailableInProjectBrowser('my-blink'))
+    );
```
nkrkv: Misguiding describe/it text. Now it is “deleting a patch opens patch from xod/core” :thinking: Does not reflect what’s going on. Why do we ever open the `clock`? Is it because we can’t delete a patch of the active tab? If so, lets add a comment about it.

Splitting into two its (one about open, another about deletion) would also make things more clear.

nkrkv:
Also, please fix the PR title

PR #825, opened 2017-10-19
Upgrade to new lerna and yarn workspaces

### LINKED ISSUE #805: Update Yarn, Lerna, Node
We’re lagging behind and loose free DX improvements. Let’s upgrade to Lerna 2.x and Yarn 1.x. This would add workspaces support, but start barking on the Node version. So, let’s update Node as well.

- [x] Bump Yarn for Travis
- [ ] Bump Yarn for AppVeyor
- [ ] Bump Yarn for Concourse
- [x] Do not cache `packages/*/node_modules` in Travis, leave just root `node_modules`
- [x] Bump Node for Travis
- [x] Bump Node for AppVeyor
- [ ] Bump Node for Concourse
- [x] Kill everything related to Lerna in `tools/`
- [x] Adjust the root `package.json` and `README.md` to support all maintenance features we had before the upgrade (dev, watch, dependency tree, scoping)

evgenykochetkov:
Closes #805 

- lerna 2.2.0
- yarn 1.10
- node 8.6
- faster builds
- much smaller desktop IDE size

For now we use our own fork of `electron-builder`, because official fix for monorepos with yarn-workspaces does not work 😕

⚠️ Also, please note that because we're dropping custom `tools/lerna-*` scripts in favour of native lerna features, some of the commands for executing common tasks have changed.


README.md
```diff
@@ -73,14 +73,15 @@ expect that the project is already built.
...
-    $ yarn build -- xod-cli
+    $ yarn build --scope xod-cli
 
-To rebuild only `xod-cli` without dependencies:
+To rebuild `xod-cli` and its dependencies:
 
-    $ yarn build -- xod-cli --only
+    $ yarn build --scope xod-cli --include-filtered-dependencies
 
+Those are standard [Lerna](https://github.com/lerna/lerna) flags, you can learn more about them [here](https://github.com/lerna/lerna#flags).
```
nkrkv: Please, don’t use words like here/there/link for link anchors. Link destination should be evident even if surrounding text would be hidden:

```
Those are standard [Lerna flags](https://github.com/lerna/lerna#flags).
```

…that’s all

packages/xod-client-browser/webpack.config.js
```diff
@@ -5,14 +5,14 @@ const CopyWebpackPlugin = require('copy-webpack-plugin');
...
-const fontAwesomePath = fs.realpathSync(pkgpath('node_modules/xod-client/node_modules/font-awesome'));
+const assetsPath = fs.realpathSync(pkgpath('../../node_modules/xod-client/src/core/assets'));
+const fontAwesomePath = fs.realpathSync(pkgpath('../../node_modules/font-awesome'));
 
 module.exports = {
   devtool: 'source-map',
   entry: [
     'babel-polyfill',
-    pkgpath('node_modules/xod-client/src/core/styles/main.scss'),
+    pkgpath('../../node_modules/xod-client/src/core/styles/main.scss'),
```
nkrkv: Looks fragile. Wouldn’t it a better idea to use that Node API function that searches for a package in parent directories? Something like:

```js
// modpath
//  :: String -- module name
//  -> String -- module path
//  -> String -- absolute path

...

entry: [
  'babel-polyfill',
  modpath('xod-client', 'src/core/styles/main.scss'),
  // ...
]
```

PR #831, opened 2017-10-23
Load bundled libraries without spawning them into workspace

### LINKED ISSUE #816: As a xoder I want new nodes to be available right away so that I have not to kill my workspace
### Rationale

On each version release, we have to write a note for existing users like “delete your workspace to get the new nodes”. This is frustrating.

We can automatically provide `xod/*` libraries at their most recent version. The idea is multiple locations lookup. Currently, we’re looking for libraries in `$workspace/lib` only. We can have two paths:

* `$workspace/lib` — primary
* `$app/lib` — secondary

Where app is a path to the installed IDE. On project load, for each library, we can look into the primary location and if found use it. Otherwise look in the secondary location.

Having this implemented, we should no longer spawn `xod/*` in a user’s workspace. That way he would see the newest versions automatically via the `$app/lib` location. As an emergency scenario, we leave the possibility to manually create `xod/core` in `$workspace/lib` so that another version would get priority over the bundled (newest) one.

### Acceptance criteria

- [x] I can load a project in a workspace without `lib` subdir and have `xod/*` libraries available and working
- [x] I can create `xod/units` in `$workspace/lib` that differs from the standard one and it would be preferred over the bundled one
- [x] `xodc transpile` resolves libraries correctly

### LINKED ISSUE #826: Creating a project named `lib` breaks workspace
### Steps to reproduce

- Create a project named `lib`

### Expected behavior

- I am warned that I should not do this?

### Actual behavior

- Error message `Could not open selected project: ENOTDIR: not a directory, scandir '/Users/user/xod/workspace/lib/main/patch.xodp'` is shown
- I can't open any project until I manually delete `main` folder from `lib`

brusherru:
It closes #816 and fixes #826.


packages/xod-cli/package.json
```diff
@@ -3,7 +3,8 @@
   "version": "0.14.0",
   "description": "XOD project: Command Line Interface",
   "scripts": {
-    "build": "babel src/ -d bin/ --source-maps",
+    "copy-workspace": "cpx \"../../workspace/__lib__/**\" \"bin/workspace/__lib__\"",
+    "build": "babel src/ -d bin/ --source-maps && yarn run copy-workspace",
```
nkrkv: The common pattern used in other places is having `build:src`, `build:workspace`, and `build` that calls both. Can we use it here as well?

packages/xod-cli/src/xodc-transpile.js
```diff
@@ -8,6 +8,8 @@ import { loadProject, readJSON, writeFile } from 'xod-fs';
 import { transformProject, transpile } from 'xod-arduino';
 import * as msg from './messages';
 
+const bundledWorkspace = path.resolve(__dirname, 'workspace');
```
nkrkv: Copying to `./bin/workspace/__lib__` looks strange. Why is the `workspace` medium required? Is it a binary (`bin` parent required)?

Maybe we can copy the libraries simply to `packages/xod-cli/__lib__`?

packages/xod-fs/src/constants.js
```diff
@@ -1,7 +1,7 @@
 export const WORKSPACE_FILENAME = '.xodworkspace';
 export const DEFAULT_WORKSPACE_PATH = '~/xod/';
 export const DEFAULT_PROJECT_NAME = 'welcome-to-xod';
-export const LIBS_FOLDERNAME = 'lib';
+export const LIBS_FOLDERNAME = '__lib__';
```
nkrkv: No new terms, please. `LIBS_DIRNAME`

packages/xod-fs/src/load.js
```diff
@@ -122,35 +122,48 @@ export const loadProjectWithoutLibs = projectPath => R.composeP(
...
+ * project directory, e.g. `/path/to/workspace/my-proj`.
+ *
+ * Also it loads a libs from additional workspace, that contains specified project.
+ * If `addWorkspaceList` is not nil and empty, it will load libs from this list of workspaces too.
+ * If all workspaces have libs with same names, will be loaded only one —
+ * from local workspace, that contains project (it has highest priority).
+ *
+ * Returns a Promise of complete `Project` (see `xod-project`).
+ */
+export const loadProject = (projectPath, addWorkspaceList = []) =>
```
nkrkv: Ough. The comment should be grammar-checked, too many mistakes.

`add` is a verb, use `additional` or better `extra`… `extraLibDirs`

I see no point in the `workspace` medium. Could it operate on library directories directly? It would make things easier to understand and maintain.

packages/xod-fs/src/spawn.js
```diff
@@ -21,11 +22,8 @@ export const spawnWorkspaceFile = workspacePath =>
     .catch(rejectWithCode(ERROR_CODES.CANT_CREATE_WORKSPACE_FILE));
 
 // :: Path -> Promise Path Error
-export const spawnStdLib = curry((stdLibPath, workspacePath) =>
-  copy(stdLibPath, resolveLibPath(workspacePath), copyOptions)
-  .then(() => workspacePath)
-  .catch(rejectWithCode(ERROR_CODES.CANT_COPY_STDLIB))
-);
+export const createLibDir = workspace => mkdir(resolveLibPath(workspace))
+  .then(() => workspace);
```
nkrkv: Are we need it?

packages/xod-fs/src/load.js
```diff
@@ -122,35 +122,48 @@ export const loadProjectWithoutLibs = projectPath => R.composeP(
   readDir
 )(projectPath);
 
-// :: Path -> Path -> Path -> Promise [File] Error
-export const loadProjectWithLibs = (projectPath, workspace, libWorkspace = workspace) => {
-  const fullProjectPath = resolvePath(path.resolve(workspace, projectPath));
-  const libPath = resolvePath(libWorkspace);
+// :: Path -> Path -> Path -> Path -> Promise [File] Error
```
evgenykochetkov: This signature does not look right🤔

packages/xod-fs/src/load.js
```diff
@@ -122,35 +122,48 @@ export const loadProjectWithoutLibs = projectPath => R.composeP(
...
+export const loadProjectWithLibs = R.curry(
+  (addWorkspaceList, projectPath, workspace) => {
```
evgenykochetkov: `addWorkspaceList` name is looks to me

packages/xod-cli/package.json
brusherru: Okay, but it's not truth 😬 
`xod-client-electron` contains `build:gui`, `build:app`, `build`, that runs both and runs `copy-workspace` script. This is almost the exact clone (but copies only libraries).
So... rename it in both places into `build:workspace`?
Also, I'm not sure this is a good name, actually we're don't build workspace, we copy it...

packages/xod-func-tools/src/index.js
```diff
@@ -332,6 +332,28 @@ export const invertMap = def(
...
+  R.reduce(
+    (acc, nextList) => R.compose(
+      R.append(R.__, acc),
+      R.reject(R.__, nextList),
+      isAmong,
+      R.flatten
+    )(acc),
+    []
+  )
+);
```
evgenykochetkov: `flatten` could make our list a bit too flat😁, `unnest` is probably better.
Also, why not use `without`? Something like 
```
R.reduce(
  (acc, nextList) => R.append(R.without(R.unnest(acc), nextList), acc),
  []
)
```

brusherru:
@nkrkv @evgenykochetkov fixed!
Check it out, please (last commit). And then I'll squash it.

nkrkv:
Oh, fix the PR title, please

nkrkv:
@brusherru Yeah, works fine :ok_hand:

brusherru:
🙌  Squashing and merging...

PR #834, opened 2017-10-24
Fix tab sorting

### LINKED ISSUE #802: Fix tab sorting behaviour
### Steps to reproduce

let's say we have 4 open tabs:
```
aa  bb  cc  dd
```

and drag `dd` between `aa` and `bb`:
```
   dd
   |
aa V bb cc
```

### Expected tab order

```
aa dd bb cc
```

### Actual tab order

```
aa dd cc bb 
```
`bb` is just swapped with `dd`

### How to fix

See here: https://github.com/xodio/xod/blob/master/packages/xod-client/src/editor/containers/Tabs.jsx#L79

brusherru:
It fixes #802

### DIFF
--- a/packages/xod-client/src/editor/containers/Tabs.jsx
+++ b/packages/xod-client/src/editor/containers/Tabs.jsx
@@ -8,8 +8,6 @@ import {
   SortableElement as sortableElement,
 } from 'react-sortable-hoc';
 
-import { swap } from 'xod-func-tools';
-
 import * as Actions from '../actions';
 import * as ProjectSelectors from '../../project/selectors';
 import { assocIndexes, indexById } from '../../utils/array';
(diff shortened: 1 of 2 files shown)


packages/xod-client/src/editor/containers/Tabs.jsx
```diff
@@ -75,8 +73,12 @@ class Tabs extends React.Component {
...
-  onSortEnd(changes) {
-    const sortedTabs = swap(changes.oldIndex, changes.newIndex, this.getTabs());
+  onSortEnd({ oldIndex, newIndex }) {
+    const tabs = this.getTabs();
+    const sortedTabs = R.compose(
+      R.insert(newIndex, tabs[oldIndex]),
+      R.remove(oldIndex, 1)
+    )(tabs);
     const newTabs = assocIndexes(sortedTabs);
     const indexedTabs = indexById(newTabs);
```
evgenykochetkov: No need for these intermediate constants anymore, it all can be in one `compose`

brusherru: Done!

PR #836, opened 2017-10-25
Display compilation limits

### LINKED ISSUE #829: Display compilation limits
### Rationale

Currently, there’s no way to know how many cloud compilations left for my account. It should be shown next to “Compile in cloud“ checkbox in the Upload dialog. Although the value is provided by a server, the UI should not be blocked for such requests.

### Acceptance criteria

- [x] When I hit “Deploy → Upload code to Arduino” I see a quota value next to checkbox label. Like “[ ] Compile in the cloud (123 left)”.
- [x] When I use the cloud compilation feature, the quota gets updated after the compilation, so that I’d see actual quota value next time I open the dialog.
- [x] UI does not wait for quota values.
- [x] If the request fails to complete successfully (no internet, service down), I see “[ ] Compile in the cloud (currently offline)”.

### How to implement

- [x] The request for quota is sent upon client start-up and put in the Redux state
- [x] The request for quota is sent again upon compilation response and overrides Redux state
- [x] The upload UI shows the quota value directly from Redux state

alexander-matsievsky:
Closes #829.

### DIFF
--- a/packages/xod-client-electron/src/upload/components/PopupUploadConfig.jsx
+++ b/packages/xod-client-electron/src/upload/components/PopupUploadConfig.jsx
@@ -41,6 +41,7 @@ class PopupUploadConfig extends React.Component {
     this.getSelectedBoard()
       .then(selectedBoard => this.getBoards(selectedBoard));
     this.getPorts();
+    this.props.updateCompileLimit();
   }
 
   onClose() {
@@ -252,6 +253,7 @@ class PopupUploadConfig extends React.Component {
   render() {
     const boards = this.renderBoardSelect();
     const ports = this.renderPortSelect();
+    const compileLimitLeft = this.props.compileLimitLeft;
 
     return (
       <PopupForm
@@ -270,10 +272,14 @@ class PopupUploadConfig extends React.Component {
             id="compileInCloud"
             type="checkbox"
             checked={this.state.doCompileInCloud}
+            disabled={compileLimitLeft <= 0}
             onChange={this.onCloudCompilationChanged}
           />
           <label htmlFor="compileInCloud">
-            Compile in the cloud (experimental)
+            Compile in the cloud&nbsp;
+            {compileLimitLeft
+              ? <span>({compileLimitLeft} left)</span>
+              : <span>(currently offline)</span>}
           </label>
         </div>
         <div className="ModalContent">
@@ -300,6 +306,8 @@ class PopupUploadConfig extends React.Component {
 PopupUploadConfig.propTypes = {
   isVisible: PropTypes.bool,
   selectedPort: PropTypes.object,
+  compileLimitLeft: PropTypes.number,
+  updateCompileLimit: PropTypes.func,
   getSelectedBoard: PropTypes.func,
   listBoards: PropTypes.func,
   listPorts: PropTypes.func,
--- a/packages/xod-client-electron/src/view/containers/App.jsx
+++ b/packages/xod-client-electron/src/view/containers/App.jsx
@@ -227,6 +227,7 @@ class App extends client.App {
       }
       // Remove listener if process is finished.
       ipcRenderer.removeAllListeners(UPLOAD_TO_ARDUINO);
+      this.props.actions.updateCompileLimit();
     });
   }
 
@@ -549,6 +550,8 @@ class App extends client.App {
         selectedPort={this.props.selectedPort}
         listBoards={this.listBoards}
         listPorts={this.listPorts}
+        compileLimitLeft={this.props.compileLimitLeft}
+        updateCompileLimit={this.props.actions.updateCompileLimit}
         onBoardChanged={this.onArduinoTargetBoardChange}
         onPortChanged={this.onSerialPortChange}
         onUpload={this.onUploadToArduino}
@@ -658,6 +661,7 @@ App.propTypes = R.merge(client.App.propTypes, {
   project: client.sanctuaryPropType(Project),
   actions: PropTypes.objectOf(PropTypes.func),
   upload: PropTypes.object,
+  compileLimitLeft: PropTypes.number,
   workspace: PropTypes.string,
   selectedPort: PropTypes.object,
 });
@@ -674,6 +678,7 @@ const mapStateToProps = R.applySpec({
   saveProcess: getSaveProcess,
   currentPatchPath: client.getCurrentPatchPath,
   selectedPort: getSelectedSerialPort,
+  compileLimitLeft: client.getCompileLimitLeft,
   popups: {
     createProject: client.getPopupVisibility(client.POPUP_ID.CREATING_PROJECT),
     projectSelection: client.getPopupVisibility(client.POPUP_ID.OPENING_PROJECT),
--- a/packages/xod-client/src/user/actions.js
+++ b/packages/xod-client/src/user/actions.js
@@ -1,0 +1,12 @@
+import { getCompileLimitUrl } from '../utils/urls';
+import { UPDATE_COMPILE_LIMIT } from './actionTypes';
+
+export const updateCompileLimit = () => dispatch =>
+  fetch(getCompileLimitUrl())
+    .then(res => (res.ok ? res.json() : null))
+    .catch(() => null)
+    .then(limit => dispatch({
+      type: UPDATE_COMPILE_LIMIT,
+      payload: limit,
+    }));
+export default {};
--- a/packages/xod-client/src/user/reducer.js
+++ b/packages/xod-client/src/user/reducer.js
@@ -1,0 +1,14 @@
+import { UPDATE_COMPILE_LIMIT } from './actionTypes';
+
+const userReducer = (state = {}, action) => {
+  switch (action.type) {
+    case UPDATE_COMPILE_LIMIT: {
+      const limit = action.payload;
+      return { ...state, limit };
+    }
+    default:
+      return state;
+  }
+};
+
+export default userReducer;
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -24,6 +24,7 @@ module.exports = {
   ],
 
   globals: {
+    fetch: true,
     window: true,
     document: true,
     describe: true,
--- a/packages/xod-client-browser/webpack.config.js
+++ b/packages/xod-client-browser/webpack.config.js
@@ -96,6 +96,7 @@ module.exports = {
       'process.env.NODE_ENV': JSON.stringify(process.env.NODE_ENV || 'development'),
       'process.env.XOD_HM_DEF': JSON.stringify(process.env.XOD_HM_DEF || false),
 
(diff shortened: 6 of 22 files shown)


brusherru:
First of all, it has non-conventional commit messages and it should be rebased on `master` 😬

alexander-matsievsky:
@brusherru It will be squashed to smth like `feat(client): display compilation limits`.

packages/xod-client/src/user/actions.js
```diff
@@ -0,0 +1,18 @@
+import url from 'url';
+import { UPDATE_COMPILE_LIMIT } from './actionTypes';
+import { ORIGIN_COMPILE } from './constants';
+
+export const updateCompileLimit = () => (dispatch) => {
+  fetch(url.format({
+    ...url.parse(process
+      ? process.env.ORIGIN_COMPILE || ORIGIN_COMPILE
```
brusherru: Unnecessary check for `process` variable... see `webpack.config.js` and `xod-client/src/utils/siteLinks.js`

brusherru: Also `ORIGIN_COMPILE` sounds different from `XOD_SITE_DOMAIN`, `XOD_FORUM_DOMAIN`... Logically, it should be something like `XOD_COMPILE_DOMAIN`.

And we've discussed that we should have the only flag just for switching between production/staging hostnames... so it should be something completely different and probably used by new functions in the `siteLinks.js`

packages/xod-client/src/user/actions.js
```diff
@@ -0,0 +1,18 @@
...
+      : ORIGIN_COMPILE),
+    pathname: '/limits',
+  }))
+    .then(res => (res.ok ? res.json() : null))
```
brusherru: What about retries, errors and unhandled promise rejection?

packages/xod-client/src/user/reducer.js
```diff
@@ -0,0 +1,20 @@
+import { UPDATE_COMPILE_LIMIT } from './actionTypes';
+
+const userReducer = (state = {}, action) => {
+  switch (action.type) {
+    case UPDATE_COMPILE_LIMIT: {
+      const limit = action.payload;
+      return {
+        ...state,
+        limit: !limit ? null : {
```
brusherru: Looks ugly 😞 
I prefer to import Ramda and use their functions to update state...

packages/xod-client-electron/src/view/containers/App.jsx
```diff
@@ -179,6 +179,7 @@ class App extends client.App {
     const errored = foldEither(
       (error) => {
         proc.fail(error.message, 0);
+        this.props.actions.updateCompileLimit();
```
brusherru: I'm not sure that we really need to `updateCompileLimit` here — it's about a transformation of the code...

packages/xod-client-electron/src/view/containers/App.jsx
```diff
@@ -224,6 +226,7 @@ class App extends client.App {
           R.reject(R.isNil)
         )([payload.message, payload.error.stdout, payload.stderr]);
         proc.fail(failureMessage, payload.percentage);
+        this.props.actions.updateCompileLimit();
       }
       // Remove listener if process is finished.
       ipcRenderer.removeAllListeners(UPLOAD_TO_ARDUINO);
```
brusherru: Looks like you don't need to call `updateCompileLimit` twice in both conditions, just call it once here.

packages/xod-client/src/user/reducer.js
```diff
@@ -0,0 +1,20 @@
...
+          ...limit,
+          left: Math.max(0, limit.limit - limit.pending - limit.used),
```
brusherru: 🤔 
I think it will be better to move it into the selector (it could be calculated, so we don't need to store in the state).

packages/xod-client-electron/src/upload/components/PopupUploadConfig.jsx
```diff
@@ -273,7 +274,10 @@ class PopupUploadConfig extends React.Component {
             onChange={this.onCloudCompilationChanged}
           />
           <label htmlFor="compileInCloud">
-            Compile in the cloud (experimental)
+            Compile in the cloud&nbsp;
+            {compileLimit
+              ? <span>({compileLimit.left} left)</span>
+              : <span>(currently offline)</span>}
```
evgenykochetkov: Probably, we should also disable this option in that case

alexander-matsievsky: @evgenykochetkov 

```typescript
compileLimit: CompilationLimit | null
```
where `null` signals the absence of the compilation quota for a user. Basically, it's not about networking.

brusherru: But the message is about networking... 😬

brusherru: + we should disable this option if no quota left

alexander-matsievsky: @brusherru https://github.com/xodio/xod/issues/829

> If the request fails to complete successfully (no internet, service down), I see “[ ] Compile in the cloud (currently offline)”.

packages/xod-client/src/user/actions.js
```diff
@@ -1,12 +1,12 @@
...
-import { ORIGIN_COMPILE } from './constants';
+import { XOD_COMPILE_DOMAIN } from './constants';
 
 export const updateCompileLimit = () => (dispatch) => {
   fetch(url.format({
     ...url.parse(process
-      ? process.env.ORIGIN_COMPILE || ORIGIN_COMPILE
-      : ORIGIN_COMPILE),
+      ? process.env.XOD_COMPILE_DOMAIN || XOD_COMPILE_DOMAIN
+      : XOD_COMPILE_DOMAIN),
```
nkrkv: There should not be a dedicated constant for compilation URL. Please, use a single env variable with fallback to “xod.io” if not set.

There is already `utils/siteLinks.js` where you can do something like:

```js

const HOSTNAME = process.env.XOD_HOSTNAME || 'xod.io';

export const compileServiceHostname = `https://compile.${HOSTNAME}`;
export const authServiceHostname = `https://auth.${HOSTNAME}`;
// etc
```

In that case, all entry point URL stuff would be in a single place and easily customizable. `siteLinks` is not the best name if we manage service URL’s too, but it could be easily renamed to something like `urls.js`

packages/xod-client-electron/src/upload/components/PopupUploadConfig.jsx
nkrkv: From a user point of view, we should disable if offline or quota exhaust. If it is trivial, please, do it.

brusherru: So, if service is offline — can we try to compile in the cloud? :)
I think, no, cause it meaningless.
But we should retry to connect, to turn it on when service becomes available again.

packages/xod-client/src/user/actions.js
```diff
@@ -0,0 +1,16 @@
...
+    ...url.parse(process.env.XOD_COMPILE_DOMAIN || XOD_COMPILE_DOMAIN),
+    pathname: '/limits',
+  }))
+    .then(res => (res.ok ? res.json() : null))
+    .then(limit => dispatch({
+      type: UPDATE_COMPILE_LIMIT,
+      payload: limit,
+    }));
+};
+export default {};
```
nkrkv: Do not export default nothing.

packages/xod-client/src/user/constants.js
```diff
@@ -0,0 +1,2 @@
+export const XOD_COMPILE_DOMAIN = 'https://compile.xod.io';
+export default {};
```
nkrkv: Ditto

packages/xod-client/src/user/actions.js
brusherru: But eslint throws errors when using not default for a single export. In the future, it will be easier to add new exports and remove empty default than to change the imports everywhere. These files will grow soon (login, account details and etc).

@nkrkv what do you think?

brusherru: Or maybe use `eslint-disable`?

nkrkv: Ah, OK, a pattern. No problem.

alexander-matsievsky:
@brusherru @nkrkv ok?

packages/xod-client-electron/src/upload/components/PopupUploadConfig.jsx
```diff
@@ -273,7 +274,10 @@ class PopupUploadConfig extends React.Component {
             onChange={this.onCloudCompilationChanged}
```
brusherru: Maybe add here `enabled={!!compileLimitLeft}`? :)

packages/xod-client/src/user/actions.js
```diff
@@ -0,0 +1,10 @@
+import { UPDATE_COMPILE_LIMIT } from './actionTypes';
+
+export const updateCompileLimit = COMPILE_HOSTNAME => () => dispatch =>
+  fetch(`${COMPILE_HOSTNAME}/limits`)
```
brusherru: Still no error handling and no retries...

alexander-matsievsky: @brusherru No error handling is required for this call.

alexander-matsievsky: @brusherru Should the retries be the responsibility of this component as opposed to the app?

brusherru: And I'm not sure about this URL composing.

I'd rather use an argument as it passed...
And in the container — don't pass just a constant, but call `getCompileLimitsUrl` from `siteLinks.js` (probably renamed into `urls.js`).

brusherru: But no one handles errors. So if something went wrong and `fetch` will reject — it will be unhandled rejection... and there will be no reaction for the action

brusherru: About retries...
In case that this function fetches data and then updates state — I think it could do retries too.
Not unlimited retries, of course.
But app should call this function later if the state still contains `null` limits. Maybe when a user clicks on `upload` menu item.
IMHO.
What do you think?

nkrkv:
Hmmm… I can’t test it. Trying to use a custom hostname fails. It is because env variable is not passed like `NODE_ENV` does:

```
packages/xod-client-electron/webpack.config.js
110:      'process.env.NODE_ENV': JSON.stringify(process.env.NODE_ENV || 'development'),

packages/xod-client-browser/webpack.config.js
96:      'process.env.NODE_ENV': JSON.stringify(process.env.NODE_ENV || 'development'),
```

PR #837, opened 2017-10-25
Nodes and fixes for easier state machine creation

### LINKED ISSUE #828: Values bound to inputs leak to outputs
### Steps to reproduce

1. Create a program with a `xod/common-hardware/led` linked to `xod/common-hardware/button`
2. Upload to a board

### Expected behavior

* The LED is initially off
* The LED reflects a state of the button, i.e. switches on when and only when the button is being pressed

### Actual behavior

* The LED is initially **on**
* After the first button press, all goes fine

### Details

In generated C++ code we could see

```cpp
   // Storage of #11 xod/core/constant_number
    xod__core__constant_number::Storage storage_11 = {
        { }, // state
        0.02 // output_VAL
    };

   ...

   // Storage of #14 xod/core/debounce_boolean
    xod__core__debounce_boolean::Storage storage_14 = {
        { }, // state
        0.02 // output_OUT
    };
```

Note that the initial value for `debounce-boolean` used inside `button` implementation is `0.02`. However, it should be `false`. The value `0.02` has leaked from `Ts` input of the `debounce-boolean` where it is bound.

Looks like the root of a problem is in `flatten.js`:

```js
const findBoundValue = R.uncurryN(2)(
  findFn => R.compose(
    Maybe,
    R.unless(
      R.isNil,
      R.compose(R.head, R.values) // because terminal nodes always have only one bound value
    ),
    findFn(R.complement(R.isEmpty)),
    R.map(Node.getAllBoundValues),
  )
);
```

The expectation of the comment is wrong. In a function `getValueTopBind` that function is called for a non-terminal node, i.e. `debounce-boolean`. And as a result, the `head` of `Node.getAllBoundValues` happens to be the value of `Ts`, not the value of `false` for `OUT` output.

evgenykochetkov:
- Closes #828
- Important fix for `defer-*` nodes(6e2643e)
- Adds a bunch of stuff that was not described in issues 😬

Would much appreciate feedback about
- naming and descriptions of new nodes
- whether comments in ad58fed are clear enough


workspace/__lib__/xod/core/clock/patch.xodp
```diff
@@ -37,10 +40,22 @@
...
+        "y": 204
       },
       "type": "xod/patch-nodes/output-pulse"
+    },
+    {
+      "boundValues": {
+        "__out__": true
+      },
+      "id": "r1jzp_DTZ",
+      "label": "EN",
```
nkrkv: I forgot a description. Proposal:

“Is the clock enabled, i.e. produces ticks? At the moment when set to true, starts counting from scratch.”

packages/xod-project/src/flatten.js
```diff
@@ -395,7 +412,26 @@ const getValueToBind = R.curry((nodesById, linksChain) => {
...
+    //    | | count | ... |
+    //    | +-O-----+     |
+    //    |   |           |
+    //    +---O-----------+
+    //        |\
+    //        | \--- here some initial value is bound
+    //        |
+    //    +---O-----------+
+    //    |  ...          |
+    //    +---------------+
```
nkrkv: :heart:

workspace/lib/xod/core/continuously-pausable/any.cpp
```diff
@@ -0,0 +1,15 @@
...
+
+void evaluate(Context ctx) {
+  bool isEnabled = getValue<input_EN>(ctx);
+
+  if (isInputDirty<input_EN>(ctx) && !isEnabled) {
+    clearTimeout(ctx);
+  } else if (isTimedOut(ctx) && isEnabled) {
+    emitValue<output_TICK>(ctx, 1);
+    setTimeout(ctx, 0);
+  }
```
nkrkv: Not sure, but isn’t a simpler (?) alternative is equivalent?

```cpp
bool isEnabled = getValue<input_EN>(ctx);
if (isEnabled) {
    if (isTimedOut(ctx))
        emitValue<output_TICK>(ctx, 1);
    setTimeout(ctx, 0);
} else {
    clearTimeout(ctx);
}
```

Eliminates dirtiness check.

workspace/lib/xod/core/continuously-pausable/patch.xodp
```diff
@@ -0,0 +1,39 @@
...
+        "x": 34,
+        "y": 204
+      },
+      "type": "xod/patch-nodes/output-pulse"
+    },
+    {
+      "boundValues": {
+        "__out__": true
+      },
+      "description": "Determines if values should be continuously emitted",
```
nkrkv: A typo? … *pulses* should be continuously emitted…

workspace/lib/xod/core/gate-pulse/patch.xodp
```diff
@@ -0,0 +1,41 @@
+{
+  "nodes": [
+    {
+      "boundValues": {
+        "__out__": true
+      },
+      "description": "If values should pass through",
```
nkrkv: If → Whether

Missing the patch-level description

workspace/lib/xod/core/continuously-pausable/any.cpp
evgenykochetkov: Yes!
I think we can even drop `isTimedOut` check and emit a pulse at the very moment EN is set to true.

nkrkv: Indeed! :+1:

tools/update-transpiled-fixtures.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+function update_cpp_fixture {
+  node ./packages/xod-cli/bin/xodc.js transpile --output=${1}/__fixtures__/arduino.cpp ${1} @/main
+}
+
+update_cpp_fixture ./workspace/blink
+update_cpp_fixture ./workspace/count-with-feedback-loops
+update_cpp_fixture ./workspace/lcd-time
+update_cpp_fixture ./workspace/two-button-switch
```
nkrkv: :+1:

PR #839, opened 2017-10-26
New view-based strings

nkrkv:
Burn old clunky list!

This is the second attempt to implement container data structures which are XOD-specifics aware, i.e. immutable, with very little RAM, and limited heap-memory features.

Here outcomes:

- The new container classes themselves based on views over the static data (less magic)
- Data chunks not copied anymore for intermediate/processing list nodes
- Iteration interface is backward compatible
- Memory allocation is only done on iteration (few bytes allocation for an iterator, freed back right after iteration)
- Multiple string concatenation bugs seem to go away

Other things involved:

- Refactoring: move C++ data types and values formatting from data generator to the renderer
- CONTINUOUSLY and BOOT are no longer in the generated code
- Make a tool to update `.bin` and `.hex` fixtures (based on PlatformIO)

Better review commit-by-commit as they’re clearly split.


packages/xod-arduino/platform/stl.h
```diff
@@ -0,0 +1,20 @@
+/*=============================================================================
+ *
+ *
+ * STL shim. Provides implementation for vital std:: constucts
```
brusherru: "constucts" -> "constructs"?

PR #842, opened 2017-10-27
Add if-else-string and concat-n nodes

nkrkv:
The PR brings few trivial but often requested nodes:

- `xod/core/if-else-string`
- `xod/core/concat-3`
- `xod/core/concat-4`
- `xod/core/concat-5`
- `xod/core/concat-6`

Hope their names are self-explanatory.


nkrkv:
@evgenykochetkov Ouch, they were, but I forgot to save obviously. Nice catch, thanks!

workspace/__lib__/xod/core/concat-3/patch.xodp
```diff
@@ -0,0 +1,109 @@
...
+      },
+      "type": "xod/patch-nodes/input-string"
+    },
+    {
+      "id": "r1c4Ajg0Z",
+      "position": {
+        "x": 0,
+        "y": 204
+      },
+      "type": "@/concat"
```
evgenykochetkov: Should those be 'absolute' paths?

nkrkv: Nope. If they would renaming the library will break things. `@` means “this project/library”, so this is perfectly valid and more correct than referring by an absolute path.

PR #843, opened 2017-10-27
Fix resizing of project browser on window resize

### LINKED ISSUE #840: Project browser does not resize on window resize
Perhaps introduced by #833 

### Steps to reproduce

1. Launch IDE (browser or desktop)
2. Make the IDE window larger

### Expected behavior

The project browser takes all vertical space from top to the Inspector.

### Actual behavior

The project browser height stays fixed, smaller than it could be

brusherru:
It fixes #840.
It's a bit hacky (using `calc` in the CSS), but to avoid using `calc` we have to rewrite few components and make a HOC from FocusTrap... Cause that is a large amount of work, I fixed it this way.

If you have any ideas how to avoid using `calc` without big refactoring — feel free to write a comment :)

### DIFF
--- a/packages/xod-client/src/core/styles/components/PatchGroup.scss
+++ b/packages/xod-client/src/core/styles/components/PatchGroup.scss
@@ -55,3 +55,7 @@
     border-bottom: 1px solid $sidebar-color-border;
   }
 }
+
+.PatchGroup:first-child .PatchGroup__trigger {
+  border-top: 0;
+}
--- a/packages/xod-client/src/core/styles/components/PatchTypeSelector.scss
+++ b/packages/xod-client/src/core/styles/components/PatchTypeSelector.scss
@@ -1,4 +1,6 @@
 .PatchTypeSelector {
+  height: calc(100% - 30px); // 30px is a height of .ProjectBrowserToolbar
+
   .options {
     display: flex;
     list-style: none;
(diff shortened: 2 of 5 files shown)


nkrkv:
> It's a bit hacky (using calc in the CSS), but to avoid using calc we have to rewrite few components

I think it’s OK. Seems like all modern browsers [support it](https://developer.mozilla.org/en-US/docs/Web/CSS/calc). What I can’t understand quickly is the magic `px` values:

- `height: calc(100% - 30px);` — why 30? I see a gap between the browser and inspector. For me 19px is pixel-perfect, but I don’t understand why 19.
- `heightRelativeToParent="calc(100% - 39px)"` — why 39?
- `height: calc(100% - 1px);` — why 1? Border, no?

Would you provide comments for the magic values and/or extract them into SCSS constants. Let them be  near the context, but their name could help to understand what’s going on.

brusherru:
30px — height of projectBrowserToolbar. (But why 19px is perfect for you — I don't know 😬 )
39px — height of options to switch patch types in the project browser
1px — border between project browser and inspector 😬 

I'll add it to the scss, but what about 19px? It looks like it depends on a platform or something else...

nkrkv:
> 39px — height of options to switch patch types in the project browser

@brusherru Aha! Do you have their title wrapped on two lines? I’m not and for me, its height is exactly 30px. We’ve found 9px, 2px left… maybe they’re borders. Maybe something other.

Can you enforce no line break at “My patches” and adjust the values accordingly?

PR #846, opened 2017-10-27
Don’t allow manually placing xod/patch-nodes/*-dead nodes

### LINKED ISSUE #841: Dead terminals should not expose to a user
### Rationale

Now, oh wonder, I can put `output-dead`, `input-dead` nodes on my patch. I can find them in quick search too.

Intuitively I can’t predict what it would lead to and placing them explicitly makes no sense. Could we filter them out on UI level?

### How to fix

Right before presenting a list of nodes to a user (in the project browser or the suggester), run an additional “client” filter (lives in `xod-client`) to filter out the dead stuff.

evgenykochetkov:
Closes #841

### DIFF
--- a/packages/xod-client/src/project/utils.js
+++ b/packages/xod-client/src/project/utils.js
@@ -8,7 +8,6 @@ import {
   calcutaleNodeSizeFromPins,
   calculatePinPosition,
 } from './nodeLayout';
-import { getProject } from './selectors';
 
 import { LINK_ERRORS } from '../editor/constants';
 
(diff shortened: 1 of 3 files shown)


packages/xod-client/src/project/utils.js
```diff
@@ -129,3 +128,8 @@ export const patchToNodeProps = (patch) => {
   };
 };
 
+export const isPatchDeadTerminal = R.compose(
```
brusherru: 🤔 
Shouldn't this function be in a `xod-project`?

PR #850, opened 2017-10-30
Add multiple selection with marquee

### LINKED ISSUE #820: As a xoder I want to select multiple items with marquee because that’s intuitive
### Rationale

We have multiple selection, but not a rectangular marquee. The UX will improve much if there will be such feature. It would let making patches faster.

The question which often arises is “Should I include items only they’re completely inside the rectangle or touching it is enough”. The problem solved wisely in QCad. If you grow the marquee left-to-right only fully covered items are included, and when you grow right-to-left even touched items are included.

qcad-marquee

Note how a border of the rectangle differs in two modes.

### Acceptance criteria

- [x] I can select multiple items with marquee with mechanics described
- [x] I can hold Ctrl at the moment of mouse button release and the selection would be added to the existing

brusherru:
It closes #820.

Preview
On preview:
Black circle — just left click
Red circle — left click + cmd (or ctrl)

### DIFF
--- a/packages/xod-client/src/editor/containers/Patch/modes/marqueeSelecting.jsx
+++ b/packages/xod-client/src/editor/containers/Patch/modes/marqueeSelecting.jsx
@@ -1,0 +1,200 @@
+import R from 'ramda';
+import React from 'react';
+import { HotKeys } from 'react-hotkeys';
+import className from 'classnames';
+
+import { COMMAND } from '../../../../utils/constants';
+import { getNewSelection } from '../../../utils';
+
+import PatchSVG from '../../../../project/components/PatchSVG';
+import * as Layers from '../../../../project/components/layers';
+
+import {
+  bindApi,
+  getMousePosition,
+  getOffsetMatrix,
+} from '../modeUtils';
+
+import {
+  isInclusiveSelection,
+  getSelectionBox,
+  filterLinksByBox,
+  filterLinksByInclusiveBox,
+  filterNodesByInclusiveBox,
+  filterNodesByBox,
+} from '../../../marqueeGeometry';
+
+// =============================================================================
+//
+// Component Utils
+//
+// =============================================================================
+
+const isSelectionModifierPressed = event => event.metaKey || event.ctrlKey;
+
+// New, dependent on arguments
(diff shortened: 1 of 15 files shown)


packages/xod-client/src/editor/containers/Patch/modes/marqueeSelecting.jsx
```diff
@@ -0,0 +1,218 @@
...
+      isMouseDown: true,
+      dragStartPosition,
+      mousePosition,
+      selection: props.selection,
+    };
+  },
+  onMouseMove(api, event) {
+    api.setState({
+      mousePosition: getMousePosition(patchSvgRef, api.props.offset, event),
+      selection: getComputedSelection(api, event),
```
evgenykochetkov: Looks like selection is calculated with previous mouse position

packages/xod-client/src/editor/containers/Patch/modes/marqueeSelecting.jsx
```diff
@@ -0,0 +1,218 @@
...
+      <rect
+        className={cls}
+        x={from.x}
+        y={from.y}
+        width={width}
+        height={height}
+      />
+    );
+  },
+  renderBlockingLayer(api) {
```
evgenykochetkov: A comment about what exactly this layer blocks(patch entities from mouse events?) would be nice

packages/xod-client/src/editor/containers/Patch/modes/selecting.jsx
```diff
@@ -39,25 +40,31 @@ const selectingMode = {
...
       props.actions.addEntityToSelection(entityType, entityId);
       setState({
+        isMouseDown: true,
         isMouseDownOnMovableObject: true,
         dragStartPosition: mousePosition,
       });
     } else {
       props.actions.selectEntity(entityType, entityId);
       setState({
+        isMouseDown: true,
```
evgenykochetkov: I don't think all those `isMouseDown`s are necessary. Maybe, a single `isMouseDownOnBackground` is clearer?

brusherru:
@evgenykochetkov @nkrkv check it out, please :)

PR #851, opened 2017-10-30
If pasted entities are the same as selected, paste them with just a little offset

### LINKED ISSUE #793: If I copy then paste on the same patch I expect the pasted cluster to be at the same position as original with a little offset
This is a tweak requested in https://github.com/xodio/xod/pull/792#issuecomment-333473893

### Steps to reproduce

Select some entities, copy and immediately paste

### Expected behavior

Entities are pasted with a slight offset to current selection(X ~ 1 pin width, Y ~ half of the node height)

### Actual behavior

Entities are pasted to viewport origin

evgenykochetkov:
Closes #793

### DIFF
--- a/packages/xod-client/src/editor/actions.js
+++ b/packages/xod-client/src/editor/actions.js
@@ -35,13 +35,14 @@ import {
 
 import {
   getBBoxTopLeftPosition,
+  getBoundingBoxSize,
   getEntitiesToCopy,
   getSelectedEntityIdsOfType,
   regenerateIds,
   resetClipboardEntitiesPosition,
 } from './utils';
 import { isInput, isEdge } from '../utils/browser';
-import { addPoints, NODE_HEIGHT, SLOT_SIZE } from '../project/nodeLayout';
+import { addPoints } from '../project/nodeLayout';
 
 import { ClipboardEntities } from '../types';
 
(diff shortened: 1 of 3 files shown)


packages/xod-client/src/editor/actions.js
```diff
@@ -285,8 +289,19 @@ export const copyEntities = event => (dispatch, getState) => {
...
+  return Maybe.Just(entities);
+};
+
+export const copyEntities = event => (dispatch, getState) => {
+  if (isInput(event)) return;
+
+  const state = getState();
+
+  // eslint-disable-next-line array-callback-return
+  getClipboardEntities(state).map((entities) => {
```
nkrkv: You don’t use the `map` result. Linter is not very happy. Perhaps, it is a job for `forEach`?

nkrkv:
Uh, that’s not the best we can do. Sorry, I did not describe clearly what’s necessary in the issue. See how it works in Google Slides:

copy-paste-multiple

On every paste, the destination position is incremented in the south-east direction, and it keeps this behavior consistent regardless of actions in between. That feels intuitive.

Currently, in XOD we get the same behavior only on the first paste. Starting from the second repeat we still get a mess. Could we do something like in Google Slides?

packages/xod-client/src/editor/actions.js
```diff
@@ -270,23 +278,30 @@ export const highlightSugessterItem = patchPath => ({
...
+
+export const copyEntities = event => (dispatch, getState) => {
+  if (isInput(event)) return;
+
+  const state = getState();
 
-  const entities = getEntitiesToCopy(currentPatch, selection);
-  event.clipboardData.setData(getClipboardDataType(), JSON.stringify(entities, null, 2));
-  event.preventDefault();
+  // linter confuses array and a Maybe
```
nkrkv: Me too :grimacing: 

I’d use the `foldMaybe(addToClipboard, noop, getClipboardEntities(state))`. Doesn’t matter so much though…

PR #869, opened 2017-11-01
Add new line at the end of all xod files

### LINKED ISSUE #859: Add new lines at end of all .xod* files
### Rationale

Having a file terminated with EOL follows best practices. Many text editors enforce this rule on save.

The `*.xod*` files do not have EOL at EOF when saved by IDE. It leads to false diffs when these files are edited manually after that.

We should add EOL at EOF for all XOD files:

- `project.xod`
- `*.xodp`
- `*.xodball`

### Acceptance criteria

- [ ] Files saved by IDE have terminal new line
- [ ] All existing files of the repo are updated

brusherru:
It closes #859.

You can check out only latest commit, cause the first one is a result of the script:
```
for file in $(find . -name project.xod); do sed -i '' -n p $file; done;
for file in $(find . -name *.xodp); do sed -i '' -n p $file; done;
for file in $(find . -name *.xodm); do sed -i '' -n p $file; done;
```
(using [OS X sed](https://stackoverflow.com/a/16198793/7224709))


packages/xod-fs/src/write.js
```diff
@@ -1,5 +1,5 @@
-
-import { curry } from 'ramda';
+import { curry, unless, test, concat, __ } from 'ramda';
```
evgenykochetkov: We import all Ramda functions in other places 🤔

brusherru: Yep, and at first I imported a whole Ramda, but this has led to a larger changeset, which is not relevant to the task (and there was an import of just `curry` function). So I decided to import some needed functions instead of the whole library.
If you think that it is better to import the entire Ramda — it's not a problem.

So, what do you think, @evgenykochetkov?

packages/xod-fs/src/write.js
```diff
@@ -10,7 +10,12 @@ export const writeFile = curry((outputPath, data, encoding) => new Promise(
     const resolvedPath = resolvePath(outputPath);
     const fstream = fileSave(resolvedPath);
 
-    fstream.write(data, encoding);
+    const dataWithEol = unless(
+      test(/\n$/g),
+      concat(__, EOL)
```
nkrkv: I’m pretty sure we should add a strict `\n`, not EOL. Otherwise, it would still produce falsy diffs (of a more strange kind) when one developer is on Windows and another is on Posix. Does it make sense?

brusherru: @nkrkv fixed :)

PR #873, opened 2017-11-02
Make height of Project Browser and Inspector adjustable

### LINKED ISSUE #521: Make height of Project Browser and Inspector adjustable
Currently, the height of Project Browser is fixed at some px value. This makes it very inconvenient (or even impossible) to work with Inspector if window height is quite small.

A simple handle that would let a user drag it to change Project Browser vs Inspector height proportion would help.

And in the case when Inspector contains too much content for its height a scrollbar would also help.

### Acceptance criteria

- [x] I can hover a split line between Project Browser and Inspector and see ↕ cursor
- [x] I can drag the handle to change Project Browser and Inspector height proportion
- [x] I’m limited to some minimal (e.g. 200px) height for either sidebar; the limit is honored on start, window resize and handle drag
- [x] If there is too much content in Inspector I can use its scrollbar to navigate to overflowed content

brusherru:
It closes #521.


brusherru:
@nkrkv oops, you're right! Just fixed it.

PR #875, opened 2017-11-03
Implement text-lcd-16x2-i2c

nkrkv:
It is an often requested node

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,3 @@
+catch.hpp linguist-vendored
+packages/xod-client-electron/arduino-libraries/** linguist-vendored
+workspace/**/__fixtures__/*.cpp linguist-generated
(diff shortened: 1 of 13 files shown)


packages/xod-client-electron/arduino-libraries/LiquidCrystal_I2C/examples/BlinkingCursor/BlinkingCursor.ino
```diff
@@ -0,0 +1,32 @@
+#include <Wire.h> 
+#include <LiquidCrystal_I2C.h>
```
evgenykochetkov: Why include these examples? 🤔

.gitattributes
```diff
@@ -0,0 +1,3 @@
+catch.hpp linguist-vendored
+packages/xod-client-electron/arduino-libraries/** linguist-vendored
+workspace/**/__fixtures__/*.cpp linguist-generated
```
evgenykochetkov: ❤️

packages/xod-client-electron/arduino-libraries/LiquidCrystal_I2C/examples/BlinkingCursor/BlinkingCursor.ino
nkrkv: It’s a clone of the library from GitHub as is. Technically we don’t need them. But should we really modify 3-rd party vendored sources? I think it would complicate maintenance (i.e. library updates).

PR #876, opened 2017-11-07
Format code with prettier

nkrkv:
This is *not* a ready-to-merge PR. It is a preview of my offer to use [prettier](https://prettier.io/). Prettier became very popular last year and I think we could use it to improve DX a bit and make our code look more homogenous.

Please, check what Prettier did with our source files. Do you think it is OK?


nkrkv:
This was a preview. Closing for now

Issue #879, opened 2017-11-08
As a xoder I want to edit nodes’ C++ implementations in IDE so that I can develop native nodes end-to-end without restarts and application switches


nkrkv:
### Rationale

Currently, a pipeline to add/edit a C++ implementation is tiresome. I have to close XOD IDE, find my `.cpp` file or create and name it properly, edit it in a text editor, open XOD IDE, open a project, Upload,… and find I just made a syntax error.

Let the edits to be more comfortable. We can embed a simple code editor for C++ implementations and have them updated right on edit. The cycle would become much shorter.

The mechanics could look like:

- A xoder places a special node `cpp-implementation` to declare the node should use C++ code
- He double-clicks the node to open code editor right inside the tab
- He edits-uploads-edit-uploads-edits-uploads until satisfied
- Then he clicks a back button to switch to the normal tab mode

### Acceptance criteria

- [ ] I can place a `cpp-implementation` node on my patch
- [ ] I can double-click the node to open the code editor inside the patch tab
- [ ] The code editor takes the whole work area; a line at the top of the editor is shown with a label “Editing node C++ implementation” and a back button is shown
- [ ] Once `cpp-implementation` is drilled down, the code editor shows the current C++ implementation if present, a template¹ otherwise
- [ ] The transpiler uses C++ implementation only if a patch has `cpp-implementation` in it
- [x] If I delete the `cpp-implementation` node, C++ itself is kept around across saves/loads and I can enable and edit it again by simply placing `cpp-implementation`
- [ ] On “Upload” the latest implementation code is used even if not saved yet
- [ ] Undo/redo work in the editor in a decent way
- [ ] The editor highlights C++ syntax
- [ ] The editor auto-indents with shift width = 4 spaces
- [x] Implementation files have name `patch.cpp`
- [x] All current `any.cpp` and `arduino.cpp` are renamed at once to the unified `patch.cpp` name

¹ A template for new implementations

```cpp
struct State {
};

{{ GENERATED_CODE }}

void evaluate(Context ctx) {
    //auto inValue = getValue<input_IN>(ctx);
    //emitValue<output_OUT>(ctx, inValue);
}
```

PR #885, opened 2017-11-14
Only one target platform, only one native implementation file — `patch.cpp`

evgenykochetkov:
This is a prerequisite for #879 

- `xod-project`

  - no notion of different impl types
  - because there is only one kind of implementation left, no need to store them in `impls`, it can be extracted straight from attachments

  - API changes:
    - removed:
      - `listImpls`
      - `hasImpls`
      - `getImplByArray`

    - added:
      - `hasImpl :: Patch -> Boolean`
      - `setImpl :: Source -> Patch -> Patch`

    - changed signatures:
      - `getImpl` is now `Patch -> Maybe Source`
      - `hasImpl` is now `Patch -> Boolean`
      - `flatten` is now `Project -> PatchPath -> Either Error Project`

- `xod-fs`
  - does not know about implementation files — they all are just attachments

+ minor changes in other backages


packages/xod-fs/test/fixtures/workspace/__lib__/xod/core/pot/patch.cpp
```diff
@@ -0,0 +1 @@
+// should be imported into impl.cpp
```
nkrkv: What’s this?

packages/xod-fs/test/load.spec.js
```diff
@@ -99,28 +96,11 @@ describe('Loader', () => {
...
-          });
-        }).then(
-          implsLoaded => assert.equal(implsLoaded, true, 'some implementations were not loaded')
-        );
-    });
+    it('should return project without libs', () =>
+      Loader.loadProjectWithoutLibs(
+        path.resolve(workspace, './__lib__/xod/core')
+      )
+    );
```
nkrkv: Looks like it no longer tests anything

packages/xod-fs/test/fixtures/workspace/__lib__/xod/core/pot/patch.cpp
evgenykochetkov: Oops. It was an outdated placeholder. Fixed!

packages/xod-project/src/attachment.js
```diff
@@ -0,0 +1,41 @@
...
+  (filename, encoding, content) => ({
+    filename,
+    encoding,
+    content,
+  })
+);
+
+export const createImplAttachment = def(
+  'createImplAttachment :: Source -> Attachment',
+  content => ({
```
brusherru: Better reuse `createAttachment`

packages/xod-project/src/attachment.js
```diff
@@ -0,0 +1,41 @@
+import R from 'ramda';
+import { def } from './types';
+import { IMPL_FILENAME } from './constants';
+
+export const createAttachment = def(
+  'createImplAttachment :: String -> String -> String -> Attachment',
```
brusherru: Wrong function name in the signature

packages/xod-project/src/flatten.js
```diff
@@ -746,29 +752,31 @@ export const extractPatches = R.curry((project, leafPaths, prefix, boundValues,
...
-      Patch.upsertNodes(newNodes)
-    )(Patch.createPatch());
-  }
-)(patch));
+const convertPatch = def(
+  'convertPatch :: Project -> [Pair PatchPath Patch] -> Patch -> Patch',
+  (project, leafPatches, patch) =>
+    R.ifElse(
+      Patch.hasImpl,
+      R.identity,
```
brusherru: `R.ifElse` could be replaced with `R.unless` in this case.

packages/xod-project/src/attachment.js
evgenykochetkov: Nice catch, fixed!

packages/xod-project/src/attachment.js
evgenykochetkov: Done

packages/xod-project/src/flatten.js
evgenykochetkov: Done

evgenykochetkov:
@nkrkv @brusherru I refactored the test in `xod-func-tools` that was stopping the build, check it out please

nkrkv:
Yeeee. Not a huge fan of wasting 100 ms to just wait… but in the scale of our whole test suite, it is OK.

PR #890, opened 2017-11-16
Integrate editor for C++ implementations

### LINKED ISSUE #879: As a xoder I want to edit nodes’ C++ implementations in IDE so that I can develop native nodes end-to-end without restarts and application switches

evgenykochetkov:
Closes #879


packages/xod-client/package.json
```diff
@@ -36,13 +37,14 @@
...
     "react-click-outside": "^2.2.0",
+    "react-codemirror": "^1.0.0",
     "react-collapsible": "git+https://github.com/xodio/react-collapsible.git#npm",
     "react-custom-scroll": "^2.0.0",
     "react-dnd": "^2.5.1",
     "react-dnd-html5-backend": "^2.5.1",
     "react-dom": "^15.4.2",
     "react-event-listener": "^0.4.5",
-    "react-fa": "git+https://github.com/xodio/react-fa.git",
+    "react-fa": "^5.0.0",
```
nkrkv: :+1:

packages/xod-client/src/editor/reducer.js
```diff
@@ -468,6 +468,25 @@ const editorReducer = (state = {}, action) => {
...
+          type: TAB_TYPES.IMPLEMENTATION_CODE,
+          isReadOnly: action.payload.isReadOnly,
+        }),
+        state
+      );
+    case EAT.EDITOR_CLOSE_IMPLEMENTATION_CODE:
+      return R.over(
+        currentTabLens,
+        R.compose(
+          // TODO: what if we are in debug mode?
```
nkrkv: Ideally, we should be switched back to the tab/place/area where we initiated the code editor opening.

Not a big deal though.

packages/xod-client/src/editor/components/CppImplementationEditor.jsx
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import CodeMirror from 'react-codemirror';
+import 'codemirror/mode/clike/clike';
+
+const CppImplementationEditor = ({ isActive, source, isReadOnly, onChange, onClose }) => {
+  const options = {
+    lineNumbers: true,
+    readOnly: isReadOnly,
```
brusherru: Is it really necessary in case that we can edit libraries?
@nkrkv, what will you say? :)

brusherru: Also, in `readOnly` mode it looks the same. And there is no explanation why User can't edit implementation.... Could be recognized as a bug.

nkrkv: @evgenykochetkov @brusherru Agree. It’s better to always allow editing.

PR #892, opened 2017-11-16
Tweak pin numbering scheme, add time-signal generation nodes

nkrkv:
The nodes introduced are listed in the commit message body.

### DIFF
--- a/packages/xod-project/src/pin.js
+++ b/packages/xod-project/src/pin.js
@@ -229,19 +229,19 @@ const getPinLabelByDirection = def(
 
 /**
  * Returns a list of Pins with unique Pin Labels.
- * It adds a suffix "_N" (where N is a counter of equal labels)
- * to the label of each Pin.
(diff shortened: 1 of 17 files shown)


nkrkv:
@brusherru Did you connect the LED to a PWM-compatible port (the one marked with ~ on the board)?

nkrkv:
@brusherru Ah, found the problem. It is in `analogWrite` or Arduino O_o. The following trivial sketch leads to the same glitch:

```cpp
uint8_t counter;

void setup() {
  pinMode(13, OUTPUT);
}

void loop() {
  analogWrite(13, ++counter);
  delay(60000 / 255); // ~ 60 seconds fade in
}
```

The issue is already filed:

- https://github.com/arduino/Arduino/issues/2735
- http://forum.arduino.cc/index.php?topic=303922.0

So… it the problem has nothing to do with wave-* nodes.

PR #893, opened 2017-11-17
Don’t propagate bound values to occupied outputs

evgenykochetkov:
No issue 😬

See [this xodball](https://gist.github.com/evgenykochetkov/ce828868a657f7bfd529631a96f88cac) for a minimal example

### DIFF
--- a/packages/xod-project/src/flatten.js
+++ b/packages/xod-project/src/flatten.js
@@ -1,7 +1,7 @@
 import R from 'ramda';
 import { Maybe, Either } from 'ramda-fantasy';
 
-import { explode, explodeEither, catMaybies } from 'xod-func-tools';
+import { explode, explodeMaybe, explodeEither, catMaybies } from 'xod-func-tools';
 
 import * as CONST from './constants';
 import * as Project from './project';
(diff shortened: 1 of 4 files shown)


packages/xod-project/src/flatten.js
```diff
@@ -437,14 +437,26 @@ const getValueToBind = R.curry((nodesById, linksChain) => {
...
+      .chain((bottomBoundValue) => {
+        const pinKey = Link.getLinkOutputPinKey(R.last(linksChain));
+        const topNode = R.last(outputNodesFromLinkChain);
+        const maybeBoundValue = Node.getBoundValue(pinKey, topNode);
+
+        // if some value is already bound to the
+        // top node's output — don't overwrite it
+        return maybeBoundValue.isJust
+          ? Maybe.Nothing()
+          : Maybe.Just(bottomBoundValue);
```
nkrkv: Use `foldMaybe`? It would survive possible (?) `Maybe` lib replacement.

evgenykochetkov: Done 👍🏻

evgenykochetkov:
🤔
It's not so simple.
We also need to consider wether the processed terminal nodes were input-terminals or output-terminals before conversion to `xod/internal/terminal-*`

evgenykochetkov:
Updated the fix.
@nkrkv @brusherru check it out again please

PR #895, opened 2017-11-20
Add sd-log node

nkrkv:
The PR consist of several commits:

- d818d2b — 3-rd party library vendoring, skip in review
- 4e0bd26, 81b0246 — the essence
- b8e1441 — fix indentation in CodeMirror (the acceptance criteria was not carefully validated by anyone :hushed:)


nkrkv:
@brusherru A very common case for SD card is creating a data logger, a widget that appends to a CSV file line by line data from few sensors every X seconds. The `sd-log` node solely serves this purpose and covers many cases.

Other things would be useful while developing IoT applications, but XOD is not powerful enough to complete them: we need the polymorphism to introduce the Internet nodes *and* do things like `sd-read`, `sd-mkdir` in a consistent way.

PR #898, opened 2017-11-21
Save only changes of projects/libraries

### LINKED ISSUE #231: Patch rename should delete the patch with old name
This is a particular case of a more complex problem. Orphan patches stay on FS when we update libraries as well

### Steps to reproduce
- Create new project
- Create new patch with name `hello`
- Save patch
- Rename the patch to `aloha`
- Save patch
### Expected behavior
- There is `aloha` directory on FS
- There is no `hello` directory on FS
### Actual behavior
- There is `aloha` **and** `hello` directory on FS

### How to fix

Implement a system of incremental project saves. Rough steps outline when save is triggered.

1. For each library referred, check if it’s directory is already in `__lib__`. If not, spawn the library there completely.
2. Compare the project about to be saved with the project at the last save.
3. For each diff on a patch level, overwrite the patch over existing files in FS (if modified in diff), delete the patch directory (if removed in diff)

Make the mechanism a part of `xod-fs` so that it can be later reused by `xod-cli`

brusherru:
It will fix #231 and some more issues that discussed.


brusherru:
@nkrkv @evgenykochetkov just updated! Check it out, please

packages/xod-client/src/utils/menu.js
```diff
@@ -22,7 +22,7 @@ const rawItems = {
   },
   saveProject: {
     key: 'saveProject',
-    label: 'Save Project',
+    label: 'Save All',
     command: COMMAND.SAVE_PROJECT,
```
nkrkv: Should we also rename the command: `SAVE_PROJECT` → `SAVE_ALL`? So that no one could be confused.

packages/xod-fs/src/patchDiff.js
```diff
@@ -0,0 +1,68 @@
...
+  R.compose(
+    R.map(createPatchChange(CHANGE_TYPES.DELETED)),
+    R.differenceWith(isEqualPatchPaths)
+  )
+);
+
+export const calculateDiff = def(
+  'calculateDiff :: [Patch] -> [Patch] -> [AnyPatchChange]',
+  R.converge(
+    (...args) => R.unnest(args),
```
evgenykochetkov: Could replace with `R.unapply(R.unnest)`, but it does not matter :)

Issue #874, opened 2017-11-02
As a xoder I want all projects I open in IDE to resolve dependencies automatically


nkrkv:
### Rationale

This is a #871 in IDE scale.

### Acceptance criteria

- [ ] Once I open a project in an IDE (browser or electron) I have dependency installed automatically
- [ ] Visually it looks like I add all libraries with “File → Add Library”. The search dialog is not shown, but the project browser behaves the same way

Issue #884, opened 2017-11-14
Consistent xodc CLI


nkrkv:
Currently, `xodc` is not usable as it could be. In some points it violates the very basic command line tools standards, poorly validates command arguments, etc. Let’s make `xodc` great again.

I suggest using [yargs](https://github.com/yargs/yargs) as the complete command-line argument parsing solution. What we’re going to achieve:

```
$ xodc --help
Usage: xodc [command] [options]

Commands:
  resave    opens a project and saves it in another location or format
  install   install libraries or toolchains
  publish   publish a library
  transpile transpile (generate C++) a XOD program
  boards    list supported board names/identifiers
  compile   compile (verify) a XOD program
  upload    upload a XOD program to the board
  help      show help for a specific command

Options:
  -h, --help      show help
  -V, --version   show the version number
  -w, --workspace use the workspace specified, defaults to $HOME/xod
  -q, --quiet     do not log messages other than errors

Run `xodc help COMMAND`  for more information on specific commands.
```

### resave ✔️

```
$ xodc resave --help
Usage: xodc resave [options] [project]

Options:
  -h, --help      show help
  -V, --version   show the version number
  -w, --workspace use the workspace specified, defaults to $HOME/xod
  -q, --quiet     do not log messages other than errors
  -o, --output    xodball or multifile directory output path, defaults to stdout

Examples:
  xodc resave . -o ~/foo.xodball             Exports the current multifile project to
                                             a xodball
  xodc resave                                Outputs the current multifile project as
                                             a xodball to stdout
  xodc resave foo.xodball -o bar.xodball     Resaves one xodball into another (useful
                                             for applying migrations)
  xodc resave foo.xodball -o /some/new/dir   Converts a xodball to a multifile project
```

### install

```
$ xodc install --help
Usage: xodc install [options] [lib|arch]

Options:
  -h, --help      show help
  -V, --version   show the version number
  -w, --workspace use the workspace specified, defaults to $HOME/xod
  -q, --quiet     do not log messages other than errors
  --api           XOD API hostname (default: xod.io)

Examples:
  xodc install                              Install all libraries the current
                                            project depends on.
  xodc install bob/utils                    Add/install the `bob/utils`
                                            library
  xodc install bob/utils@0.3.0              Add/install library at the
                                            specific version
  xodc install arduino:sam                  Intall the toolchain for the SAM
                                            architecture of the Arduino package
```

### publish ✔️

```
$ xodc publish --help
Usage: xodc publish [options] [project]

[...]

PR #901, opened 2017-11-23
Resolve dependencies when installing libraries using cli-tools

### LINKED ISSUE #871: As a CLI pro-xoder I want library dependencies to be resolved automatically so that I can bootstrap projects without a hassle
### Rationale

If a project contains nodes from other libraries and I haven’t yet installed those libraries (and dependencies or that libraries recursively) by hand I would get dead references. It’s not a user-friendly behavior.

We can recursively check for used libraries by enumerating all patch paths and have them installed automatically.

### Acceptance criteria

- [ ] I can add and use few nodes from a third-party library `bob/tools` in a project, then export the project
- [ ] Then I can switch to a clean workspace and import the project
- [ ] The `bob/tools` gets installed automatically on project import along with its transitive dependencies
- [x] I can run `xodc install` being somewhere under workspace directory and have all dependencies of all projects to be resolved and installed

### How to implement

- Create a `xod-pm` package for dependency resolving. Later, it will be used by `xod-client-electron` and `xod-client-browser` so `xod-pm` should not deal with FS.
- While resolving transient dependencies, put all downloaded stuff into a swap directory to guarantee an atomic update

brusherru:
It will close #871.

Some of the acceptance criteria (3/4) is undone, cause it relates to IDE story (issue #874).
If you think it should be done, I could complete it in the same PR, but it will close both issues :)
It's just about integration the same function into `xod-client` and `xod-client-electron`.


packages/xod-pm/src/fetch.js
```diff
@@ -87,12 +95,39 @@ export const fetchLibXodball = R.curry((swaggerUrl, libQuery) =>
...
         ),
         parseLibQuery
       )(libQuery)
     )
 );
+
+// :: URL -> LibName -> Promise [Project] Error
+export const fetchLibrary = R.curry((swaggerUrl, libQuery) => {
+  // :: [LibName] -> [Project] -> [LibName] -> Promise [Project] Error
+  const fetchReqursively = R.curry((fetchedLibNames, fetchedLibs, libNamesToFetch) => {
```
evgenykochetkov: Probably, instead of storing libs and lib names separately, it would be more convenient to store them in an object with lib names as keys.

packages/xod-pm/src/fetch.js
```diff
@@ -69,8 +77,8 @@ export const fetchLibData = R.curry((swaggerUrl, libQuery) =>
     )
 );
 
-// :: URL -> LibName -> Promise Xodball
-export const fetchLibXodball = R.curry((swaggerUrl, libQuery) =>
+// :: URL -> LibName -> Promise Project Error
+export const fetchOneLibrary = R.curry((swaggerUrl, libQuery) =>
```
evgenykochetkov: I would rename `fetchOneLibrary/fetchLibrary` pair to something like `fetchLibrary/fetchLibraryAndDependencies`

packages/xod-pm/src/fetch.js
brusherru: Nice idea! 👍

packages/xod-func-tools/src/monads.js
```diff
@@ -116,6 +116,17 @@ export const eitherToPromise = foldEither(
...
 
+// :: (() -> b) -> (a -> a) -> Maybe a -> Promise a b
+export const maybeToPromise = R.curry(
+  (nothingFn, justFn, maybe) => new Promise(
+    (resolve, reject) => (
+      maybe.isJust ? resolve(maybe.value) : reject()
+    )
+  )
+    .catch(nothingFn)
+    .then(justFn)
```
evgenykochetkov: `justFn` will always run, even if `maybe` is a `Just`

packages/xod-func-tools/test/monads.spec.js
```diff
@@ -125,6 +126,26 @@ describe('moands', () => {
...
+      maybeToPromise(
+        () => new Error('It should not been thrown'),
+        a => a,
+        Maybe.Just(52)
+      ).then(val => assert.equal(val, 52))
+    );
+    it('returns rejected promise', () =>
+      maybeToPromise(
+        () => new Error('It is Nothing!'),
+        a => a,
```
evgenykochetkov: please add a test where `justFn` is not an identity function

packages/xod-func-tools/src/monads.js
brusherru: You mean is a `Nothing`? :)
Yep, if in `nothingFn` we return resolved promise...
I don't know how to handle it in another way (calling them inside `resolve()` or `reject()` functions makes nested promises 🤢 ).

evgenykochetkov: > You mean is a Nothing? :)
Yes, sorry :)

evgenykochetkov: > Yep, if in nothingFn we return resolved promise...

screen shot 2017-11-23 at 19 53 17

packages/xod-func-tools/test/monads.spec.js
brusherru: Fixed! Thanks for the catch.

nkrkv:
@brusherru The three items are not related to IDE. They mean exporting/importing with CLI. Until #884 is done, the commands are called `pack` and `unpack` AFAIK.

I think that’s not a big problem if one would have to explicitly `xodc install` after `xodc unpack`, so indeed we can omit these items.

packages/xod-func-tools/src/monads.js
```diff
@@ -116,6 +116,15 @@ export const eitherToPromise = foldEither(
   Promise.resolve.bind(Promise)
 );
 
+// :: (() -> b) -> (a -> a) -> Maybe a -> Promise a b
+export const maybeToPromise = R.curry(
```
nkrkv: :+1:

PR #905, opened 2017-11-27
Add account pane

### LINKED ISSUE #889: Account pane
### Rationale

Prior to publishing a library (as described in #872) a xoder should be authorized. We should make a sidebar pane for this. The pane could also show resource quotas.

### Mockup

05-user

05-user2

Here are expected deviations from the not-logged-in mockup:

- Label “Not registered user” → “Guest xoder”
- Align “Compilations left” to the left
- Add right above the form “Login to increase quotas and access more features”
- Add “Forgot password” link below the password field
- Change the button labels: “Login” → “Log In”, “Register” → “Sign Up”

…and from the logged-in mockup:

- Add “Logged in as” line before the username
- Align “Compilations left” to the left
- Deaccent the button, make it ghost (i.e. outline only)
- Change the button label: “Logout” → “Log Out”

### Acceptance criteria

- [ ] I can toggle the pane by hitting “View → Toggle Account Pane”
- [ ] The login status is preserved across IDE launches (stored for a year)
- [ ] When network interaction for login is in progress, the button is disabled, a spinner is shown
- [ ] If the credentials are wrong, a common popup error is shown

evgenykochetkov:
Closes #889

### DIFF
--- a/packages/xod-client-browser/src/containers/App.jsx
+++ b/packages/xod-client-browser/src/containers/App.jsx
@@ -42,6 +42,7 @@ class App extends client.App {
     this.hideInstallAppPopup = this.hideInstallAppPopup.bind(this);
 
     props.actions.openProject(props.initialProject);
+    props.actions.fetchGrant();
   }
 
   onResize() {
@@ -206,6 +207,7 @@ class App extends client.App {
         [
           onClick(items.toggleHelpbar, this.props.actions.toggleHelpbar),
           onClick(items.toggleDebugger, this.props.actions.toggleDebugger),
+          onClick(items.toggleAccountPane, this.props.actions.toggleAccountPane),
         ]
       ),
       submenu(
--- a/packages/xod-client/src/core/components/Button.jsx
+++ b/packages/xod-client/src/core/components/Button.jsx
@@ -1,0 +1,37 @@
+import React from 'react';
+import cn from 'classnames';
+import PropTypes from 'prop-types';
+
+const Button = ({
+  as = 'button',
+  className,
+  disabled,
+  light,
+  small,
+  children,
+  ...restProps
+}) => React.createElement(
+  as,
+  {
+    ...restProps,
+    className: cn('Button', className, {
+      'Button--light': light,
+      'Button--small': small,
+      disabled,
+    }),
+    disabled,
+    role: as !== 'button' ? 'button' : restProps.role,
+  },
+  children
+);
+
+Button.propTypes = {
+  as: PropTypes.string,
+  className: PropTypes.string,
+  disabled: PropTypes.bool,
+  light: PropTypes.bool,
+  small: PropTypes.bool,
+  children: PropTypes.node,
+};
+
+export default Button;
--- a/packages/xod-client/src/user/actions.js
+++ b/packages/xod-client/src/user/actions.js
@@ -6,9 +6,9 @@ import {
   getLoginUrl,
   getLogoutUrl,
 } from '../utils/urls';
+import { addError } from '../messages/actions';
 import * as ActionTypes from './actionTypes';
 import * as Messages from './messages';
-
 
 export const updateCompileLimit = (startup = false) => dispatch =>
   fetch(getCompileLimitUrl(), {
@@ -63,12 +63,10 @@ export const login = (username, password) => (dispatch) => {
     .catch((err) => {
       const errMessage = err.status === 403
         ? Messages.INCORRECT_CREDENTIALS
-        : Messages.UNAVAILABLE;
+        : Messages.SERVICE_UNAVAILABLE;
+      dispatch(addError(errMessage));
 
-      return dispatch({
-        type: ActionTypes.LOGIN_FAILED,
-        payload: errMessage,
-      });
+      dispatch({ type: ActionTypes.LOGIN_FAILED });
     });
 };
 
--- a/packages/xod-client/src/user/actions.js
+++ b/packages/xod-client/src/user/actions.js
@@ -1,5 +1,14 @@
-import { getCompileLimitUrl } from '../utils/urls';
-import { UPDATE_COMPILE_LIMIT } from './actionTypes';
+import 'url-search-params-polyfill';
+
+import {
+  getCompileLimitUrl,
+  getWhoamiUrl,
+  getLoginUrl,
+  getLogoutUrl,
+} from '../utils/urls';
+import * as ActionTypes from './actionTypes';
+import * as Messages from './messages';
+
 
 export const updateCompileLimit = (startup = false) => dispatch =>
   fetch(getCompileLimitUrl(), {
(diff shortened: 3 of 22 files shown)


packages/xod-client/src/user/actions.js
```diff
@@ -1,13 +1,68 @@
...
+      }
+
+      return dispatch(getGrant());
+    });
+};
+
+export const logout = () => (dispatch) => {
+  fetch(getLogoutUrl(), { credentials: 'include' })
+    // to ignore keycloak's cors errors.
+    // TODO: add 'headless' logout route on server?
```
nkrkv: Yes, please :smiley:

packages/xod-client-browser/src/containers/App.jsx
```diff
@@ -206,6 +207,7 @@ class App extends client.App {
         [
           onClick(items.toggleHelpbar, this.props.actions.toggleHelpbar),
           onClick(items.toggleDebugger, this.props.actions.toggleDebugger),
+          onClick(items.toggleAccountPane, this.props.actions.toggleAccountPane),
```
nkrkv: Looks like it is missing in the Electron version.

nkrkv:
Let me suggest few improvements:

- Accent the “Log In” button as shown in the mockup
- Point the ”Sign Up” button to `https://xod.io/auth/#signup`. The site can’t interpret `#signup` right now, but we could teach it, that’s a feature required in many places

packages/xod-client/src/user/containers/AccountPane.jsx
```diff
@@ -0,0 +1,88 @@
...
+        role="button"
+        tabIndex="0"
+        className="close-button"
+        onClick={actions.toggleAccountPane}
+      >
+        &times;
+      </a>
+
+      {foldMaybe(
+        <div className="username">Guest XODer</div>,
```
nkrkv: If “Xoder” is a legal word it should be cased like a regular word: “Guest Xoder”

packages/xod-client/src/user/containers/AccountPane.jsx
```diff
@@ -0,0 +1,88 @@
...
+          <div>
+            <div>Logged in as</div>
+            <div className="username">{username}</div>
+          </div>
+        ),
+        user
+      )}
+
+      <div className="compilationsLeft">
+          Compilations left: {compilationsLeft == null ? 'currently offline' : compilationsLeft}
```
nkrkv: Recently I realized that “1000 compilations left” sounds like it is the limit ever. But it is for today. I think we should emphasize it. So I’d add a “Daily quotas” heading right above “Compilations left”

evgenykochetkov:
@nkrkv done!

packages/xod-client/src/core/components/Button.jsx
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+import cn from 'classnames';
+import PropTypes from 'prop-types';
+
+const Button = ({
```
brusherru: 🙌

packages/xod-client/src/user/actions.js
```diff
@@ -1,13 +1,65 @@
...
+export const toggleAccountPane = () => ({
+  type: ActionTypes.TOGGLE_ACCOUNT_PANE,
+});
+
+const setGrant = grant => ({
+  type: ActionTypes.SET_AUTH_GRANT,
+  payload: grant,
+});
+
+export const getGrant = () => dispatch =>
```
brusherru: @nkrkv used to tell me "this naming is not clear that a function with side-effects, asynchronous and etc. Let's name it with "load" or "fetch" verb. ;)

brusherru: Also, it will be good to add some comments, why we need this one in addition to `login` action.
(now I remember about the renewal of expired tokens and etc, but I can easily forget about it a couple of months later :)

packages/xod-client/src/user/containers/AccountPane.jsx
```diff
@@ -0,0 +1,91 @@
...
+  compilationsLeft,
+  actions,
+  isAuthorising,
+  authError,
+}) => {
+  if (!isVisible) return null;
+
+  return (
+    <div className="AccountPane">
+      <a
```
brusherru: Looks like a recurring pattern, it would be great to make it a separate component like you did with `Button` :)

packages/xod-client/src/user/actions.js
```diff
@@ -1,13 +1,65 @@
...
+
+  fetch(getLoginUrl(), { method: 'POST', credentials: 'include', body: form })
+    .then((res) => {
+      if (!res.ok) {
+        throw new Error(res);
+      }
+
+      return dispatch(getGrant());
+    })
+    .catch(() => dispatch({ type: ActionTypes.LOGIN_FAILED }));
```
brusherru: What if login service is unavailable, or network error occurred?
Maybe error should be passed into action payload?

P.S.
Also all these fetches, probably, should be wrapped with `retryOrFail` function from `xod-func-tools` to do retries on network/service unavailable errors...

packages/xod-client/src/user/reducer.js
```diff
@@ -1,10 +1,35 @@
...
+    case ActionTypes.UPDATE_COMPILE_LIMIT:
+      return R.assoc('limit', action.payload, state);
+    case ActionTypes.LOGIN_STARTED:
+      return R.merge(state, {
+        isAuthorising: true,
+      });
+    case ActionTypes.LOGIN_FAILED:
+      return R.merge(state, {
+        isAuthorising: false,
+        authError: 'Incorrect username or password',
```
brusherru: 🤔 
See comment for `login` action ;)

packages/xod-client/src/user/actions.js
nkrkv: @evgenykochetkov Agree with @brusherru for both points. Suggest `getGrant` → `fetchGrant`

packages/xod-client/src/user/actions.js
nkrkv: Oooh, very good point. Thought to test it in UI, but now it’s apparent network failures will not be handled gracefully.

evgenykochetkov:
@brusherru done, check it out please!

packages/xod-client/src/user/constants.js
```diff
@@ -0,0 +1,5 @@
+// eslint-disable-next-line import/prefer-default-export
+export const LOGIN_ERRORS = {
```
brusherru: Better to move it into `messages.js` (we're trying to move all messages for users in files `messages.js` to provide easy translation/grammar fixes by our contributors in the future).
Also, it could be separate constants and `eslint-disable` comment could be removed.

evgenykochetkov: Done!

nkrkv:
Kinda :beetle: 

### Steps to reproduce

1. Disconnect the machine from network
2. Click “Log In”

### Expected behavior

- A common mechanism error is shown like “Network connection is down”

### Actual behavior

- Looks like a no-op from user’s standpoint

The same happens with “Log out” press, of course.

nkrkv:
Also, I think it’s better to use the common error mechanism even for wrong username/password error. @brusherru @evgenykochetkov what do you think about it?

PR #907, opened 2017-11-28
Resolve library dependencies from client

### LINKED ISSUE #874: As a xoder I want all projects I open in IDE to resolve dependencies automatically

brusherru:
It will close #874.

It installs libraries with dependencies from client too.
Also, now it will not install (and download) libraries that already exists in a workspace.


packages/xod-client/src/project/messages.js
```diff
@@ -0,0 +1,3 @@
+export const PROJECT_HAS_INVALID_NODETYPES_IN_SOME_PATCHES = 'Project has invalid NodeTypes in some Patches';
```
nkrkv: I would be confused if see the message. Much better if it would be: “Patch @/foo can’t resolve its reference to bu/ba/ga”. Can we improve it such way?

packages/xod-pm/src/utils.js
```diff
@@ -51,6 +57,15 @@ export const parseLibQuery = R.compose(
   R.match(/^([^@/]+?)\/([^@/]+?)(?:@([^@/]+?))?$/)
 );
 
+// :: String -> LibName
+// It parses libQuery ("xod/core" or "xod/core@0.11.0")
+// and returns lib name without version ("xod/core").
+export const getPureLibName = libQuery => R.compose(
+  qp => `${qp.owner}/${qp.name}`,
+  explodeMaybe(`Expected correct library name format, like "xod/core@0.11.0", but get "${libQuery}"`),
```
nkrkv: “but get” → “got”.

Are you sure the pure is the best name here? I think about functional purity. Maybe something like `stripVersionFromLibName` is better?

PR #929, opened 2017-12-04
Handle network error in publishing library pipeline

### LINKED ISSUE #918: Wrong error message when disconnected from the network during publishing library
Moved from PR #911.

### Steps to reproduce
1. Log in.
2. Click "File -> Publish Library"
3. Click "Publish"
4. Disconnect the machine from network, while publishing in the progress

### Expected behavior
- A common mechanism error is shown like “Network connection is down”.
- Publish button appeared and I can click it again.
- In the rest, nothing changed.

### Actual behavior
- Error message "Cannot read `body` from undefined" appeared
- Publish button appeared and I can click it again.

brusherru:
Fixes #918


packages/xod-pm/src/messages.js
```diff
@@ -1,4 +1,5 @@
 export const CANT_PARSE_LIBRARY_REQUEST = 'Canʼt parse request to fetch library data';
+export const NO_NETWORK_CONNECTION = 'No network connection';
```
nkrkv: Maybe the traditional “Service is unavailable” is a more precise wording?

Perhaps the network connection is up, but our site is down.

brusherru: In that case, we'll get not an Error, we'll get an object with fields like `status`, `body` and etc. So we can detect network error and HTTP errors...

But it's really strange that this message appears only here, not everywhere where we can get the network error. So I'll replace it, but, I hope, in the future we could separate error "Service is unavailable" and "No Internet connection" :)

PR #939, opened 2017-12-06
Fix code generation when native patches implemented locally

nkrkv:
Local patches used to get C++ namespace names containing `@` symbol which is illegal
in C++ identifiers.

It made impossible to have a native patch in the project being transpiled.

### DIFF
--- a/packages/xod-arduino/src/transpiler.js
+++ b/packages/xod-arduino/src/transpiler.js
@@ -38,17 +38,23 @@ const kebabToSnake = R.replace(/-/g, '_');
 
 const createPatchNames = def(
   'createPatchNames :: PatchPath -> { owner :: String, libName :: String, patchName :: String }',
-  (path) => {
-    // TODO: this handles @/local-patches incorrectly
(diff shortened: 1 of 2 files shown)


packages/xod-arduino/src/transpiler.js
```diff
@@ -401,3 +407,7 @@ export const transformProjectWithDebug = def(
 );
 
 export const transpile = renderProject;
+
+export const forUnitTests = {
```
brusherru: A new pattern of exporting private functions for tests?
Will do the same thing everywhere in the future?

nkrkv: @brusherru @evgenykochetkov Just a proposal.

I find it useful because it’s easier to scan a module to understand what is its “real” API. Often the API includes only a tiny fraction of all symbols of the global scope. Having the explicit split one could be more focused when reviewing essential public API functions and relax a bit when looking at very local implementation details.

If you agree, let’s use it.

PR #940, opened 2017-12-06
Brand-new look of messages in the IDE

### LINKED ISSUE #925: Snackbar messages
### Rationale

Current messages are often too dark and hard to read or even notice. The idea is to apply an alternative system that can make important messages more visible and keep background messages nonobtrusive. To address readability issue we can split a message into title part (what happened), note part (how to fix), and buttons to read more or fix immediately.

### Mockup

An error takes all width of the *patch board*. The content takes 70% of the width. Of that width, the text takes 65% and 35% dedicated to buttons.

As shown on the mockup, persistent messages are rendered on a colorful background, and transient messages a printed on dark background with a colored title.

### How to implement

- Allow `addMessage` in `src/messages/actions.js` to trivially accept URLs
- Where appropriate, change formatting functions in `messages.js` to return `{ title, note, buttons :: [String] }` rather than a plain string and use that to build args for `addMessage`.

brusherru:
It closes #925 😎


nkrkv:
Much better.

On my machine, the X in close button has an ugly offset.

close-misalign

Can be fixed by adjusting CSS:

```css
.CloseButton:after {
  ...
  width: 18px; /* the same as for .CloseButton:before */
  right: 6px; /* the same as for .CloseButton:before */
}
```

Also, note text is hard to read. 

too-light-font

I suggest change `font-weight` from 200 to 300. And to compensate it a bit add `opacity` 0.9.

notes-adjusted-font

PR #942, opened 2017-12-09
UI performance optimisations and tools for benchmarking

evgenykochetkov:
This PR contains:
- Performance tweaks(~35% increase)
- Upgrade to React 16.2 for performance gains and just to get with the times
- Switch to redux dev tools **extension** for both browser and electron versions. No more docked panel that slowed down our dev builds.
- [why-did-you-update](https://github.com/maicki/why-did-you-update/) for catching unnecessary  rerenders — set `WHY_DID_YOU_UPDATE` env variable to true to enable it
- UI performance benchmark with puppeteer. Run it with `yarn benchmark:browser`, then load results from `packages/xod-client-browser/benchmark/tracing-results` in Chrome's performance tab


package.json
```diff
@@ -8,6 +8,7 @@
   "license": "AGPL-3.0",
   "scripts": {
     "bootstrap": "lerna bootstrap",
+    "benchmark:browser": "lerna run benchmark --scope xod-client-browser --stream",
```
brusherru: It will be nice to add benchmark script into CI pipeline and store benchmarks somewhere (GCS?) to detect speed changes. 
Or just fail the test if the benchmark runs more than allowed time. In this case, this time will have to increase in the same PR, and that will pay attention to an optimization problem or will have to optimize the just added code.
We should discuss that, I think.
Maybe it should be implemented later in other PR.

packages/xod-client-browser/benchmark/index.js
```diff
@@ -0,0 +1,116 @@
...
+    if (col === 0) {
+      await page.tracing.start({ path: getTracingResultsPath('adding_last_nodes') });
+    }
+
+    for (let row = ROWS; row >= 0; row -= 1) {
+      // press 'add node' button
+      await page.click('.PatchGroupItem.isSelected .add-node');
+
+      await page.waitForSelector('.Node.is-selected');
+      const nodeHandle = await page.$('.Node.is-selected');
```
brusherru: Isn't better to name it `nodeHandle` -> `nodeElement`?

brusherru: Or if it returns not a real element, but a special wrapper or something like this, it could be named as "selector" / "cursor" and etc. IMHO
Maybe I'm just used to all that "handle" is about the processing of any actions, calls and other things. In short, a function.

packages/xod-client-browser/benchmark/puppeteerUtils.js
```diff
@@ -0,0 +1,24 @@
+const getBoundingClientRect = (page, handle) => page.evaluate((el) => {
```
brusherru: The same thing as before (`handle` -> `element` | `selector`)

brusherru:
Also, it will be great to add information about performance optimization in the PR description.
Like "Node placing become faster for ~60% (0.3sec -> 0.18sec)".

package.json
evgenykochetkov: Yep, the original plan was to store profiles in GCS and implement it in another PR.

packages/xod-client-browser/benchmark/index.js
evgenykochetkov: It's a [term from official docs](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#class-elementhandle).

packages/xod-client-browser/benchmark/puppeteerUtils.js
evgenykochetkov: How about `elementHandle`? It does not sound like function and is consistent with official naming

brusherru:
> Performance tweaks(~35% increase)

Write a detailed summary, please.

PR #946, opened 2017-12-11
Fix watch nodes not updating in debug mode

evgenykochetkov:


packages/xod-client-electron/test-func/1-blink.spec.js
```diff
@@ -38,7 +38,6 @@ describe('IDE: Blink project', () => {
         .then(() => ide.page.assertNoPatchesAreOpen())
         // close patch group to restore initial state
         .then(() => ide.page.findPatchGroup(patchGroup).click())
```
brusherru: Maybe better to use `expandPatchGroup` in this case?

packages/xod-client/src/project/components/Node.jsx
```diff
@@ -23,21 +23,10 @@ class Node extends React.Component {
...
-        'position',
-        'dead',
-        'isSelected',
-        'isGhost',
-        'isDragged',
-        'hidden',
-        'noEvents',
-        'linkingPin',
-        'pinLinkabilityValidator',
+      R.omit([
```
brusherru: 👍

PR #947, opened 2017-12-11
Add `babel-plugin-transform-class-properties`, 

evgenykochetkov:
All those `this.foo = this.foo.bind(this)` in constructors are quite annoying.
[transform-class-properties](https://babeljs.io/docs/plugins/transform-class-properties/) plugin is widely adopted in the community(it's even included in Create React App). Corresponding proposal is at stage 3. And there is a nice [codemod](https://github.com/reactjs/react-codemod#manual-bind-to-arrow) that completely automates the transition.

The first commit contains all required preparations, the second one is a result of running a codemod.


nkrkv:
After the short discussion, we concluded the change is questionable and needs more investigation.

@evgenykochetkov would you close the PR for now?

PR #956, opened 2017-12-14
Add articles on C++ node creation

nkrkv:
Five new articles are here :car: :tractor: :blue_car: :truck: :fire_engine:


docs/guide/cpp-state/README.md
```diff
@@ -0,0 +1,163 @@
...
+we need to store a single counter value, so our struct will have a single field.
+Let’s call it `counterValue`:
+
+```cpp
+struct State {
+    Number counterValue = 0;
+};
+```
+
+All state value regardless of type start with their default values. The default
```
brusherru: I'm not sure, but maybe "All state value**s**". And maybe "regardless of type" should be wrapped with commas. :)

docs/guide/cpp-state/README.md
```diff
@@ -0,0 +1,163 @@
...
+value of numbers is `0`, so this assignment is not required. We add it here just
+to demonstrate a possibility to initialize with another initial value.
```
brusherru: It's not clear, If I need default `0` value, should I specify `Number counterValue` or not.
On the other hand, the article is for advanced guys who may do something with C++, and perhaps this remark is not actual.

docs/guide/cpp-state/README.md
```diff
@@ -0,0 +1,163 @@
...
+struct State {
+    Number counterValue = 0;
+};
+
+\{{ GENERATED_CODE }}
+
+void evaluate(Context ctx) {
+    State* state = getState(ctx);
+
+    if (isInputDirty<input_INC>(ctx)) {
```
brusherru: Wow-wow. Slow down.
Is user already knows function `isInputDirty`?

brusherru: In the next article, you put a link to C++ API. So maybe it will be good to place it here too (or just say "Don't be afraid, we'll tell about this later"). :)

docs/guide/nodes-for-xod-in-cpp/README.md
```diff
@@ -0,0 +1,135 @@
...
+    // Note we use `OUT` label to access our unlabeled output terminal.
+    // `OUT` is the default name when you omit the label.
+    emitValue<output_OUT>(ctx, result);
+}
+```
+
+You may wonder why we used `output_OUT` to refer our unlabeled output terminal.
+That’s a default name when you omit the label. Likewise, if you omit a label for
+an input, use `input_IN` to refer to it. And if you have got several unnamed
+terminals use `input_IN1`, `input_IN2`, `output_OUT1`, `output_OUT2`, etc.
```
brusherru: Maybe will be useful to add a hint about hovering pin in the Inspector to get these names or just count it from left to right and then from the top down.

docs/guide/cpp-state/README.md
nkrkv: Not sure what do you mean. Isn’t “*so this assignment is not required*” says it is optional? Or do you talk about another thing?

docs/guide/cpp-state/README.md
nkrkv: Done

docs/guide/cpp-state/README.md
nkrkv: Wrote down an introduction to `isInputDirty`

BTW, the order in Guide section is very synthetic. It’s better to look at it as a pile of blablapedia articles

docs/guide/nodes-for-xod-in-cpp/README.md
nkrkv: Done

docs/guide/cpp-state/README.md
brusherru: I mean could User just leave State empty, or he still need to write `Number counterValue` but without value ;)

nkrkv: Ahem… he uses `counterValue`, so he definitely required to declare it. OK, I’ve got the problem, will clarify it a bit.

Issue #932, opened 2017-12-05
Tooltipish help for quick search


nkrkv:
### Mockup

![03-help-search](https://user-images.githubusercontent.com/146383/33601924-d12e4d72-d9be-11e7-89b2-c76a9f86f953.png)

The tooltip logic is the same as in #930 

### Icons

- [search.zip](https://github.com/xodio/xod/files/1530736/search.zip)
- [cross.zip](https://github.com/xodio/xod/files/1530738/cross.zip)

### Acceptance criteria

- [x] The quick search field and suggestions are shown at the work area center minus 100% of their width
- [x] I see the help tooltip for any selected suggested item
- [x] The suggestion is confirmed by a single-click (behavior change)
- [x] An item switches on mouse move *or* up/down arrow press (behavior change)
- [x] I see text input icons as shown in the mockup

PR #966, opened 2017-12-20
Brand-new tooltipish Helpbox for ProjectBrowser

### LINKED ISSUE #930: Tooltipish node help in project browser
### Acceptance criteria

- [x] The help is invoked via node context menu
- [x] The tooltip switches to the new node when it is selected
- [x] The tooltip closes on (x) click or ESC
- [x] Width grows with node width grow
- [x] Nodes with 8+ pins in width are scaled down on preview
- [x] The tooltip never goes out of the screen. If too low, the tip moves down, the content rectangle stays visible

brusherru:
It closes #930.


packages/xod-client/src/core/styles/abstracts/variables.scss
```diff
@@ -6,8 +6,8 @@ $font-family-display: 'Roboto Slab', sans-serif;
 $font-size-xs: 8px;
 $font-size-s: 9px;
 $font-size-m: 11px;
-$font-size-l: 18px;
-$font-size-xl: 28px;
+$font-size-l: 14px;
+$font-size-xl: 18px;
```
evgenykochetkov: 👍🏻

nkrkv:
:heart: Few issues:

- [ ] The scrollbar in a help tooltip is native and ugly. Can we change it to the XOD one?
- [ ] When a project browser item goes out of the visible area I expect the tooltip to close automatically (equivalent to (x) click). It does not happen.

help-bug

packages/xod-client/src/core/selectors.js
```diff
@@ -29,56 +29,25 @@ export const hasUnsavedChanges = createSelector(
...
-      R.when(
-        () => focusedArea === FOCUS_AREAS.PROJECT_BROWSER,
-        R.reverse
-      )
-    )([
-      maybeSelectedNodeTypeInWorkarea,
-      maybeSelectedPathInProjectBrowser,
-    ]);
-  }
+export const getPatchforHelpbox = createSelector(
```
nkrkv: Capitalize “For”?

packages/xod-client/src/core/selectors.js
```diff
@@ -29,56 +29,25 @@ export const hasUnsavedChanges = createSelector(
...
+  [Project.getProject, ProjectBrowser.getSelectedPatchPath],
+  (project, selectedPatchPath) => R.compose(
+    R.chain(XP.getPatchByPath(R.__, project)),
+    Maybe
+  )(selectedPatchPath)
+);
+export const getPatchforQuickHelp = createSelector(
```
nkrkv: Capitalize “For”?

packages/xod-client/src/projectBrowser/components/PatchGroupItemContextMenu.jsx
```diff
@@ -5,7 +5,14 @@ import { ContextMenu, MenuItem, connectMenu } from 'react-contextmenu';
 
 import { PATCH_GROUP_CONTEXT_MENU_ID } from '../constants';
 
-const callCallbackWithPatchPath = onClick => (event, data) => onClick(data.patchPath);
+const callCallbackWithPatchPath = onClick => (event, data) => {
```
evgenykochetkov: I think we should rename this function 🤔

packages/xod-client/src/core/styles/components/Helpbox.scss
```diff
@@ -0,0 +1,59 @@
...
+  transition: opacity 0.1s;
+
+  &--hidden {
+    opacity: 0;
+    pointer-events: none;
+  }
+
+  &-content {
+    padding: 12px;
+    overflow: auto; // TODO
```
nkrkv: TODO what? Does this thing cause native scrollbars?

packages/xod-client/src/editor/constants.js
```diff
@@ -95,3 +95,7 @@ export const SIDEBAR_IDS = {
 };
 
 export const PANEL_CONTEXT_MENU_ID = 'PANEL_CONTEXT_MENU_ID';
+
+// Event name for pub/sub to update position of Helpbox
+// while User scrolls ProjectBrowser or selects another Patch
+export const UPDATE_HELPBOX_POSITION = 'UPDATE_HELPBOX_POSITION';
```
nkrkv: Should we better call it `SCROLL_PROJECT_BROWSER`? That’s the fact, and updating the helpbox position is one of the reactions (although the only one at the moment)

packages/xod-client/src/core/styles/components/Helpbox.scss
brusherru: Yep, forget it 😬

packages/xod-client/src/editor/constants.js
brusherru: In fact, it listens to scroll and click events, so it's not just "SCROLL"...
But you're right, that's only about updating position from ProjectBrowser.

I propose to leave it as is and rename it to a more appropriate name later (during #932) when it will receive update events from the quick search.

brusherru:
All fixed, except renaming of `UPDATE_HELPBOX_POSITION`.

Also did small fixes of sidebar system (when any Panel was opened on a whole sidebar it's not fits window height and the latest item was out of the viewport, so I use flexbox and problem disappeared) and rename `Dock To Right` into `Dock to Right` and `to Dock Left`. :)

nkrkv:
I have circle background behind (x) disappeared. Can you bring it back?

Also, I’m not sure we should expose the disappeared helpbox again when the scrolled item is visible again. It is questionable. Seems like nothing bad about the current behavior, but imagine a scenario when a user invokes help, scroll the item out so that it disappears and does other things for five minutes. Then it scrolls the project browser and boom :boom: helpbox appears. He already forgot about it. Wouldn’t it look like a bug?

brusherru:
> Then it scrolls the project browser and boom 💥 helpbox appears. He already forgot about it.

Sounds reasonable. Fixed it.
And fixed background behind cross in close button.

PR #971, opened 2017-12-22
Fix infinite retries while fetching lib datas / libs

### LINKED ISSUE #970: Entering invalid lib name in library suggester results in repeated network requests
### Steps to reproduce

1. Open IDE
2. Select "file -> add library" from the main menu
3. Type something like `xod/common-`
4. Observe console and network tab

### Expected behavior

There is a single check for a library with this name

### Actual behavior

IDE repeatedly tries to get info about `xod/common-`

### How to fix

https://github.com/xodio/xod/blob/ab7bac3985b52f1f09c593899d0e5f3ab69f0b46/packages/xod-pm/src/fetch.js#L36-L63

- check for more cases, not only 404(or validate lib name before checking)
- do not use "original" `fetchLibData` function as a retry function, because it will never stop

brusherru:
It fixes #970.

### DIFF
--- a/packages/xod-pm/src/fetch.js
+++ b/packages/xod-pm/src/fetch.js
@@ -19,9 +19,9 @@ import { parseLibQuery, unfoldMaybeLibQuery, rejectUnexistingVersion, getLibName
 // =============================================================================
 
 // :: (b -> c) -> (() => Promise a b) -> Promise a c
-const retryExcept404 = retryOrFail(
+const retryExceptAny400 = retryOrFail(
   [500, 1000, 2000, 3000],
-  res => (res.status && res.status === 404),
+  res => (res.status && res.status >= 400 && res.status < 500),
 );
 
 // =============================================================================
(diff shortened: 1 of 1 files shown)


packages/xod-pm/src/fetch.js
```diff
@@ -33,67 +33,80 @@ const retryExcept404 = retryOrFail(
...
+    return tryFn()
+      .catch(
+        retryExceptAny400(
+          rejectFetchResult({
+            message: MSG.cantFindLibVersion(params),
+            errorCode: ERR_CODES.CANT_GET_LIB_XODBALL,
+            request: libQuery,
+            params,
+          }),
+          () => fetchLibrary(swaggerUrl, libQuery)
```
evgenykochetkov: Shouldn't `tryFn` be here instead?

packages/xod-pm/src/fetch.js
```diff
@@ -19,9 +19,9 @@ import { parseLibQuery, unfoldMaybeLibQuery, rejectUnexistingVersion, getLibName
 // =============================================================================
 
 // :: (b -> c) -> (() => Promise a b) -> Promise a c
-const retryExcept404 = retryOrFail(
+const retryExceptAny400 = retryOrFail(
   [500, 1000, 2000, 3000],
-  res => (res.status && res.status === 404),
+  res => (res.status && String(res.status)[0] === '4'),
```
nkrkv: :open_mouth: Number arithmetics with strings? `(res.status >= 400 && res.status < 500)` ?

brusherru: I'm so stupid 🙈

PR #972, opened 2017-12-22
Tweak Helpbox positioning (right-sided)

brusherru:
There is no issue.
But now Helpbox will render to the left of a referred element if it could not fit the window width.
E.G. Helpbox renders fine for right-sided ProjectBrowser.

I'm not sure that `rightSided` is a good name for this, maybe it should be `leftSided` or something else. Open to suggestions.


nkrkv:
I guess “right-hand side” and “left-hand side” are proper terms. Or simply LHS and RHS.

PR #977, opened 2017-12-25
Add code highlighting in Cpp Editor

brusherru:
There is no issue.

It will highlights:
* Types from XOD (`Number`, `State`, `XString` and etc),
* Built-in functions (`getValue`, `setTimeout` and etc),
* Special keywords (`GENERATED_CODE`, `input_$$$` and etc)
* Some built-in functions from Arduino (`digitalRead`, `pinMode` and etc), that are approved by XOD Team (for example, synchronous functions which will lead to a holdup of running transactions, like `delay()`, is not included)

### DIFF
--- a/packages/xod-client/src/core/styles/components/CodeMirror.scss
+++ b/packages/xod-client/src/core/styles/components/CodeMirror.scss
@@ -433,21 +433,20 @@ span.CodeMirror-selectedtext { background: none; }
 .cm-s-xod .cm-operator { color: $chalk-bright; }
 .cm-s-xod .cm-variable-2 { color: $apricot-light; }
 .cm-s-xod .cm-variable-3, .cm-s-xod .cm-type { color: $apricot-light; }
-.cm-s-xod .cm-builtin { color: $apricot-light; }
+.cm-s-xod .cm-builtin { color: $blueberry; }
 .cm-s-xod .cm-atom { color: $red-bright; }
 .cm-s-xod .cm-number { color: $red-bright; }
-.cm-s-xod .cm-def { color: #77A0FF; }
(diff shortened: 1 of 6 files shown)


packages/xod-client/src/editor/components/CppImplementationEditor.jsx
```diff
@@ -15,11 +19,18 @@ const CppImplementationEditor = ({
   const options = {
     lineNumbers: true,
     readOnly: isInDebuggerTab,
-    mode: 'text/x-c++src',
+    mode: 'xod', // 'text/x-c++src',
```
nkrkv: I guess we should call it `text/x-c++xod` to keep consistent. Eventually, we could invent yet more XOD-specific modes, like XOD Markdown doc.

packages/xod-client/src/core/styles/components/CodeMirror.scss
```diff
@@ -433,21 +433,20 @@ span.CodeMirror-selectedtext { background: none; }
 .cm-s-xod .cm-operator { color: $chalk-bright; }
 .cm-s-xod .cm-variable-2 { color: $apricot-light; }
 .cm-s-xod .cm-variable-3, .cm-s-xod .cm-type { color: $apricot-light; }
-.cm-s-xod .cm-builtin { color: $apricot-light; }
+.cm-s-xod .cm-builtin { color: #77A0FF; }
```
nkrkv: This color value is duplicated in few places. Perhaps it is reasonable to extract as a named constant (blueberry?)

brusherru:
All fixed. Renamed mode into `xodCpp`, but defined MIME type that associated with this mode: `text/x-c++xod`.

Also, now `Tab` indent code with spaces.

PR #981, opened 2017-12-26
Fix console-log implementation

evgenykochetkov:


workspace/__lib__/xod/core/console-log/patch.cpp
```diff
@@ -15,11 +15,10 @@ void evaluate(Context ctx) {
...
-        for (auto it = line->iterate(); it; ++it)
-            Serial.write((char)*it);
-        Serial.write('\r');
-        Serial.write('\n');
-        Serial.flush();
-    }
+
+    for (auto it = line->iterate(); it; ++it)
+        Serial.write((char)*it);
+    Serial.write('\r');
```
brusherru: 🤔  Looks like it will write empty lines even if a line is empty on each pulse...

nkrkv: Yes. What do you see in JS or any other REPL when you do an empty “print”? An empty line. That’s fine and expected.

brusherru: 👌

PR #1016, opened 2018-01-12
Fix copy text from Quick Help

### LINKED ISSUE #1000: Can’t copy text from Quick Help
A bug that hardens new nodes crafting.

### Steps to reproduce

1. Show Quick Help panel
2. Select a node with description
3. Try to copy its description into clipboard (select text + Ctrl+C, or select text then right click)

### Expected behavior

The text behaves like regular text on a web page, i.e. it is copied on Ctrl+C, context menu is available

### Actual behavior

Nothing

brusherru:
It closes #1000.
Also, I tweak styles of cursors inside QuickHelp and Helpbox to change it more accurately to `text` (not everywhere, only on text),

### DIFF
--- a/packages/xod-client/src/editor/actions.js
+++ b/packages/xod-client/src/editor/actions.js
@@ -312,16 +312,19 @@ const getClipboardEntities = (state) => {
 };
 
 export const copyEntities = event => (dispatch, getState) => {
-  if (isInput(document.activeElement)) return;
+  // If user clicked somewhere on Patch (select Node or something else
+  // except editing comment node) activeElement will be <div />
+  // with className `PatchWrapper`.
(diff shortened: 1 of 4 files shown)


packages/xod-client/src/editor/actions.js
```diff
@@ -312,7 +312,7 @@ const getClipboardEntities = (state) => {
 };
 
 export const copyEntities = event => (dispatch, getState) => {
-  if (isInput(document.activeElement)) return;
+  if (isInput(document.activeElement) || document.activeElement.className !== 'PatchWrapper') return;
```
evgenykochetkov: Please add a comment why this new check is required

PR #1019, opened 2018-01-12
Preserve patch offset when starting debugger

### LINKED ISSUE #1012: When debugger starts patch offset should be preserved
### Steps to reproduce

1. Create a patch with a sensor and `watch`
2. Select all, move down, pan, move down… 1.5-2 screens at total
3. Upload with debugger
4. Wait for the debugger to start

### Expected behavior

- The nodes are visible
- They are only offset down a bit to contain the breadcrumbs line (see #1011)

### Actual behavior

- I see an empty canvas, I’m lost
- If I stop debug, I still see the empty canvas
- That’s hard to understand intuitively, but the nodes are 1.5-2 screens to the bottom

evgenykochetkov:
Closes #1012

### DIFF
--- a/packages/xod-client/src/editor/reducer.js
+++ b/packages/xod-client/src/editor/reducer.js
@@ -599,9 +599,9 @@ const editorReducer = (state = {}, action) => {
       const currentPatchPath = currentTab.patchPath;
       const currentOffset = currentTab.offset;
       return R.compose(
+        setTabOffset(currentOffset, DEBUGGER_TAB_ID),
         drillDown(action.payload.patchPath, null),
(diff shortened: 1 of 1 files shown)


packages/xod-client/src/editor/reducer.js
```diff
@@ -599,9 +599,9 @@ const editorReducer = (state = {}, action) => {
       const currentPatchPath = currentTab.patchPath;
       const currentOffset = currentTab.offset;
       return R.compose(
+        setTabOffset(currentOffset, DEBUGGER_TAB_ID),
```
nkrkv: Could you explain the magic? I can’t understand why order matters here.

PR #1022, opened 2018-01-16
Fix wrong wrapping of errors with dead reference error message

### LINKED ISSUE #999: Misguiding “Dead references” errors
Try the scenario from #998 

It shows “Patch @/main contains dead references: `<string with the actual problem>`. Fix or delete them to continue”

The text wrapper about dead references applied improperly. We should only use it when the problem is really in dead references.

brusherru:
It closes #999.

### DIFF
--- a/packages/xod-client/src/core/containers/App.jsx
+++ b/packages/xod-client/src/core/containers/App.jsx
@@ -10,6 +10,7 @@ import {
   toXodball,
   isValidIdentifier,
   IDENTIFIER_RULES,
+  isDeadRefError,
 } from 'xod-project';
 import {
   transformProject,
@@ -26,7 +27,7 @@ import PopupProjectPreferences from '../../project/components/PopupProjectPrefer
 import PopupPublishProject from '../../project/components/PopupPublishProject';
 
 import * as actions from '../actions';
-import { NO_PATCH_TO_TRANSPILE } from '../../editor/messages';
+import { NO_PATCH_TO_TRANSPILE, deadRefErrorMessage } from '../../editor/messages';
 import composeMessage from '../../messages/composeMessage';
 
 export default class App extends React.Component {
@@ -45,13 +46,27 @@ export default class App extends React.Component {
   onShowCodeArduino() {
     return R.compose(
       foldEither(
-        R.compose(
-          this.props.actions.addError,
-          R.when(
+        R.cond([
+          [
+            isDeadRefError,
+            err => this.props.actions.addError(
+              deadRefErrorMessage(err),
+              true
+            ),
+          ],
+          [
             R.is(Error),
-            err => composeMessage(err.message),
-          )
-        ),
+            err => this.props.actions.addError(
+              composeMessage(err.message)
+            ),
+          ],
+          [
+            R.T,
+            (x) => {
+              throw Error(`Expected Either.Left Error, but got just ${typeof x} "${x}"`);
+            },
+          ],
+        ]),
         this.props.actions.showCode
       ),
       R.map(transpile),
--- a/packages/xod-client/src/editor/messages.js
+++ b/packages/xod-client/src/editor/messages.js
@@ -1,3 +1,4 @@
+import { getDeadRefErrorReason } from 'xod-project';
 import composeMessage from '../messages/composeMessage';
 
 export const PATCH_FOR_NODE_IS_MISSING = 'Patch for this node is missing.';
@@ -20,4 +21,12 @@ export const LIB_SUGGESTER_NOTHING_FOUND = 'No library found';
 export const NO_PATCH_TO_TRANSPILE = composeMessage(
   'No patch opened',
   'Open a patch to upload and try again'
+);
+
+export const deadRefErrorMessage = err => composeMessage(
+  err.title,
+  [
+    getDeadRefErrorReason(err),
+    'Fix or delete them to continue.',
+  ].join('\n')
 );
--- a/packages/xod-project/src/flatten.js
+++ b/packages/xod-project/src/flatten.js
@@ -154,12 +154,17 @@ export const extractLeafPatches = def(
               CONST.ERROR.IMPLEMENTATION_NOT_FOUND,
               { patchPath: Patch.getPatchPath(patch) }
             )
-          ),
+          )
         ),
       ],
       [
         R.T,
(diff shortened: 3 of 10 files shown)


brusherru:
Added one commit. Check it out and I'll squash it if it's okay.
In the future, when we'll add xod stack traces to all Errors (at least that occurred in `flatten`), it could be a little refactored and current `DeadRefError` will become `XodError` or something like this.

packages/xod-client/src/core/containers/App.jsx
```diff
@@ -45,13 +46,27 @@ export default class App extends React.Component {
...
-          )
-        ),
+            err => this.props.actions.addError(
+              composeMessage(err.message)
+            ),
+          ],
+          [
+            R.T,
+            (x) => {
+              throw Error(`Expected Either.Left Error, but got just ${typeof x} "${x}"`);
```
evgenykochetkov: 👍🏻
awesome

nkrkv:
Can’t check due to a bug

### Steps to reproduce

1. Apply the following diff to the `blink` project:

```diff
diff --git a/workspace/blink/main/patch.xodp b/workspace/blink/main/patch.xodp
index ec81813e..9a5fb0db 100644
--- a/workspace/blink/main/patch.xodp
+++ b/workspace/blink/main/patch.xodp
@@ -21,6 +21,17 @@
         "nodeId": "rkwIEFrWb",
         "pinKey": "HkyxURuSPyW"
       }
+    },
+    {
+      "id": "ry2wNFB-Z",
+      "input": {
+        "nodeId": "SyeDNFBWZ",
+        "pinKey": "HJ6HsEngZ"
+      },
+      "output": {
+        "nodeId": "rkwIEFrWb",
+        "pinKey": "HkyxURuSPyWz"
+      }
     }
   ],
   "nodes": [
```

2. Show code for Arduino

### Expected behavior

An error about dead links

### Actual behavior

Nothing

### Console output

```
Uncaught Error: Maybe is expected to be Just, but its Nothing.
    at explode (monads.js:56)
    at cond.js:42
    at _arity.js:5
    at _pipe.js:3
    at _pipe.js:3
    at _arity.js:5
    at flatten.js:190
    at _curryN.js:37
    at _arity.js:5
    at converge.js:41
```

I suggest switch from `explode` to `explodeMaybe` to also show meaningful assertion message rather than generic “Maybe is expected to be Just”

packages/xod-client/src/editor/messages.js
```diff
@@ -21,3 +22,11 @@ export const NO_PATCH_TO_TRANSPILE = composeMessage(
   'No patch opened',
   'Open a patch to upload and try again'
 );
+
+export const deadRefErrorMessage = err => composeMessage(
+  err.title,
+  [
+    getDeadRefErrorReason(err),
+    'Fix or delete them to continue.',
```
nkrkv: Suggest changing

```
@/main → @/foo → @bar: Can’t find the patch in the project with specified path: “@/bar”
Fix or delete the dead references to continue
```

To

```
Patch @/bar is missing. Required by @/main → @/foo.
Delete @/bar nodes or bring back the patch to continue.
```

packages/xod-client/src/editor/messages.js
```diff
@@ -21,3 +22,11 @@ export const NO_PATCH_TO_TRANSPILE = composeMessage(
   'No patch opened',
   'Open a patch to upload and try again'
 );
+
+export const deadRefErrorMessage = err => composeMessage(
+  err.title,
+  [
+    getDeadRefErrorReason(err),
```
nkrkv: Too much logic for `messages.js`. I’d move it next to `composeMessage` leaving here only very simple text formatting:

```js
deadReference = (deadPatch, pathToDeadPatch) => composeMessage(
  'Patch contains dead references',
  [
    `Patch ${deadPatch} is missing. Required by ${pathToDeadPatch}`,
    `Delete ${deadPatch} nodes  or bring back the patch to continue.`,
  ].join('\n')
);
```

packages/xod-client/src/core/containers/App.jsx
```diff
@@ -45,13 +46,27 @@ export default class App extends React.Component {
...
       foldEither(
-        R.compose(
-          this.props.actions.addError,
-          R.when(
+        R.cond([
+          [
+            isDeadRefError,
+            err => this.props.actions.addError(
+              deadRefErrorMessage(err),
+              true
```
nkrkv: Oh, branching. Again, dead reference is something special. Nope, `App.jsx` should not care.

We need polymorphism at some point anyway, but it would be logical if it happens at the latest point possible. I think type-based decisions should be encapsulated in a module with very simple API. Say, `formatError :: Error -> TheTypeComposeMessageReturns`

In other words, this hunk should look like:

```js
foldEither(
  err => this.props.actions.addError(formatError(err), true),
  this.props.actions.showCode
)
```

packages/xod-project/src/flatten.js
```diff
@@ -257,39 +284,43 @@ const createCastNode = R.curry((patchTuples, nodes, link) => R.compose(
  * @param {Array<Array<Path, Patch>>} patchTuples
  * @param {Array<Node>} nodes
  * @param {Link} link
- * @returns {Array<Node|Null, Array<Link>>}
+ * @returns {Either<Error|Array<Node|Null, Array<Link>>>}
```
nkrkv: `Either<Error| ...` →  `Either<Error, ...`

packages/xod-project/src/flatten.js
```diff
@@ -854,26 +888,30 @@ const convertPatch = def(
...
-        R.append([path, convertedPatch]),
-        removeNotImplementedInXodNodes,
-        removeTerminalPatches
-      )),
-      R.sequence(Either.of),
-      addCastPatches(project, usedCastNodeTypes)
-    )(leafPatches);
-  }
+  (project, path, patch, leafPatches) => R.compose(
+    composeDeadRefError(path),
```
nkrkv: The `composeDeadRefError` is misguiding. Sounds like we always create an error here rather than qualifying the existing.

```js
foldEither(
  err =>  Either.Left(assocPatchPathToError(path, err)),
  R.identity
)
```

brusherru:
The final solution is yet to be discussed, cause we have to do a refactoring to make all errors shows properly with "xod stack trace" and etc.

PR #1023, opened 2018-01-16
Fix code editor: cursors, highlighting and tweak it a little

### LINKED ISSUE #1005: In code editor cursor should be I-beam
### Steps to reproduce

- Open a C++ node implementation
- Hover the editor

### Expected behavior

The cursor is I-beam

### Actual behavior

The cursor is arrow

### LINKED ISSUE #1006: Code editor does not highlight pin labels with lower-case letters
### Steps to reproduce

In a C++ implementation enter:

```cpp
emitValue<output_Tc>(ctx, 42);
```

### Expected behavior

`output_Tc` is highlighted

### Actual behavior

`output_Tc` has default grey color

### LINKED ISSUE #1007: Code editor highlights commented out XOD keywords
### Steps to reproduce

In C++ implementation add:

```cpp
// emitValue<output_OUT>(ctx, 42);
/*
  emitValue<output_DONE>(ctx, 1);
*/
```

### Expected behavior

The code is rendered in gray as a regular comment

### Actual behavior

The `emitValue` and `output_*` are colored.

### How to fix

Give comment rules a higher priority over the XOD keywords.

### LINKED ISSUE #1013: Style code editor scrollbars
The embedded code editor scrollbars are native and ugly (because it is dark) on Win and Linux. We can either:

1. Use custom scroll as it is done in most places
2. Minimize the effect by styling them the same way it is done for the deployment panel

The second option is OK if the first is too cumbersome to implement.

brusherru:
It fixes #1005 :heavy_check_mark: , fixes #1007 :heavy_check_mark: , closes #1006 :heavy_check_mark:  and closes #1013 :heavy_check_mark: .

Also, I have added some addons:
* easy toggle comments by `Ctrl+/` or `Cmd+/` (but only line comments, block comments doesn't want to work 😞 )
* auto close opening brackets
* highlight trailing spaces with red dots
* auto clear trailing spaces on empty lines


packages/xod-client/src/editor/codemirrorXodMode.js
```diff
@@ -3,19 +3,65 @@ import 'codemirror/addon/mode/simple';
...
 (() => {
   let xodOverlay;
   CodeMirror.defineMode('xodCpp', (config, parserConfig) => {
     xodOverlay = CodeMirror.simpleMode(config, {
       mode: { spec: CodeMirror.modes.clike },
-      start: getTokens(),
+      start: getTokens().concat(
+        [
+          { regex: /\/\/.*/, token: 'comment' },
+          { regex: /\/\*/, token: 'comment', next: 'comment' },
```
nkrkv: Quite hard to understand why `getTokens` are external and the comments are concated here. I’m pretty sure the `getTokens` should not belong to `xod-arduino` at all (see its `README`, the better name for the package would be `xod-arduino-transpiler`).

Wouldn’t it more clear to get rid of `getTokens` in a separate module and define them right in this module as they are much more codemirror-related than Arduino-related? There automatically will be less hassle with the config split.

packages/xod-arduino/src/getTokens.js
```diff
@@ -2,7 +2,7 @@
 
 const getTokens = () => [
   {
-    regex: /\b((?:input|output)_[A-Z0-9-_]+)\b/gm,
+    regex: /\b((?:input|output)_[A-Za-z0-9-_]+)\b/gm,
```
nkrkv: Dash is extraneous. C++ identifiers are not allowed to contain dashes.

packages/xod-client/src/editor/components/CppImplementationEditor.jsx
```diff
@@ -40,7 +40,13 @@ const CppImplementationEditor = ({
         }
       },
       'Shift-Tab': CM => CM.indentSelection('subtract'),
+      'Ctrl-/': 'toggleComment',
+      'Cmd-/': 'toggleComment',
     },
+    showTrailingSpace: true,
+    autoCloseBrackets: true,
+    autoClearEmptyLines: true,
```
nkrkv: I worry about this option. I find it harmful. Imagine a xoder just browses a project without intention to change anything. A simple cursor movement would easily *change* the project. Yes, it will be a simple whitespace removal, but it leads to save/discard question, false diffs for Git, etc…

I think it would be enough to just annoy a coder with the red dots and do not clear empty lines implicitly.

brusherru: Fixed, now it will clear empty lines only on "enter" keypress.
And clear only lines above the cursor, while it does not meet not empty line.

brusherru:
@nkrkv check out latest commit, please! :)

packages/xod-client/src/editor/containers/Editor.jsx
```diff
@@ -151,6 +151,10 @@ class Editor extends React.Component {
           const source = XP.getImpl(patch).getOrElse(IMPL_TEMPLATE);
 
           const onChange = src => this.updatePatchImplementationDebounced(patchPath, src);
+          const currentPatchPath = explodeMaybe(
+            'No currentPatchPath, but currentTab exists',
+            this.props.currentPatchPath
+          );
```
evgenykochetkov: 👍🏻

PR #1024, opened 2018-01-16
Fix panning mode: show grab cursors and disable hovering of Patch contents

### LINKED ISSUE #1009: While panning the cursor should change
### Steps to reproduce

1. Pan with middle-button
2. Pan with space + drag

### Expected behavior

- The open hand cursor is shown when spacebar is held down
- The grab-hand cursor is shown when spacebar+left button held down or middle button is held down

### Actual behavior

Arrow cursor everywhere

brusherru:
It closes #1009

### DIFF
--- a/packages/xod-client/src/editor/containers/Patch/modes/panning.jsx
+++ b/packages/xod-client/src/editor/containers/Patch/modes/panning.jsx
@@ -133,6 +133,8 @@ const panningMode = {
               linkingPin={api.props.linkingPin}
             />
           </g>
+          {/* It's an overlay <rect> to disable hovering effects on patch contents */}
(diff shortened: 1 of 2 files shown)


packages/xod-client/src/editor/containers/Patch/modes/panning.jsx
```diff
@@ -133,6 +133,7 @@ const panningMode = {
               linkingPin={api.props.linkingPin}
             />
           </g>
+          <rect x="0" y="0" width="100%" height="100%" fill="transparent" stroke="none" />
```
evgenykochetkov: I think a comment explaining the purpose of this `rect` would be nice

PR #1025, opened 2018-01-16
Prevent overlaping of Patch contents by Breadcrumbs

### LINKED ISSUE #1011: Breadcrumbs overlap patch canvas in the most common scenario
### Steps to reproduce

1. Create a patch with a bounding box created by nodes (not comments)
2. Close the patch and open it again to pan it to the origin
3. Upload with debugger

### Expected behavior

The contents of the patch board are offset down so that space is “eaten” from the bottom.

### Actual behavior

The breadcrumbs overlap the patch.

regular-patch

debugging-patch

### How to fix

Move breadcrumbs out of the patch board in the layout, make it block-element on its own like it behaves in the code editor.

brusherru:
It closes #1011.

### DIFF
--- a/packages/xod-client/src/core/styles/components/Debugger.scss
+++ b/packages/xod-client/src/core/styles/components/Debugger.scss
@@ -1,9 +1,4 @@
 .debug-session-stop-button {
-  z-index: 10;
-  position: absolute;
-  top: 3px;
-  right: 3px;
-
   padding: 8px 15px 10px 15px;
 
   .fa {
@@ -41,19 +36,19 @@
         margin-left: 7px;
         color: $light-grey-bright;
       }
-  
+
       .progress {
-        
+
         padding-left: 14px;
-  
+
         .progress-trail {
           width: 150px;
           height: 7px;
           border-radius: 7px;
-  
+
           background-color: $dark;
           overflow: hidden;
-  
+
           .progress-line {
             height: 7px;
             background-color: $green;
--- a/packages/xod-client/src/debugger/containers/Breadcrumbs.jsx
+++ b/packages/xod-client/src/debugger/containers/Breadcrumbs.jsx
@@ -8,31 +8,35 @@ import { bindActionCreators } from 'redux';
 import { drillDown } from '../actions';
 import { getRenerableBreadcrumbChunks, getBreadcrumbActiveIndex } from '../../editor/selectors';
 
-const Breadcrumbs = ({ chunks, activeIndex, actions }) => (
-  <ul className="Breadcrumbs Breadcrumbs--debugger">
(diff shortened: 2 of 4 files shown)


packages/xod-client/src/core/styles/components/Debugger.scss
```diff
@@ -1,7 +1,7 @@
 .debug-session-stop-button {
   z-index: 10;
   position: absolute;
-  top: 3px;
+  top: 33px;
```
nkrkv: It looks quite fragile. I think the better solution is allowing `BreadcrumbsContainer` to have children which will be flex-aligned to the right. The same will help when we’d add buttons for the embedded code editor.

What do you think?

packages/xod-client/src/editor/containers/Editor.jsx
```diff
@@ -212,24 +212,26 @@ class Editor extends React.Component {
       null,
       R.ifElse(
         R.propEq('id', DEBUGGER_TAB_ID),
-        () => <Breadcrumbs />,
+        () => <Breadcrumbs>
```
nkrkv: Overcomplication/overnesting detected. Instead of `ifElse`, would a simple ternary read better?

```js
const maybeBreadcrumbs = R.chain(
  tab => (tab.id === DEBUGGER_TAB_ID && this.props.isDebugSessionRunning) ?
    Maybe.of(
      <button
        className="debug-session-stop-button Button Button--light"
        onClick={this.props.stopDebuggerSession}
      >
      <Icon name="stop" /> Stop debug
      </button>
    ) : Maybe.Nothing()
  this.props.currentTab
);

const breadcrumbs = foldMaybe(null, R.identity, maybeBreadcrumbs);
```

brusherru:
@nkrkv I simplified it in latest commit, but in another way. Check it out, please.
If it's ok — I'll squash it into one commit.

packages/xod-client/src/debugger/containers/Breadcrumbs.jsx
```diff
@@ -8,31 +8,38 @@ import { bindActionCreators } from 'redux';
...
+  </div>
 );
 
 Breadcrumbs.propTypes = {
   chunks: PropTypes.arrayOf(PropTypes.object),
   activeIndex: PropTypes.number,
   actions: PropTypes.objectOf(PropTypes.func),
+  children: PropTypes.oneOfType([
+    PropTypes.arrayOf(PropTypes.node),
+    PropTypes.node,
```
evgenykochetkov: `PropTyes.node` is "Anything that can be rendered: numbers, strings, elements or an array (or fragment) containing these types.". No need for `arrayOf`.

PR #1027, opened 2018-01-18
Make main menu items consistent, follow OS guidelines

nkrkv:
No issue, few minor fixes:

- Apply Proper Case to Menu Items
- Add an ellipsis to menu items which open a dialog
- Add “File → Exit” for Linux and Windows

### DIFF
--- a/packages/xod-client-electron/src/view/containers/App.jsx
+++ b/packages/xod-client-electron/src/view/containers/App.jsx
@@ -400,6 +400,12 @@ class App extends client.App {
       submenu,
     } = client.menu;
 
+    // macOS makes Quit item automatically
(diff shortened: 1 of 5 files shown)


packages/xod-client-electron/src/view/containers/App.jsx
```diff
@@ -332,6 +332,10 @@ class App extends client.App {
     );
   }
 
+  onExitClicked() {
+    let w = remoteElectron.getCurrentWindow().close();
```
evgenykochetkov: Not sure why `let w`, though

nkrkv: Yeah, copy-paste blindness. ESLint barked at me. Fixed.

PR #1034, opened 2018-01-18
Add ability to use markdown in comment nodes

### LINKED ISSUE #1028: As a xoder I want markdown-style links in comments so that I can point to cryptic URL’s with readable anchors
### Rationale

Now the only way to place a link into a comment is making it URL-alike. That's a pain if URL is long and contains query args. It would be nice to support Markdown-style hrefs: `...You should [create a new node](https://xod.io/docs/guide/nodes-for-xod-in-xod/)...`

### Acceptance criteria

- [ ] Existing behavior is preserved for plain-URL links like `https://example.com`
- [ ] I can use markdown syntax to hide an URL behind the anchor text

evgenykochetkov:
Closes #1028


nkrkv:
Merge?

PR #1042, opened 2018-01-22
Add a utility to integrate dev-version of desktop IDE with Linux desktop

nkrkv:
Just for DX at the moment.

But also it’s a sample of how file associations are set up on Linux if one day we’ll improve `electron-builder` to support them.


tools/linux-desktop-integrate.sh
```diff
@@ -0,0 +1,62 @@
+#!/bin/bash
+#
+# Run this script to create desktop entry for the development version
+# of XOD IDE. XOD file types will be associated with the development
+# version too.
+#
+
+set -e
+
+# The first is for per-user, the second is system-wide
```
brusherru: Maybe better accept an option "--system -s" to run it with desired paths easily?

nkrkv:
I cancel the PR for a while. A proper association requires using special `xdg-` utilities rather than copying files manually. And the association should be possible right within IDE.

Issue #1036, opened 2018-01-18
Unified welcome patch, ask no questions on IDE start


nkrkv:
### Rationale

When I start the desktop IDE for a second+ time I’m asked to select a project to open. If I’d close the dialog I’ll get IDE in a broken state. The dialog also suggests switching workspace as one of the primary actions. That’s confusing. Let’s get rid of the dialog.

When a xoder starts desktop IDE for the second+ time, he should see the following:

- A new `untitled` project (not bound to a location on FS) is created for him and opened automatically
- All standard libraries are available
- There is an only patch `main` in the project and it is opened in a tab
- On the patch, xoder sees few comments with the general sense:

```
Welcome to XOD again!

Did you complete the built-in tutorial? We recommend you to do it:

»»                                                          ««
»»»»           [Open Built-in Tutorial Project]           ««««
»»                                                          ««

Starting a new project? Press Ctrl+A and Delete to delete this
comment and start tinkering.

Or use “File” menu to open an existing project.
```

### Prerequisites

#1029, #1035 

### Acceptance criteria

- [ ] When I start IDE for the 2+ time I see the welcome patch of an untitled project as described
- [ ] If I hit “File → Save All” I’m asked for the project filename and location
- [ ] Clicking the link opens the tutorial project
- [ ] The patch is panned to center on start

PR #1043, opened 2018-01-24
Unbind project and workspace directories

### LINKED ISSUE #1030: Unbind project and workspace directories
### Rationale

We have a plan to unify xodballs and multifile project and use the native OS dialog to save/load both kinds. In this scenario, it would be very hard to tell a xoder his multifile project must be an immediate child of the workspace directory. It would potentially lead to an infinite amount of wtfs.

So we can allow placing a multifile project anywhere on FS. We need libraries and they are bound to a workspace. OK, let make the workspace a distinct concept. It can be anywhere and by default is at `$HOME/xod`. Such behavior is something like Photoshop does for storing brushes, swatches, etc: a user can change this dir in preferences, but for an average user it does not matter much.

### Acceptance criteria

- [x] I can copy a multifile XOD project anywhere on FS, double-click the `project.xod` and it will open correctly in XOD
- [x] The last active workspace directory will be used for libraries
- [x] By default the workspace directory is `$HOME/xod`
- [x] I can switch the workspace directory like I do already: “File → Select Workspace”¹
- [x] `xodc` supports `--workspace` flag where appropriate, which accepts directory path
- [x] `xodc` uses `XOD_WORKSPACE` env var if `--workspace` is not passed
- [x] `xodc` assumes `$HOME/xod` as the last fallback

¹ Suggest renaming to “File → Switch Workspace” for consistent terminology.

brusherru:
It closes #1030.
If you run IDE by opening any of associated files — you can get popup "Please select Project" sometimes (sometimes not, there is a racing condition now). But I prefer to avoid writing temporary kludge, cause it will be removed later in the #1036.

### DIFF
--- a/packages/xod-cli/src/utils.js
+++ b/packages/xod-cli/src/utils.js
@@ -1,4 +1,5 @@
 import * as R from 'ramda';
+import { expandHomeDir, DEFAULT_WORKSPACE_PATH } from 'xod-fs';
 
 /**
  * Picks a program by command and run it.
@@ -13,6 +14,14 @@ export const runCommand = R.uncurryN(2, options =>
       R.compose(R.equals(true), R.prop(R.__, options)),
       R.tap(() => fn(options))
     )(command)
+  )
+);
+
+// :: Nullable Path -> Path
+export const getWorkspacePath = R.compose(
+  expandHomeDir,
+  R.defaultTo(
+    process.env.XOD_WORKSPACE || DEFAULT_WORKSPACE_PATH
   )
 );
 
--- a/packages/xod-cli/src/xodc-install.js
+++ b/packages/xod-cli/src/xodc-install.js
@@ -3,23 +3,31 @@ import * as xodFs from 'xod-fs';
 import { fetchLibData, fetchLibsWithDependencies } from 'xod-pm';
 import * as messages from './messageUtils';
 import * as MSG from './messages';
+import { getWorkspacePath } from './utils';
+
+// :: Path -> Promise Path Error
+const ensureWorkspacePath = wsPath => new Promise((resolve, reject) => {
+  if (xodFs.isWorkspaceDir(wsPath)) return resolve(wsPath);
+  return reject(
+    new Error(`Directory "${wsPath}" is not a workspace directory. Workspace directory must contain ".xodworkspace" file.`)
+  );
+});
 
 /**
  * Installs the library version from package manager.
  * @param {LibUri} libQuery - Library identifier in the package manager (`owner/name@version`).
- * @param {Path} distPath - Path to workspace, where should be library installed
+ * @param {Path} workspace - Path to workspace, where should be library installed
  * @return {Promise.<void>}
  */
-export default function install(swaggerUrl, libQuery, distPath) {
-  return Promise.all([
-    fetchLibData(swaggerUrl, libQuery),
-    xodFs.findClosestWorkspaceDir(distPath),
-  ])
+export default function install(swaggerUrl, libQuery, workspace) {
+  const wsPath = getWorkspacePath(workspace);
+  return ensureWorkspacePath(wsPath)
+    .then(() => fetchLibData(swaggerUrl, libQuery))
     .then(
-      ([libData, wsPath]) => xodFs.scanWorkspaceForLibNames(wsPath)
-        .then(libNames => [libData, wsPath, libNames])
+      libData => xodFs.scanWorkspaceForLibNames(wsPath)
+        .then(libNames => [libData, libNames])
     )
-    .then(([libData, wsPath, libNames]) => {
+    .then(([libData, libNames]) => {
       messages.notice(MSG.libraryFoundAndInstalling(libQuery));
 
       const params = libData.requestParams;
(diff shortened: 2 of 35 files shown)


packages/xod-client-electron/src/app/utils.js
```diff
@@ -12,3 +15,42 @@ export const errorToPlainObject = R.when(
     R.identity,
   ])
 );
+
+// :: String -> Maybe Path
+const maybePath = R.ifElse(
+  path.isAbsolute.bind(path),
```
nkrkv: The `maybePath` name is misguiding. Why we ever need a voodoo dance with `isAbsolute`? I’m sure there is a reason, but I don’t understand which one. A comment will help.

packages/xod-client-electron/src/app/utils.js
```diff
@@ -12,3 +15,42 @@ export const errorToPlainObject = R.when(
...
+ * It accepts an app and returns a getter function,
+ * that will return `Promise Path Null`.
+ */
+// :: App -> (_ -> Promise Path Null)
+export const getFilePathToOpen = (app) => {
+  // Windows & Linux
+  let pathToOpen = R.compose(
+    R.chain(maybePath),
+    Maybe
+  )(process.argv[1]);
```
nkrkv: Isn’t `process.argv[1]` fragile?

- `electron . /path/to/project`
- `node /path/to/electron /path/to/xod /path/to/project`
- `xod-client-electron /path/to/project`

Does `yarn start:electron -- -- /path/to/project` work on your machine?

brusherru:
@nkrkv check it out, please :)

packages/xod-cli/src/xodc-install.js
```diff
@@ -3,17 +3,20 @@ import * as xodFs from 'xod-fs';
...
  * @return {Promise.<void>}
  */
-export default function install(swaggerUrl, libQuery, distPath) {
+export default function install(swaggerUrl, libQuery, workspace) {
   return Promise.all([
     fetchLibData(swaggerUrl, libQuery),
-    xodFs.findClosestWorkspaceDir(distPath),
+    xodFs.findClosestWorkspaceDir(
+      getWorkspacePath(workspace)
+    ),
```
evgenykochetkov: So we pass a `Path` here, which is defaulted to a default workspace path if it's `null`, and then scan that path and it's parent directories for `.xodworkspace`🤔
Do we really need it? I think something like `R.compose(ensureThatItsAWorkspace, getWorkspacePath)(workspace)` should be enough.

packages/xod-fs/src/save.js
```diff
@@ -71,7 +71,8 @@ export const saveArrangedFiles = R.curry((pathToWorkspace, data) => {
 
 // :: Path -> Project -> Promise Project Error
 export const saveProjectEntirely = R.curry(
-  (workspacePath, project) => Promise.resolve(project)
+  (distPath, project) => Promise.resolve(project)
+    // .then(R.tap(a => console.dir(a, { depth: 2, colors: true })))
```
evgenykochetkov: Sneaky commented out console.log

packages/xod-fs/test/welcomeToXod.spec.js
```diff
@@ -11,7 +11,7 @@ describe('welcome-to-xod', () => {
   const projectPath = 'welcome-to-xod';
 
   it('should load as a valid project', () =>
-    loadProjectWithLibs([], projectPath, workspace)
+    loadProjectWithLibs([workspace], path.resolve(workspace, projectPath))
```
evgenykochetkov: A test with multiple workspaces would be awesome(for example, finding one library in the first workspace, and another in the other one, or "preferring" a version of a library which exists in both workspaces from first workspace, etc), but I am not insisting 😁

brusherru: Added one, but into `loadLib.spec.js`

brusherru:
@nkrkv check it out, please

packages/xod-fs/src/save.js
```diff
@@ -71,7 +71,7 @@ export const saveArrangedFiles = R.curry((pathToWorkspace, data) => {
 
 // :: Path -> Project -> Promise Project Error
 export const saveProjectEntirely = R.curry(
-  (workspacePath, project) => Promise.resolve(project)
+  (distPath, project) => Promise.resolve(project)
```
nkrkv: `dirPath` ?

packages/xod-cli/src/utils.js
```diff
@@ -16,4 +17,12 @@ export const runCommand = R.uncurryN(2, options =>
   )
 );
 
+// :: Nullable Path -> Path
+export const getWorkspacePath = R.compose(
+  expandHomeDir,
+  R.defaultTo(
+    process.env.XOD_WORKSPACE || DEFAULT_WORKSPACE_PATH
```
nkrkv: I found that:

```js
export const DEFAULT_WORKSPACE_PATH = '~/xod/';
```

From my experience keeping trailing slashes while defining dirs lead to huge spaghetti in the future. As a rule of thumb: never end a path component with a slash, let the consumers add one if they want.

Can we fix it?

brusherru: Done

PR #1044, opened 2018-01-24
Add commands to pan patch to origin and center

### LINKED ISSUE #1035: As a xoder I want to pan my patch to origin and center so that I can’t get lost
### Rationale

Regardless of how I used to pan my patch board, I want to be able to reset the position. That’s a mental milestone.

Also, this feature is required to make a nice-aligned welcoming patch.

### Acceptance criteria

- [ ] I can “View → Pan to Origin” to superpose the patch bounding box top left corner with the patch board top left corner (the default padding should be applied so that contents do not stick to the edges)
- [ ] I can “View → Pan to Center” to superpose the patch bounding box center with the patch board center
- [ ] Home key shortcuts to “Pan to Origin”
- [ ] CtrlOrCmd+Home shortcuts to “Pan to Center”

A nice option if it’s trivial to implement with CSS transitions would be an animation with a large constant speed (say, 2000px per second).

evgenykochetkov:
Closes #1035 

Also, adds missing hotkey handlers for ‘New Project’ and ‘Rename Project’


packages/xod-client/src/editor/actions.js
```diff
@@ -197,6 +198,50 @@ export const setCurrentPatchOffset = newOffset => ({
...
+    currentPatch,
+    project,
+    {
+      nodes: XP.listNodes(currentPatch),
+      comments: XP.listComments(currentPatch),
+    }
+  );
+
+  // TODO: too dirty?
+  const patchBoundingRect = document.querySelector('.PatchWrapper-container').getBoundingClientRect();
```
nkrkv: I think so. Imagine, how would you unit-test the action if you’d have to? There will be no DOM. Actions should not touch presentation.

A possible solution is to add viewport rect property to Redux state and oblige the component who owns `.PatchWrapper-container` to keep it up to date.

evgenykochetkov:
@nkrkv @brusherru please check out last two commits, and if everything is ok I'm squashing and merging

packages/xod-client/src/editor/containers/Patch/index.jsx
```diff
@@ -139,6 +140,11 @@ class Patch extends React.Component {
         className={cn('PatchWrapper-container', currentMode)}
         ref={(r) => { this.dropTargetRootRef = r; }}
       >
+        <ReactResizeDetector
+          handleWidth
+          handleHeight
+          onResize={this.props.actions.patchWorkareaResized}
```
brusherru: It's spamming with actions on window resize.
Maybe better to add debounce here?

PR #1046, opened 2018-01-25
"Magic" links that trigger simple IDE actions

### LINKED ISSUE #1029: As a xoder I want create links for simple IDE actions so that I can create an “Open Tutorial” link
### Rationale

At the welcome patch, we’ll need a comment-link which leads to opening the tutorial project. So a mechanism to convert a hyperlink into the action is required.

### Acceptance criteria

- [ ] I can write a comment with markup like `bla bla [Open Tutorial](xod://actions/OPEN_TUTORIAL) bla bla` and it will be rendered as a regular hyperlink
- [ ] Clicking the link is equivalent to “File → Open Tutorial Project”
- [ ] The “File → Open Tutorial Project” replaces the current project with the tutorial spawned from nowhere whatever it means for Electron and browser¹. 
- [ ] The “Open Tutorial Project” item is duplicated in the “Help” menu
- [ ] If I set a `xod://` URL to something unknown/unsupported, executing it will show a snack-bar error like “Bad XOD link” / “URL xod://foo/bar/baz is malformed and can’t be opened”

¹ The tutorial is guaranteed to be in the state bundled with the current version, without any changes applied by a xoder during previous sessions.

### How to implement

- Handle `xod://` URL scheme individually
- `actions` as the first component denote a Redux action should be dispatched
- The second component is the action name
- Query args are used to pass payload

evgenykochetkov:
Closes #1029

The only supported link for the moment is `xod://actions/open-tutorial`


packages/xod-client-browser/src/containers/App.jsx
```diff
@@ -46,7 +48,35 @@ class App extends client.App {
...
     };
 
-    props.actions.openProject(props.initialProject);
+    this.urlActions = {
+      [client.URL_ACTION_TYPES.OPEN_TUTORIAL]: this.onOpenTutorial,
+    };
+
+    document.addEventListener('click', (e) => {
+      if (R.allPass([
+        R.complement(R.isNil),
```
nkrkv: `XF.notNil`

packages/xod-client-browser/src/containers/App.jsx
```diff
@@ -46,7 +48,35 @@ class App extends client.App {
       [client.COMMAND.RENAME_PROJECT]: this.props.actions.requestRenameProject,
     };
 
-    props.actions.openProject(props.initialProject);
+    this.urlActions = {
+      [client.URL_ACTION_TYPES.OPEN_TUTORIAL]: this.onOpenTutorial,
+    };
+
+    document.addEventListener('click', (e) => {
```
nkrkv: This makes the constructor really long. May I suggest extracting the handler to its own method? So here will left something like `document.addEventListener('click', this.onAnchorClick)`?

packages/xod-client-browser/src/containers/App.jsx
```diff
@@ -181,6 +217,7 @@ class App extends client.App {
           onClick(items.renameProject, this.props.actions.requestRenameProject),
           items.separator,
           importProject,
+          onClick(items.openTutorialProject, this.onOpenTutorial),
```
nkrkv: I’m sorry, now I think duplicating the item is not the best idea. I think it would be better if tutorial project access will be in the Help menu only.

packages/xod-client-electron/src/view/containers/App.jsx
```diff
@@ -147,6 +148,23 @@ class App extends client.App {
...
+    };
+    ipcRenderer.on(
+      EVENTS.XOD_URL_CLICKED,
+      (event, { actionName, params }) => {
+        const action = this.urlActions[actionName];
+
+        if (action) {
+          action(params);
+        } else {
+          this.props.actions.addError({ title: 'Invalid action name' });
```
nkrkv: Such phrases should reside in `messages.js` (`packages/xod-client-electron/src/shared/messages.js` in this case). Suggest:

```js
// messages.js
const invalidActionName = action => composeMessage(
  'Bad link',
  `XOD link points to an unknown action '${action}'`
);

// App.jsx
this.props.actions.addError(Messages.invalidActionName(action));
```

packages/xod-client-electron/src/view/containers/App.jsx
```diff
@@ -261,10 +279,14 @@ class App extends client.App {
     this.showPopupProjectSelection();
   }
 
-  onSelectProject(projectMeta) { // eslint-disable-line
+  onSelectProject(projectMeta) { // eslint-disable-line class-methods-use-this
```
nkrkv: :thinking: Isn’t it better to really mark them `static`?

packages/xod-client/src/index.js
```diff
@@ -75,6 +76,7 @@ export * from './popups/constants';
 export { lowercaseKebabMask } from './utils/inputFormatting';
 export { default as sanctuaryPropType } from './utils/sanctuaryPropType';
 export * from './core/urlActions';
+export * from './core/messages';
```
nkrkv: I think in that particular case we loose context when consume API

```js
client.invalidUrlActionName(...) // <- I can’t guess what this thing does
```

Isn’t it a better idea to:

```js
// index.js
export * as Messages from './core/messages'

// App.jsx
client.Messages.invalidUrlActionName(actionName)
```

evgenykochetkov: Done!

PR #1052, opened 2018-01-31
Fix save project (backup pipeline fails on working with symlinks)

brusherru:
There is no issue. 🙈 

Fix and little refactoring of the backup process in the `xod-fs`.

Also detected another bug: when tries to save project many times, before it complete, it occurs bug. It fixed too by preventing start new save process before previous was finished.

### DIFF
--- a/packages/xod-fs/src/backup.js
+++ b/packages/xod-fs/src/backup.js
@@ -1,8 +1,38 @@
+import * as R from 'ramda';
 import fs from 'fs-extra';
 import path from 'path';
 import copy from 'recursive-copy';
 
 const lastDir = dir => dir.split(path.sep).filter(name => name !== '').pop();
+
+// CopyResult :: {
+//   src   :: Path,
+//   dest  :: Path,
+//   stats :: Stats,
+// }
+// :: Path -> Path -> Promise [CopyResult] Error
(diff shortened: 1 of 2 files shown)


nkrkv:
Does not fixed the problem for me

brusherru:
@nkrkv check it out, please!

packages/xod-fs/src/backup.js
```diff
@@ -23,17 +53,18 @@ export class Backup {
   }
 
   make() {
-    return new Promise((resolve, reject) => {
-      if (!this.isDataExist) { resolve('data is not exist'); return; }
-      if (!this.isTempExist) { fs.mkdirSync(this.path.temp); }
-      if (!this.isDataTempExist) { fs.mkdirSync(this.path.dataTemp); }
+    if (!this.isDataExist) {
+      return Promise.reject(
+        new Error('Can not make a backup: Data is not exist')
```
nkrkv: Suggest:

```js
`Backup failed: '${this.path.data}' does not exist`
```

nkrkv:
The `this.setState({ saveInProgress: false })` is very fragile. We need a more bullet-proof solution to the problem.

PR #1053, opened 2018-02-01
Unified welcome patch

### LINKED ISSUE #1036: Unified welcome patch, ask no questions on IDE start

evgenykochetkov:
Closes #1036


packages/xod-client-electron/src/app/utils.js
```diff
@@ -24,31 +23,33 @@ export const errorToPlainObject = R.when(
...
+    ),
+    R.last,
+    R.tail
   )(process.argv);
 
   // MacOS
-  app.on('will-finish-launching', () => {
-    app.on('open-file', (event, filePath) => {
+  app.once('will-finish-launching', () => {
+    app.once('open-file', (event, filePath) => {
```
nkrkv: :+1:

packages/xod-client-electron/src/app/workspaceActions.js
```diff
@@ -91,11 +88,14 @@ export const getDefaultProjectPath = () => resolveDefaultProjectPath(getPathToBu
...
-        FS_ERROR_CODES.WORKSPACE_DIR_NOT_EMPTY,
-      ]),
+      R.both(
+        notNil,
+        isAmong([
+          FS_ERROR_CODES.INVALID_WORKSPACE_PATH,
+          FS_ERROR_CODES.WORKSPACE_DIR_NOT_EXIST_OR_EMPTY,
+          FS_ERROR_CODES.WORKSPACE_DIR_NOT_EMPTY,
+        ]),
+      ),
```
nkrkv: :thinking: why doesn’t it work without the `notNil` check? I think it should. If there’s a problem, I think it’s a problem of `isAmong` implementation and should be fixed there.

packages/xod-client-electron/src/app/utils.js
```diff
@@ -24,31 +23,33 @@ export const errorToPlainObject = R.when(
...
-      () => isDevelopment,
-      R.nth(2), // electron . /optional/path/to/project.xod
-      R.nth(1) // xod.app /optional/path/to/project.xod
-    )
+      R.anyPass([
+        R.isNil,
+        R.startsWith('-'),
+        R.startsWith('data:'),
+        R.equals('.'),
+      ]),
```
nkrkv: A short comment note would be nice.

packages/xod-client-electron/src/app/workspaceActions.js
evgenykochetkov: It's because hm-def rightfully complains that the thing we are trying to find has the wrong type. `[String] -> Null -> Boolean` does not satisfy`[a] -> a -> Boolean`. I'm not sure we should relax it to `[a] -> b -> Boolean`

brusherru: Error that passed here could not contain any `errorCode`, so it fails, cause signature `isAmong :: [a] -> a`, but got `[String] -> undefined`.
But initially, it was thought that any error will be accompanied by an error code, and therefore before this condition was not required. So in a perfect way, we should add error codes everywhere in new parts of pipeline... or easiest way — leave as is.

PR #1056, opened 2018-02-01
Fix backup symlinked projects

brusherru:
There is no issue


brusherru:
This is not actual anymore since we're opening files from anywhere — we'll get a full path to the file, not a path to symlinked project.

PR #1050, opened 2018-01-30
Unified save load UX

brusherru:
It will close #1033.

In addition to the issue, it removes "Rename Project" item from the menu and adds input with validation into "Project preferences".

Also, there are some differences from issue description:
- Don't change project name on Save (even it was empty/untitled and etc)
- Project, by default, could be an empty String
- User can't publish Project without name (he'll get an error message inside Publish Popup)
- In Project Browser group with Users project use the name of the Project, or, if it's empty, "My Project"
- Electron version of IDE will show a Path to the Project in the window title

PR #1060, opened 2018-02-01
Fix save Project as xodball

brusherru:
There is no issue.

Fixes error:
When trying to save the project as a xodball file — it told "Project saved", but it's not.

Appeared in some of PRs after #1050. 😬

### DIFF
--- a/packages/xod-fs/src/save.js
+++ b/packages/xod-fs/src/save.js
@@ -1,3 +1,4 @@
+import os from 'os';
 import path from 'path';
 import * as R from 'ramda';
 import * as XP from 'xod-project';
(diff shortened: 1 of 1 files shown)


packages/xod-fs/src/save.js
```diff
@@ -25,37 +26,37 @@ import { calculateDiff } from './patchDiff';
...
+            path: realRootDir,
+            virtualFile,
           }
         );
       }
-      const isArray = (data instanceof Array);
-      const dataToSave = isArray ? data : [data];
-      const pathToTemp = path.resolve(projectDir, '../.tmp');
-      const backup = new Backup(projectDir, pathToTemp);
+      const isArray = (virtualFile instanceof Array);
```
evgenykochetkov: Please use `Array.isArray` for this

PR #1072, opened 2018-02-05
Add save/load roundtrip tests into `xod-fs` package

brusherru:
No issue.


packages/xod-fs/src/save.js
```diff
@@ -179,8 +179,8 @@ export const saveProjectAsXodball = def(
   (projectPath, project) => {
     const projectDir = path.dirname(projectPath);
     return R.compose(
-      R.always(projectPath),
-      saveVirtualFile(projectDir),
```
evgenykochetkov: I think you can just  change `compose` to `composeP` instead

PR #1084, opened 2018-02-19
Validate variadic markers

### LINKED ISSUE #1064: As a xoder I want my variadic marker nodes to be validated
### Rationale

Making a node variadic imposes few restrictions on what terminals its patch has:

1. `inputCount - arityStep >= outputCount` 
2. `outputCount > 0`
3. There should be no more than a single `variadic-*` marker
4. The types of the middle-inputs which will be later cascaded (indexes `[inputCount - arityStep - outputCount .. inputCount - arityStep - 1]`) must be strictly equal to the types of the outputs

### Prerequisites

#1063 

### Acceptance criteria

- [x] If either of criteria is not satisfied, the `variadic-*` is drawn red (like a dead node)
- [x] A problem node provides tooltip with a reason of it marked so¹
- [ ] An attempt to transpile a program containing variadic marker error fails with a clear error message²

¹ Suggested messages:

- A variadic-2 patch node with 3 outputs should have at least 5 inputs
- A variadic patch node should have at least one output
- A patch node is allowed to have at most one `variadic-*` marker
- Types of inputs `X`, `IN1` should match the types of outputs `OUT1`, `OUT2`

²
> **Program contains invalid variadic markers**
> Patch @/main -> @/foo -> @/bar contains an invalid variadic-2 node
> A variadic patch node should have at least one output
> Fix the patch to continue

### How to implement

Make functions `validatePatchVariadics :: Patch -> Either Error Patch` and `validateProjectVariadics :: Project -> Either Error Project`. Use them to validate a patch on nodes add/remove, and validate a project before transpiling.

### How to test

Make unit tests for all of the cases mentioned in rationale.

brusherru:
It closes #1064.

But latest one item from acceptance criteria could not be finished in ideal way, cause we need to do a massive refactoring over "dead reference errors", flatten Patch tree and validate leaves and etc.

Error is appearing, but it looks like:
```
Patch @/01-hello contains dead references:
A variadic patch should have at least one output
Fix or delete them to continue.
````

### DIFF
--- a/packages/xod-client/src/project/components/Node.jsx
+++ b/packages/xod-client/src/project/components/Node.jsx
@@ -86,6 +86,12 @@ class Node extends React.Component {
 
     const isTerminalNode = XP.isTerminalPatchPath(type);
 
+    const errTitle = (this.props.dead && this.props.errors.length > 0)
+      ? R.compose(
+        R.join(';\n'),
+        R.map(R.prop('message'))
+      )(this.props.errors) : null;
+
     return (
       <svg
         key={id}
@@ -101,10 +107,10 @@ class Node extends React.Component {
           onDoubleClick={this.onDoubleClick}
           id={id}
           data-label={nodeLabel} // for func tests
-          title={nodeLabel}
+          title={errTitle || nodeLabel}
         >
           {this.renderBody()}
-          {!this.props.isDragged ? <title>{nodeLabel}</title> : null}
+          {!this.props.isDragged ? <title>{errTitle || nodeLabel}</title> : null}
         </g>
         <g className="pins" id={`nodePins_${id}`}>
           {pinsArr.map(pin =>
@@ -139,6 +145,9 @@ Node.propTypes = {
   size: PropTypes.any.isRequired,
   position: PropTypes.object.isRequired,
   dead: PropTypes.bool,
+  errors: PropTypes.arrayOf(
+    PropTypes.instanceOf(Error)
+  ),
   isSelected: PropTypes.bool,
   isGhost: PropTypes.bool,
   isDragged: PropTypes.bool,
@@ -152,6 +161,7 @@ Node.propTypes = {
 };
 
 Node.defaultProps = {
+  errors: [],
   dead: false,
   isSelected: false,
   isGhost: false,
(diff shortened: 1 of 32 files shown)

### CODE REVIEW

File: packages/xod-client/src/project/components/Node.jsx [hunk 1]
```diff
@@ -86,6 +86,12 @@ class Node extends React.Component {
 
     const isTerminalNode = XP.isTerminalPatchPath(type);
 
+    const errTitle = (this.props.dead && this.props.errors.length > 0)
+      ? R.compose(
+        R.join(';\n'),
+        R.map(R.prop('message'))
```
nkrkv: Can be shortened to `R.pluck('message')`

packages/xod-client/src/project/components/Node.jsx
```diff
@@ -100,10 +106,11 @@ class Node extends React.Component {
           onMouseUp={this.onMouseUp}
           onDoubleClick={this.onDoubleClick}
           id={id}
-          title={nodeLabel} // this is for func-tests
+          data-label={nodeLabel} // for func tests
+          title={errTitle || nodeLabel}
         >
           {this.renderBody()}
-          {!this.props.isDragged ? <title>{nodeLabel}</title> : null}
+          {!this.props.isDragged ? <title>{errTitle || nodeLabel}</title> : null}
```
nkrkv: Not sure `errTitle || nodeLabel` is a proper decision.

1. Code duplication with a line above
2. `nodeLabel` is inaccessible in case there are errors

Could we provide a better experience with rich tooltips which show the base name (if overriden) *and* the error (if present)?

packages/xod-client/src/project/selectors.js
```diff
@@ -133,6 +133,26 @@ const addDeadFlag = R.curry(
...
+  (patch, renderableNode) => R.when(
+    R.compose(
+      XP.isVariadicPath,
+      R.prop('type')
+    ),
+    node => R.compose(
+      foldEither(
+        err => R.compose(
+          R.assoc('errors', [err]),
+          R.assoc('dead', true)
```
nkrkv: What “dead” does here? I thought that a separate error class: dead references is one thing and variadic marker misuse is another.

packages/xod-project/src/patch.js
```diff
@@ -1220,13 +1220,19 @@ export const listSharedInputPins = def(
 );
 
 /**
- * It computed a variadic pins and validates Patch.
- * It returns Either Patch or Validation Error.
+ * It checks Patch for existing of variadic marker.
```
nkrkv: _Checks a patch for variadic marker existence_

PR #1089, opened 2018-02-26
Expand variadic nodes

### LINKED ISSUE #1065: Expand variadic nodes
A variadic node has an `arityLevel` which can be anything from one to infinity. `arityLevel = 1` means the node has exactly the same number of inputs as the number of input terminals in its patch, i.e. it is not changed. That’s a default level for all nodes. `arityLevel = 2` means it gets extra `arityStep` inputs, `arityLevel = N` means it gets extra `arityStep × (N - 1)` pins.

To make a variadic node useful it should expand to a cascade of the first arity-level nodes:

```
Shared              Level 1     Level 2       Level 3

  +       +    +    +    +      +    +        +     +
  |       |    |    |    |      |    |        |     |
  |       |    |    |    |      |    |        |     |
  |       |    |    |    |      |    |        |     |
  |       |    |    |    |      |    |        |     |
  +----+--+----+----+----+--+   |    |        |     |
  |    |                    |   |    |        |     |
  |    |                    |   |    |        |     |
  |    |                    |   |    |        |     |
  |    +------------+----+--+   |    |        |     |
  |                 |    |      |    |        |     |
  |                 |    |      |    |        |     |
  +-----------+-----+----+------+----+----+   |     |
  |           |                           |   |     |
  |           |                           |   |     |
  |           |                           |   |     |
  |           +-----------------+----+----+   |     |
  |                             |    |        |     |
  +------------------------+----+----+--------+-----+--+
                           |                           |
                           |                           |
                           |                           |
                           +---------------------------+
```

- Extraneous leftmost inputs are forwarded as is to all nodes in the cascade
- Outputs go into middle inputs
- N-th arity level inputs go to rightmost inputs of the n-th level in the cascade

The feature involves introducing “variadic input pins” which do not correspond to a terminal node but appeared as a consequence of arity level scale. Let’s use original ID + a unique delimiter + the level index for its IDs, e.g. `0124abcd...2` (three dots is the delimiter, 2 is the index).

The result of implementing the feature should be the core which supports variadic nodes for :100: 
The only thing left to do is IDE UI support.

### Prerequisites

#1063 

### Acceptance criteria

- [ ] I prepare a `variadic-1` node named `var-add` and set its `arityLevel` to 3 in JSON with a text editor, link the last input to a constant 42, bind 5, 13, 10 to other inputs (in JSON), connect it to a watch, upload and see 70 as the result.
- [ ] I can reproduce a similar, but complex scenario with a select-alike node (arity step is 2) which also has a leftmost input for the initial value (links to all).

### How to implement

- Make an `expandVariadics` function, insert it before flatten in the transpilation pipeline.
- Create transient patches with the cascades

evgenykochetkov:
Closes #1065

### DIFF
--- a/packages/xod-project/src/expandVariadicNodes.js
+++ b/packages/xod-project/src/expandVariadicNodes.js
@@ -1,0 +1,306 @@
+import * as R from 'ramda';
+import { explodeMaybe, explodeEither } from 'xod-func-tools';
+
+import { def } from './types';
+
+import * as Pin from './pin';
+import * as Node from './node';
+import * as Link from './link';
+import * as Patch from './patch';
+import * as Project from './project';
+import { TERMINAL_PIN_KEYS, PIN_DIRECTION } from './constants';
+import { getExpandedVariadicPatchPath } from './patchPathUtils';
+
+//
+// expanding a single patch
+//
+// @see https://raw.githubusercontent.com/wiki/xodio/xod/images/illustrations-for-xod-source-code/xod-project/expandVariadicNodes.png
+
+const getNodeX = R.pipe(Node.getNodePosition, R.prop('x'));
+
+const nodeIdLens = R.lens(Node.getNodeId, R.assoc('id'));
+
+
+// helpers for creating nodes inside expanded patch
+
+const createAdditionalValueTerminalGroups = (
+  patch,
+  desiredArityLevel,
+  originalTerminalNodes,
+  variadicPins,
+) => {
+  const arityStep = R.compose(
+    explodeMaybe('Patch is guaranteed to be variadic at this point'),
+    Patch.getArityStepFromPatch
+  )(patch);
+
+  // for positioning new terminals
+  const rightmostInputTerminalX = R.compose(
+    xs => Math.max(...xs),
+    R.map(getNodeX),
+    R.filter(Node.isInputPinNode)
+  )(originalTerminalNodes);
+
+
+  // :: [ Node ]
+  const originalValueTerminalNodes = R.compose(
+    R.sortBy(getNodeX),
+    R.map(R.compose(
+      R.flip(Patch.getNodeByIdUnsafe)(patch),
+      Pin.getPinKey,
+    )),
+    R.prop('value'),
+  )(variadicPins);
+
+  const DISTANCE_BETWEEN_TERMINALS = 50;
+  const valueTerminalsGroupWidth = arityStep * DISTANCE_BETWEEN_TERMINALS;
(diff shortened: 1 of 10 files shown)


packages/xod-project/src/internal/patchPathUtils.js
```diff
@@ -1,6 +1,15 @@
...
 
+const EXPANDED_PATCH_NAME_SUFFIX = '-$';
+
+export const getExpandedPatchPath = R.curry(
+  (arityLevel, originalPath) => [
+    originalPath,
+    EXPANDED_PATCH_NAME_SUFFIX,
+    arityLevel,
+  ].join('')
+);
```
nkrkv: The `packages/xod-project/src/internal/patchPathUtils.js` gives no idea we’re talking about variadics here. So I’d call the function more verbose: `getExpandedVariadicPatchPath

packages/xod-project/src/expandVariadicNodes.js
```diff
@@ -0,0 +1,344 @@
...
+
+  return getDependenciesMap(
+    project,
+    R.concat(restPatchPaths, currentPatchDeps),
+    R.assoc(currentPatchPath, currentPatchDeps, depsMap)
+  );
+};
+
+// :: PatchPath -> Project -> [PatchPath]
+export const getPatchDependencies = R.curry(
```
nkrkv: Not sure this should live in `expandVariadicNodes.js`

packages/xod-project/src/expandVariadicNodes.js
```diff
@@ -0,0 +1,344 @@
...
+  );
+
+  // :: [ [Node] ]
+  return R.compose(
+    R.map(terminalGroupIndex => R.addIndex(R.map)((node, index) =>
+      R.compose(
+        R.over(
+          nodeIdLens,
+          // TODO: extract to a public util. where to put it?
+          nodeId => `${nodeId}-$${terminalGroupIndex}`
```
nkrkv: I think no. It is an internal business of variadic expansion.

packages/xod-project/src/expandVariadicNodes.js
```diff
@@ -0,0 +1,344 @@
...
+      )(node),
+      originalValueTerminalNodes
+    )),
+    R.range(2), // [2; desiredArityLevel]
+    R.inc
+  )(desiredArityLevel);
+};
+
+// TODO: better name?
+const createExpandedNodeInstances = (patch, desiredArityLevel) => {
```
nkrkv: `createExpansionNodes`? Because a node is already === an instance (of the patch).

Issue #1069, opened 2018-02-02
Describe variadic nodes and arity concepts in docs


nkrkv:
There should be guide articles on:

- Using variadic nodes: what is arity, how they are expanded
- Creating a variadic node: marker nodes, requirements for terminals

PR #1095, opened 2018-02-28
Implement xod/core/join

nkrkv:
A useful node per se very useful for illustrative purposes of #1069


workspace/__lib__/xod/core/join/patch.xodp
```diff
@@ -0,0 +1,107 @@
...
+    {
+      "id": "HJrBNr4dG",
+      "position": {
+        "x": 68,
+        "y": 306
+      },
+      "type": "xod/patch-nodes/output-string"
+    },
+    {
+      "arityLevel": 2,
```
evgenykochetkov: Constructing variadics using other variadics. Nice!

PR #1100, opened 2018-03-01
Show if a pin is variadic in help panel

evgenykochetkov:


packages/xod-client/src/editor/components/PatchDocs.jsx
```diff
@@ -101,9 +104,24 @@ OutputPins.defaultProps = {
 
 
 const PatchDocs = ({ patch, minimal }) => {
+  const variadicPinKeys = R.compose(
```
brusherru: Maybe this function should be in the `xod-project`?

packages/xod-client/src/editor/components/PatchDocs.jsx
```diff
@@ -101,9 +104,24 @@ OutputPins.defaultProps = {
...
+      R.always([]),
+      R.map(XP.getPinKey),
+    ),
+    XP.listVariadicValuePins
+  )(patch);
+
   const [inputPins, outputPins] = R.compose(
     R.partition(XP.isInputPin),
     XP.normalizePinLabels,
+    R.map(
```
brusherru: Why not do it in the `xod-project`?

PR #1109, opened 2018-03-05
Add prettier

evgenykochetkov:
The whole thing is set up with [eslint-plugin-prettier](https://github.com/prettier/eslint-plugin-prettier)
Just use the good old `yarn lint` to check and `yarn lint --fix` to format


evgenykochetkov:
UPD: changed `trailingComma` setting to `es5` for aestetic reasons and because it produces a smaller diff

nkrkv:
OK

PR #1128, opened 2018-03-14
Legalize parametric patch names

### LINKED ISSUE #1114: Legalize parametric patch names
### Rationale

To support generics we should allow patch names which define the type(s) they specialize for:

- `if-else(number)`
- `if-else(string)`
- `cast(boolean,pulse)`

### Acceptance criteria

The following are legal and can be created via IDE:

- [ ] `if-else(number)`
- [ ] `cast(boolean,pulse)`

The following is not:

- [ ] `if-else(number`
- [ ] `if-else ( number )`
- [ ] `if-else)`
- [ ] `if-else()`
- [ ] `if-else(number)string`

### Out of scope

- Any special meaning for the patches with parenthesis in their name.
- Fancy rendering of parametric patch names

brusherru:
It closes #1114.

Also tweaked rename popup. Now input already contains current patch name and autoselected.

### DIFF
--- a/packages/xod-project/src/internal/patchPathUtils.js
+++ b/packages/xod-project/src/internal/patchPathUtils.js
@@ -8,13 +8,44 @@ export const getExpandedVariadicPatchPath = R.curry(
     [originalPath, EXPANDED_PATCH_NAME_SUFFIX, arityLevel].join('')
 );
 
+// =============================================================================
+// RegExp parts
+// =============================================================================
+
+// foo2
+const alphanumeric = '[a-z0-9]+';
+
+// foo-2-bar
+const alphanumericWithHypens = `${alphanumeric}(-${alphanumeric})*`;
+
+// foo-2-bar,foo-5-bar
+const alphanumericWithHypensAndCommans = `${alphanumeric}(-${alphanumeric}|,${alphanumeric})*`;
+
+// -$5
+const variadicLevel = '-\\$\\d';
+
+// (foo-2-bar,foo-5-bar)
+const types = `(\\(${alphanumericWithHypensAndCommans}\\)){0,1}`;
+
+// foo2(foo-2-bar,foo-5-bar)-$5
+const patchBaseNameRegExp = new RegExp(
+  `^${alphanumericWithHypens}(${types}|${variadicLevel})$`
+);
+
+// foo-2-bar
+const identifierRegExp = new RegExp(`^${alphanumericWithHypens}$`);
+
+// =============================================================================
+// Validating functions
+// =============================================================================
+
 // :: String -> Boolean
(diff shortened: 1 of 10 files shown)


packages/xod-project/src/internal/patchPathUtils.js
```diff
@@ -8,13 +8,31 @@ export const getExpandedVariadicPatchPath = R.curry(
     [originalPath, EXPANDED_PATCH_NAME_SUFFIX, arityLevel].join('')
 );
 
+// =============================================================================
+// RegExp parts
+// =============================================================================
+const alphanumeric = '[a-z0-9]+';
+const alphanumericWithHypens = `${alphanumeric}(-${alphanumeric})*`;
+const alphanumericWithHypensAndCommans = `${alphanumeric}(-${alphanumeric}|,${alphanumeric})*`;
+const variadicLevel = '-\\$\\d';
```
nkrkv: I love the idea, but let’s remove this easter egg until we’ll make the resolution for the different variadic levels?

packages/xod-project/src/internal/patchPathUtils.js
```diff
@@ -8,13 +8,31 @@ export const getExpandedVariadicPatchPath = R.curry(
...
+const types = `(\\(${alphanumericWithHypensAndCommans}\\)){0,1}`;
+
+const patchBaseNameRegExp = new RegExp(
+  `^${alphanumericWithHypens}(${types}|${variadicLevel})$`
+);
+const identifierRegExp = new RegExp(`^${alphanumericWithHypens}$`);
```
nkrkv: Cool readable regex constructor :+1:

packages/xod-project/test/patchPathUtils.spec.js
```diff
@@ -8,6 +8,11 @@ import * as Helper from './helpers';
 
 chai.use(dirtyChai);
 
+// Kludge to use instead of "assert.isTrue" / "assert.isFalse"
+// Cause dirtyChai makes it works wrong
+const isTrue = a => assert.equal(a, true);
+const isFalse = a => assert.equal(a, false);
```
nkrkv: I filed #1129. To drop the kludge easily once the refactoring will land, maybe it is better to over-monkey-patch `assert.isTrue` and `assert.isFalse` right here and remove the `dirtyChai` from this particular module?

If agree, lets make it and leave a TODO referring the #1129

packages/xod-project/src/internal/patchPathUtils.js
brusherru: It's just about PatchPath type, to validate flattened variadic nodes.
Early validation of patch paths did not work and so this rule appeared only now.

PR #1132, opened 2018-03-16
Allow to create abstract nodes

### LINKED ISSUE #1116: As a xoder I want to create abstract nodes so that I can define interface for future specializations
### Rationale

An end-user wants to simply place `if-else` node regardless of the concrete type of its arguments, not the `if-else(number)` or `if-else(string)`. To make it possible we introduce a concept of the abstract node.

Later it will resolve to a particular specialization or throw an error. But before that, we should be able to create it.

To make a node abstract, a xoder places `abstract` marker node on the patch node and at least one terminal of generic type `t1`, `t2`, `t3`. If generic type terminals are absent, that’s an error which is shown in a tooltip like it is done for `variadic-*` nodes.

### Acceptance criteria

- [ ] I can place the `abstract` marker node
- [ ] I can place `input(t1)`, `output(t1)`, `input(t2)`, `output(t2)`, `input(t3)`, `output(t3)` terminal nodes
- [ ] The generic terminal nodes render the same way as regular terminals do and they are colored gray
- [ ] The generic pins render the same way as regular pins do and they are colored gray
- [ ] If a patch node has `abstract` and no inputs of `t1`, the marker is drawn in red and shows “At least one generic `input(t1)` terminal is required”
- [ ] If a patch node has `abstract`, `input(t1)`, `input(t3)`, but no `input(t2)`, the marker errors with “Generic inputs t1, t2, t3 should be employed sequentially. Add t2 or remove t3”
- [ ] Documentation boxes show “generic t1” as the type label for generic pins

### Out of scope

Generic pins linking may not be supported

evgenykochetkov:
Closes #1116

### DIFF
--- a/packages/xod-client/src/core/styles/components/Pin.scss
+++ b/packages/xod-client/src/core/styles/components/Pin.scss
@@ -21,8 +21,12 @@
 .Pin {
   .symbol {
     fill: $color-pin-fill;
-    stroke: $color-pin-fill; // pins borders are colored by type and don't have a 'default' color.
+    stroke: $color-datatype-generic;
     stroke-width: 2px;
+
+    &.is-connected {
+      fill: $color-datatype-generic;
+    }
 
     &.string {
       @include symbol-coloring($color-datatype-string);
--- a/packages/xod-client/src/editor/components/inspectorWidgets/index.js
+++ b/packages/xod-client/src/editor/components/inspectorWidgets/index.js
@@ -6,7 +6,7 @@ import BoolWidget from './pinWidgets/BoolPinWidget';
 import NumberWidget from './pinWidgets/NumberPinWidget';
 import PulseWidget from './pinWidgets/PulsePinWidget';
 import StringWidget from './pinWidgets/StringPinWidget';
-import DeadPinWidget from './pinWidgets/DeadPinWidget';
+import DisabledInputWidget from './pinWidgets/DisabledInputWidget';
 import IOLabelWidget from './IOLabelWidget';
 import DescriptionWidget from './DescriptionWidget';
 import LabelWidget from './LabelWidget';
@@ -76,7 +76,7 @@ const WIDGET_MAPPING = {
     props: { type: 'string' },
   },
   [WIDGET_TYPE.DEAD]: {
-    component: DeadPinWidget,
+    component: DisabledInputWidget,
     props: {
       type: PIN_TYPE.DEAD,
     },
@@ -86,7 +86,7 @@ const WIDGET_MAPPING = {
 export const getNodeWidgetConfig = type =>
   isGenericType(type)
     ? {
-        component: DeadPinWidget,
+        component: DisabledInputWidget,
         props: { type },
       }
     : WIDGET_MAPPING[type];
--- a/packages/xod-client/src/editor/components/inspectorWidgets/index.js
+++ b/packages/xod-client/src/editor/components/inspectorWidgets/index.js
@@ -1,5 +1,5 @@
 import * as R from 'ramda';
-import { PIN_TYPE } from 'xod-project';
+import { PIN_TYPE, isGenericType } from 'xod-project';
 
 import HintWidget from './HintWidget';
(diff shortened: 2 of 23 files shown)


packages/xod-client/src/core/styles/components/Pin.scss
```diff
@@ -21,9 +21,13 @@
 .Pin {
   .symbol {
     fill: $color-pin-fill;
-    stroke: $color-pin-fill; // pins borders are colored by type and don't have a 'default' color.
+    stroke: $color-datatype-generic;
```
nkrkv: Minus kludge :+1:

packages/xod-client/src/editor/components/inspectorWidgets/index.js
```diff
@@ -42,47 +42,57 @@ const submitOnEnter = {
...
+      type: PIN_TYPE.DEAD,
     },
-    [WIDGET_TYPE.DEAD]: {
+  },
+};
+
+export const getNodeWidgetConfig = type => {
+  if (isGenericPinType(type)) {
+    return {
       component: DeadPinWidget,
```
nkrkv: Naming bore: not sure using the *dead* pin widget is a good idea for generics. Better generalize it to something like `DisabledInputWidget`.

packages/xod-client/src/editor/components/inspectorWidgets/index.js
```diff
@@ -42,47 +42,57 @@ const submitOnEnter = {
...
       props: {
-        type: PIN_TYPE.DEAD,
+        type,
       },
-    },
-  },
+    };
+  }
+  return WIDGET_MAPPING[type];
```
nkrkv: Suspect a ternary will make the whole function more elegant:

```javascript
export const getNodeWidgetConfig = type =>
  isGenericPinType(type)
  ? {
    component: DisabledInputWidget,
    props: { type }
  }
  : WIDGET_MAPPING[type];
```

packages/xod-project/src/utils.js
```diff
@@ -102,6 +102,8 @@ export const generateId = shortid.generate;
  */
 export const validateId = R.test(/^[a-zA-Z0-9\-_]+$/);
 
+export const isGenericPinType = R.test(/^t[1-3]$/);
```
nkrkv: Will shortening to `isGenericType` will introduce any confusion? I think `isGenericType` better reflects the essence.

packages/xod-client/src/editor/components/inspectorWidgets/pinWidgets/PinWidget.jsx
```diff
@@ -18,12 +18,19 @@ const isNonBindableOutput = R.both(
   R.complement(R.prop('isBindable'))
 );
 
-const isBindingForbidden = R.either(isLinkedInput, isNonBindableOutput);
+const isGenericPin = R.pipe(R.prop('dataType'), isGenericPinType);
```
nkrkv: Don’t we have `XP.getPinType`? I think we should.

packages/xod-client/src/project/selectors.js
```diff
@@ -175,6 +175,16 @@ const addVariadicErrors = R.curry((patch, renderableNode) =>
   )(renderableNode)
 );
 
+// :: Patch -> RenderableNode -> RenderableNode
+const addAbstractPatchErrors = R.curry((patch, renderableNode) =>
+  R.when(R.compose(R.equals(XP.ABSTRACT_MARKER_PATH), R.prop('type')), node =>
```
nkrkv: Expect a getter instead of `R.prop('type')`

packages/xod-client/src/editor/components/inspectorWidgets/pinWidgets/PinWidget.jsx
evgenykochetkov: We have it, but it looks at `type` prop.

evgenykochetkov:
@nkrkv Done!

PR #1136, opened 2018-03-20
Use parametric patch names for cast and defer nodes

evgenykochetkov:
Continues #1114 and #1128

### DIFF
--- a/packages/xod-project/src/patchPathUtils.js
+++ b/packages/xod-project/src/patchPathUtils.js
@@ -158,10 +158,19 @@ export const getCastPatchPath = (typeIn, typeOut) =>
 // defer-* nodes
 //
 
-const deferNodeRegExp = new RegExp(`xod/core/defer-(${dataTypes.join('|')})$`);
+const legacyDeferNodeRegExp = new RegExp(
+  `^xod/core/defer-(${dataTypes.join('|')})$`
+);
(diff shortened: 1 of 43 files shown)


packages/xod-project/src/patchPathUtils.js
```diff
@@ -137,12 +137,20 @@ export const getVariadicPath = n => `${PATCH_NODES_LIB_NAME}/variadic-${n}`;
...
 //
 
-const castTypeRegExp = new RegExp(
+const legacyCastTypeRegExp = new RegExp(
   `xod/core/cast-(${dataTypes.join('|')})-to-(${dataTypes.join('|')})$`
 );
+// TODO: When custom types will be added this should be generalized.
+//       Also, casting from/to generic types should be forbidden.
+const castTypeRegExp = new RegExp(
+  `xod/core/cast\\((${dataTypes.join('|')}),(${dataTypes.join('|')})\\)$`
```
brusherru: What about begin of the line in the regular expressions?

evgenykochetkov: Good point. Fixed.

evgenykochetkov: + updated regexps for checking constant-*, variadic-* and internal terminals

PR #1137, opened 2018-03-20
Node deprecation feature

### LINKED ISSUE #1124: As a library author I want deprecate some nodes so that its users are discouraged from their use
### Rationale

Libraries evolve, some initial design decisions become obsolete. To encourage the “right” usage and discourage an obsolete design we should introduce a deprecation mechanism.

To make a node deprecated a library author places a `xod/patch-nodes/deprecated` marker node. The reason for deprecation he puts into the marker description.

A library user does not see the deprecated nodes by default. Although they can be listed by setting up Project Browser filter. The filter icon is the same as seen on the deployment pane. The item in the project browser gets a little “dp” icon.

01---01-main

Deprecated nodes demonstrate the fact in their help boxes with a red note *before* the description body. Also, they render exclamation icon on their body (as on `absolute`).

02---onpatch

For the sake of testing the feature the following standard library nodes may be deprecated safely:

- xod/core/concat-3
- xod/core/concat-4
- xod/core/concat-5
- xod/core/concat-6
- xod/core/discretize-2
- xod/core/discretize-3
- xod/core/discretize-4
- xod/core/nth-number-2
- xod/core/nth-number-3
- xod/core/nth-number-4

### Acceptance criteria

- [x] I can place a `deprecated` marker on the `concat-3` node and provide the description “Use `xod/core/concat` with adjusted arity instead”
- [x] Now, when I work on a regular project and expand `xod/core` the `concat-3` is not listed
- [x] If I enter the `concat` query in the quick search, the `concat-3` would not be listed
- [x] I can hit the Project Browser’s Filter button and check the “Deprecated nodes” item
- [x] The `concat-3` would be listed in the project browser with a deprecation icon
- [x] A help popup for the `concat-3` shows the note “Deprecated: Use `xod/core/concat` with adjusted arity instead” right before the short description
- [x] An instance of the `concat-3` node shows “Deprecated” as the first line of the node tooltip and renders an exclamation icon
- [x] All the nodes from `xod/core` listed above are deprecated

brusherru:
It closes #1124


packages/xod-client/src/projectBrowser/containers/ProjectBrowser.jsx
```diff
@@ -283,6 +297,17 @@ class ProjectBrowser extends React.Component {
         title="Add library"
         onClick={this.onClickAddLibrary}
       />,
+      <ContextMenuTrigger
+        id="AAA"
```
evgenykochetkov: Why "AAA"?

brusherru: Oops, forgot to change it 😬

packages/xod-client/src/project/components/nodeParts/RegularNodeBody.jsx
```diff
@@ -20,8 +20,9 @@ const RegularNodeBody = props => (
     <rect className="body" {...NODE_BODY_RECT_PROPS} />
     <NodeLabel
       text={props.label || XP.getBaseName(props.type)}
-      width={props.size.width}
+      width={props.size.width - 4}
       height={props.size.height}
+      x={2}
```
nkrkv: What are these? Isn’t it left after the exclamation icon?

packages/xod-project/src/patch.js
```diff
@@ -1011,6 +1011,27 @@ export const resolveNodeTypesInPatch = patch =>
     listNodes
   )(patch);
 
+/**
+ * Gets a deprecated marker description.
+ */
+export const getDeprecationPatchReason = def(
```
nkrkv: Suggest `getPatchDeprecationReason` or even `getDeprecationReason`

packages/xod-client/src/project/components/nodeParts/RegularNodeBody.jsx
brusherru: Nope. Moved it into a constant and left comment.
It prevents NodeLabel from sticking to the Node outline.

brusherru:
@nkrkv I've fixed the performance issue. Check it out, please

PR #1140, opened 2018-03-23
Switch Node Specialization from Inspector

### LINKED ISSUE #1130: As a xoder I want to select specialization/abstraction for a node in Inspector so that I can switch between them quickly
### Rationale

There are cases when a xoder is forced to choose a particular specialization for a generic node:

- Incoming link types are in conflict
- The user wants to bind generic values rather than link them

We can add a drop-down to the inspector to make it possible.

Imagine we have `if-else`, `if-else(number)`, `if-else(string)`. And the `if-else(number)` is duplicated in `foo/bar` For any of these nodes, the inspector should provide the drop-down “Type” with the same items:

- “Generic (auto)”
- “number (foo/bar)”
- “number (xod/core)”
- “string (xod/core)”

If it was a node with two generic types:

- “Generic (auto)”
- “number, string (foo/bar)”
- “number, string (xod/core)”
- “string, string (xod/core)”

Selecting an item physically change the node type (patch path). If its links become invalid they should be kept: links would render errors using the existing validation mechanisms.

### Acceptance criteria

Given the example of `if-else` above:

- [x] I can place `if-else` and see the drop-down with the four items and the first one selected
- [ ] I can link a `constant-string` to a generic input, the “Generic (auto)” selection is kept
- [x] I can select “string (xod/core)” and the type for the node will become `xod/core/if-else(string)`
- [x] The drop-down still enumerates the same four items with “string (xod/core)” selected
- [x] I can place a `xod/core/sin` node and it will not show the “Type” drop-down at all

brusherru:
It closes #1130


brusherru:
@nkrkv, check it out, please.

I've updated option labels in type specialization dropdown:
`Auto-detect (xod/core)` -> `Auto-detect type (xod/core)`
`number (xod/core)` -> `Type: number (xod/core)`
and etc

packages/xod-client/src/project/reducer.js
```diff
@@ -316,6 +316,33 @@ export default (state = {}, action) => {
       );
     }
 
+    case AT.NODE_CHANGE_SPECIALIZATION: {
+      const { nodeId, patchPath, nodeType } = action.payload;
+      const currentPatchLens = XP.lensPatch(patchPath);
+
+      // TODO: Replace these composition with smarter function
+      // from `xod-project`, that will switch NodeType and drop
+      // bound values only in pins, that changed their types.
```
evgenykochetkov: 👍🏻

PR #1141, opened 2018-03-23
Add utility markers

### LINKED ISSUE #1139: As a library author I wanna mark some patch nodes as utilities so they are not exposed to an end-user
### Rationale

Often, usable and useful nodes are composed of several not-so-friendly internal *utilities*. The utilities clutter the node list and lead to confusion for end-users. We can make a mechanism to hide such nodes by default.

The mechanism is very similar to node deprecation (#1124) with a little difference in presentation. Having utility nodes on a patch is not a problem, thus they should not render any highligts.

As as the first practical use of the feature, the following nodes from the standard library may be hidden:

- `xod/common-hardware/adxl335-convert`
- `xod/common-hardware/dht2x-pack`
- `xod/common-hardware/dhtxx-read-raw`
- `xod/common-hardware/gp2y0a-linearize`
- `xod/common-hardware/st-imu-generic-sensor`
- `xod/common-hardware/st-imu-normalize-acc`
- `xod/common-hardware/st-imu-normalize-va`
- `xod/common-hardware/st-imu-round-sensitivity`
- `xod/core/cast(*,*)`

### Acceptance criteria

- [ ] I can place a `xod/patch-nodes/utility` marker on the `xod/common-hardware/dht2x-pack`
- [ ] The `dht2x-pack` is not listed in the project browser by default
- [ ] I can enlist utility nodes by hitting the filter button → Utility nodes
- [ ] When enabled, the `xod/common-hardware/dht2x-pack` renders a soft-color “UT” icon next to the item
- [ ] The `dht2x-pack` is not listed in the quick search results
- [ ] Utility nodes for the current project are always listed in the project browser and suggester

brusherru:
It closes #1139 
And marked old-fashioned `cast` and `defer` nodes in the `xod-core` as deprecated.

Please, pay attention to:
- deprecation messages
- how nodes like `cast-boolean-to-pulse` looks in the Project Browser (it has two markers: deprecated and utility)


workspace/__lib__/xod/core/cast-boolean-to-pulse/patch.xodp
```diff
@@ -1,6 +1,23 @@
...
+    {
+      "id": "H19x_iG9G",
+      "position": {
+        "x": 170,
+        "y": 102
+      },
+      "type": "xod/patch-nodes/utility"
+    },
+    {
+      "description": "Use xod/core/cast or its specialization.",
```
nkrkv: I would say we don’t want to even encourage the direct use of `cast*` nodes. So the better message here and in other places is:

> Use `xod/core/pulse-on-true` instead

brusherru: 👌 
Updated deprecated reasons for old-fashioned `cast` nodes:
- `boolean-to-number` -> `if-else`
- `boolean-to-string` -> `if-else`
- `boolean-to-pulse` -> `pulse-on-true`
- `number-to-boolean` -> `equal`
- `number-to-string` -> `format-number`

PR #1147, opened 2018-04-05
Shallow deduction of pin types

### LINKED ISSUE #1117: As a xoder I want generic pins to be resolved when linking so that I preview outcome and is protected against incompatibility errors
### Rationale

When a xoder makes a link with a generic pin it can happen that all the necessary info to resolve generic types to concrete types is available. In this case, we should color the resolved pins accordingly and use the common rules when the xoder tries to link another pin of the resolved type.

However, if one links two generic pins of not yet resolved types there’s still not enough info, but the link should be allowed. It will render in gray until resolved.

Type resolution involves two passes:

- Top-down
- Bottom-up

The top-down pass gives *strong* type resolution. If a pin has strong type resolution, it behaves like it has that type on its own, i.e. colors itself accordingly, forbids incompatible types linking, marks the incoming/outgoing links with an error if they are not compatible, etc.

After all possible resolutions are done with top-down pass the pin types which are left unresolved are processed in the bottom-up pass. The bottom-up pass gives *weak* type resolution. If a pin resolves its type weakly, it colors itself accordingly but places no restrictions on linking.

When a link having generic type on either end is created or removed, the two-pass resolution runs again.

### Prerequisites

#1116, #1118 

### Acceptance criteria

Consider a patch

```
+------------+      +------------+
|            |      |            |
| const-num  |      | const-str  |
|            |      |            |
+------+-----+      +-----+------+


          +-----+------+
          |            |
          | generic-1  |
          |            |
          +-----+------+

          +-----+------+
          |            |
          | generic-2  |
          |            |
          +-----+------+


+-----+------+      +------+-----+
|            |      |            |
| displ-num  |      | displ-str  |
|            |      |            |
+------------+      +------------+
```

Where `const-*` are constants of a particular type, `displ-*` are some kinds of display of the strict particular type, and `generic-*` are nodes having `t1` as input and output. Then, sequentially:

- [x] I can link `generic-1` to `generic-2`. The link and pins will be kept unresolved (gray).
- [x] I can link `const-num` to `generic-1`. Both generics will resolve to the number type
- [x] I can link `generic-2` to `displ-str`. The generics keep the resolution to the number type
- [x] I can delete the link to `const-num`. The generics will resolve weakly to the string type (and are colored yellow)
- [x] I can delete the link to `displ-str`. The generics become unresolved.
- [x] I can link `generic-2` to `displ-num`. The generics resolve to the number type.
- [ ] I can link `const-str` to `generic-1`. The generics re-resolve to the string and the link to `displ-num` is marked invalid
- [x] I can delete the invalid link and then will be not allowed to create it directly again

### Out of scope

- Any type substitution. The information about type resolution may not propagate to the transpilation process anyhow.
- Showing an error or forbidding a link if the specialization for given types does not exist

evgenykochetkov:
Closes #1117


packages/xod-project/src/patch.js
```diff
@@ -1303,17 +1304,35 @@ export const isAbstractPatch = def(
...
-        R.map(Pin.getPinType),
+      const [genericInputTypes, genericOutputTypes] = R.compose(
+        R.map(R.map(Pin.getPinType)),
+        R.partition(Pin.isInputPin),
         R.filter(Pin.isGenericPin),
         listPins
       )(patch);
 
-      if (R.isEmpty(genericPinTypes)) {
+      // for example, there could be no output-t2 is there is no input-t2
```
nkrkv: *if* there is

packages/xod-project/src/patch.js
```diff
@@ -1303,17 +1304,35 @@ export const isAbstractPatch = def(
...
+
+      if (!R.isEmpty(orphanGenericOutputTypes)) {
+        return Either.Left(
+          new Error(orphanGenericOutputs(orphanGenericOutputTypes))
+        );
+      }
+
+      const allGenericPinTypes = R.compose(
+        R.sort(R.ascend(R.identity)),
+        R.uniq,
```
nkrkv: I’ve found at least three places in this PR with such composition. Maybe, it makes sense to encapsulate the func in a utility:

```js
const sortUniq = R.compose(
  R.sort(R.ascend(R.identity)),
  R.uniq
);
```

packages/xod-project/src/typeDeduction.js
```diff
@@ -0,0 +1,252 @@
...
+      Utils.isGenericType,
+      () =>
+        R.pathOr(Maybe.Nothing(), [nodeId, pinKey], previouslyDeducedPinTypes),
+      Maybe.of
+    )
+);
+
+// :: DeducedPinTypes -> (NodeId -> Patch) -> [Link] -> [(PinKey, DataType)]
+const getPinTypesFromLinks = (
+  getOutwardNodeIdFromLink, // outputs for top-down pass, inputs for bottom-up pass
```
nkrkv: Outward :thinking: a very good word for the case :heart:

packages/xod-project/src/typeDeduction.js
```diff
@@ -0,0 +1,252 @@
...
+                R.propOr([], R.__, pinKeysByGenericType)
+              ),
+              R.always(R.objOf(pinKey, dataType))
+            ),
+            Pin.getPinType
+          )
+        ),
+        Patch.getPinByKey(pinKey)
+      )(abstractPatch)
+    )
```
nkrkv: Deep nesting detected: map -> compose -> map -> compose. It should be simlifiable to just several maps in a sequence.

packages/xod-project/src/typeDeduction.js
```diff
@@ -0,0 +1,252 @@
...
+  )(pinTypesFromLinks);
+};
+
+// filters out intermediate "broken" input(or output) pins,
+// which we needed only during resolving process
+// :: Map NodeId (Map PinKey (Maybe DataType)) -> Map NodeId (Map PinKey (Maybe DataType))
+const omitIntermediateBrokenPins = (getPatchByNodeId, listPins) =>
```
nkrkv: I’m a bit confused about the “broken” term. It’s too wide. Do you mean “Ambiguous”?

packages/xod-client/src/project/nodeLayout.js
```diff
@@ -21,6 +21,7 @@ export const LINK_HOTSPOT_SIZE = {
 export const NODE_CORNER_RADIUS = 5;
 
 export const PIN_RADIUS = 6;
+export const PIN_INNER_RADIUS = PIN_RADIUS - 3;
```
nkrkv: I think it’s better to use 2px border rather than 3 because non-linked pins have a 2px stroke.

packages/xod-project/src/patch.js
evgenykochetkov: I've managed to find only two, both in `validateAbstractPatch`.
Do you think a local utility is ok?
I thinks `sortUniq` is too specific to put in `xod-func-tools`

evgenykochetkov:
All done

packages/xod-project/src/patch.js
nkrkv: Ah, two of them were the same one spread over two commits. My bad. The local utility is OK.

PR #1156, opened 2018-04-16
Add patch trace in the Errors

### LINKED ISSUE #1032: Add xod patch traces for all errors occurred in flatten function
### Rationale
Now we could click "Deploy -> Show Code for Arduino" and get error "Patch @/zz does not have an implementation". If we imagine that we have a lot of nesting of patches into each other, it can be difficult to find this one and replace with something else.
We already have an Errors with such stack trace in dead references error (#1022). So we need just to extend this to the rest of the errors.

### Acceptance criteria
- [ ] I can place nested patches, that contains patch with empty "not-implemented-in-xod" Node at the end. And I'll get an error with stack trace, like "@/main -> @/a -> @/b -> @/c -> @/d: Patch @/d does not have an implementation."

P.S.
It looks like there is one type of errors left, that could be extended with "xod stack trace".
But maybe stack traces will be handy in other errors too... 🤔

brusherru:
It closes #1032

### DIFF
--- a/packages/xod-client/src/messages/components/SnackBarMessage.jsx
+++ b/packages/xod-client/src/messages/components/SnackBarMessage.jsx
@@ -31,34 +31,14 @@ class SnackBarMessage extends React.Component {
     this.props.onCloseMessage(this.props.message.id);
   }
 
-  getMessagePath() {
+  getMessageTrace() {
     const { message } = this.props;
-    const path = message.payload.path;
-    if (!path || path.length === 0) return null;
+    const trace = message.payload.trace;
+    if (!trace || trace.length === 0) return null;
 
     return (
-      <span className="path">
-        {R.compose(
-          R.append(
-            <span key="fin" className="colon">
-              :
-            </span>
-          ),
-          R.init,
-          R.unnest,
-          R.map(pathPart => [
-            <span key={pathPart} className="part">
-              {pathPart}
-            </span>,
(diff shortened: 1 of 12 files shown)


packages/xod-client/src/messages/components/SnackBarMessage.jsx
```diff
@@ -31,6 +31,38 @@ class SnackBarMessage extends React.Component {
     this.props.onCloseMessage(this.props.message.id);
   }
 
+  getMessagePath() {
```
nkrkv: I think you’re messing terminology a bit. The “path” term is used widely in XOD as a notion of owner → libname → patchname. You mean another dimension here. I suggest sticking to a new term here. “Trace” is a good variant you’ve suggested.

packages/xod-func-tools/src/errors.js
```diff
@@ -23,6 +23,37 @@ export const failOnNothing = R.curry((errorType, payload) =>
   foldMaybe(fail(errorType, payload), Either.of)
 );
 
+/**
+ * Updates path of the Error, produced by any of `fail` functions.
+ * Will be used in functions that validates something recursively.
+ */
+export const prependPatchPathToError = def(
```
nkrkv: `prependTraceToError`

brusherru:
@nkrkv renaming is done!

Issue #766, opened 2017-08-30
Simplify Inspector Widgets, split general input handling utils and code specific to Inspector


evgenykochetkov:
### Rationale

Widgets in the Inspector are messy, to add/remove some property it will affect a lot of files, also there is a function `getPinWidgetProps` in the `NodeInspector` that transforms RenderablePin (using a mix of `xod-project` functions for Pin and Ramda functions) into something that will be passed into PinWidgets... That's all should be refactored to keep the code simpler and clearer.

### Affected modules and packages

`xod-client`

evgenykochetkov:
Also see #1169

PR #1164, opened 2018-04-19
Add universal bound literals

### LINKED ISSUE #1158: Switch to universal bound value literals
As a prerequisite to binding values on generic type pins, we need a unified system which unambiguously maps a string to a typed data value. So instead of relying on JSON type system, let's store all bound values as strings with an enforced format.

Type | Now | Then
---- | ---- | ---
Boolean | true | "True"
Number | 42.1 | "42.1"
String | "Hello" | "\"Hello\""
Pulse | "ON_BOOT" | "On Boot"

Currently, the sole purpose of keeping bound values is passing them as is to C++ transpiler. No other logic depends on concrete values. Since it is crucial to keep backward compatibility we can adjust:

```js
Handlebars.registerHelper('cppValue', (type, value) => {...})
```

so that it can deal with the old and new form at the same time. The `type` is known, so, for example, it can interpret both `"True"` and `true` as C++ `true` if `type` is `PIN_TYPE.BOOLEAN`.

To aid a gradual switch to literals and allow binding to generics, the widgets should be changed so that they can *get* values in both the old and new form and *set* values on a commit in the new literal form only. Widgets for all four basic types are affected.

Furthermore, the string widget should be smart enough to automatically perform unquote + enquote right before commit. So a xoder who forgets to enquote manually will get the right literal automatically.

brusherru:
It closes #1158.
Also fixed some minor bug, that dispatches `NODE_PROPERTY_UPDATE` twice.

### DIFF
--- a/packages/xod-arduino/src/templates.js
+++ b/packages/xod-arduino/src/templates.js
@@ -1,7 +1,8 @@
 import * as R from 'ramda';
 import Handlebars from 'handlebars';
 
-import { PIN_TYPE } from 'xod-project';
+import { unquote } from 'xod-func-tools';
+import { ensureLiteral, PIN_TYPE } from 'xod-project';
 
 import { def } from './types';
 
@@ -60,7 +61,11 @@ const cppType = def(
 // Formats a plain JS string into C++ string object
 const cppStringLiteral = def(
   'cppStringLiteral :: String -> String',
-  R.ifElse(R.isEmpty, R.always('XString()'), str => `XStringCString("${str}")`)
+  R.ifElse(
+    R.isEmpty,
+    R.always('XString()'),
+    str => `XStringCString("${R.replace(/"/g, '\\"', str)}")`
+  )
 );
 
 // =============================================================================
@@ -88,7 +93,7 @@ Handlebars.registerHelper('json', JSON.stringify);
 Handlebars.registerHelper(
   'decltype',
   (type, value) =>
-    type === PIN_TYPE.STRING && value !== ''
+    type === PIN_TYPE.STRING && unquote(value) !== ''
       ? 'static XStringCString'
       : `constexpr ${cppType(type)}`
 );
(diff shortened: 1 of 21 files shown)


packages/xod-client/src/editor/components/inspectorWidgets/Widget.jsx
```diff
@@ -79,15 +86,34 @@ export default function composeWidget(Component, widgetProps) {
...
-      const parsedValue = this.parseValue(this.state.value);
-      if (parsedValue !== this.parseValue(this.props.value)) {
-        this.props.onPropUpdate(
-          this.props.entityId,
-          this.props.kind,
-          this.props.keyName,
-          parsedValue
-        );
-      }
+      // Prevent of commiting widgets without changes provided by User
```
evgenykochetkov: Another thing to solve in #766 later on 😉

packages/xod-project/src/constants.js
```diff
@@ -22,11 +22,17 @@ export const PIN_TYPE = {
   T3: 't3',
 };
 
+export const INPUT_PULSE_PIN_BINDING_OPTIONS = {
+  Never: 'Never',
```
evgenykochetkov: What's wrong with all-caps keys?
IMHO, `INPUT_PULSE_PIN_BINDING_OPTIONS['On Boot']` would look a bit awkward

brusherru: Yep. But in this case, we don't need to have another map that will point from 'On Boot' literal to 'ON_BOOT' key in this map. Or not? 🤔

packages/xod-arduino/src/templates.js
```diff
@@ -114,15 +115,26 @@ Handlebars.registerHelper('global', function global(options) {
...
+        R.equals('True'),
+        R.always('true'),
+        R.always('false')
+      ),
+      [PIN_TYPE.NUMBER]: R.identity,
+      [PIN_TYPE.STRING]: R.compose(
+        cppStringLiteral,
+        R.replace(/"/g, '\\"'),
+        unquote
+      ),
```
nkrkv: The function exploded a bit. `R.replace(/"/g, '\\"')` should semantically belong to `cppStringLiteral`.

The repetitive code for `BOOLEAN`, perhaps, better expressed with a simple ternary.

brusherru:
@nkrkv all fixed

Issue #1122, opened 2018-03-14
Toughen linking rules for generic pins


nkrkv:
### Rationale

The type resolution process must *not* depend on the order of link creation. To make it possible we have to sacrifice implicit casts if a link goes to a generic type input. That is if a pin has strong type resolution only an output of exactly the same type is allowed to be linked to it.

Consider an example:

```
+------------+      +------------+
|            |      |            |
| const-bool |      | const-num  |
|            |      |            |
+-----+------+      +-----+------+


          +--+--+--+---+
          |            |
          | if-else    |
          |            |
          +--+---------+
             |
             |
          +--+---------+
          |            |
          |    LCD     |
          |            |
          +------------+
```

Where the `if-else` node has two inputs and an output of type t1. Depending on the order of linking to the `const-*` we can get it resolved to the boolean or number. As a consequence, we’ll see either 0/42 on the display or true/false. That can be fixed with exact type check enforcement.

### Prerequisites

#1117 

### Acceptance criteria

For the example above:

- [ ] I can link `const-bool` to `if-else` and the `if-else` will resolve t1 to boolean
- [ ] I will not be allowed to link `const-num` to another input of `if-else` although the implicit cast exists. An error is shown in this case¹

¹
```markdown
**Incompatible types**
Expected link of `boolean` type, got `number`. Generic types must exactly match
even if an implicit cast exists. Solutions possible:
* convert the `number` to `boolean` with an additional node
* use `if-else(...)` specialization explicitly
* unresolve the `t1` type of `if-else` by deleting its incoming links
```

nkrkv:
@evgenykochetkov Is it done? I guess yes, not sure though if we can safely close the issue.

PR #1167, opened 2018-04-23
Refactor abstract nodes resolution

evgenykochetkov:
- rebind compatible values when changing node type
- replace "flatten and then resolve" algorithm with a recursive one
   - solves the problem with generic variadic patches composed from other generic patches
   - allows more precise error messages(prerequisite for #1122)
- get rid of `PatchSignature` stuff

### DIFF
--- a/packages/xod-project/src/messages.js
+++ b/packages/xod-project/src/messages.js
@@ -42,8 +42,16 @@ export default {
     )}`,
     trace,
   }),
-  CANT_FIND_SPECIALIZATIONS_FOR_ABSTRACT_PATCH: ({ patchPath }) => ({
-    title: `Can't find specializations for abstract patch ${patchPath}`,
+  CANT_FIND_SPECIALIZATIONS_FOR_ABSTRACT_PATCH: ({
+    patchPath,
+    expectedSpecializationName,
+    trace,
+  }) => ({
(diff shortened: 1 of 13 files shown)


evgenykochetkov:
@nkrkv @brusherru fixed

packages/xod-project/src/messages.js
```diff
@@ -42,8 +42,14 @@ export default {
...
-  CANT_FIND_SPECIALIZATIONS_FOR_ABSTRACT_PATCH: ({ patchPath }) => ({
-    title: `Can't find specializations for abstract patch ${patchPath}`,
+  CANT_FIND_SPECIALIZATIONS_FOR_ABSTRACT_PATCH: ({
+    patchPath,
+    expectedSpecializationName,
+    trace,
+  }) => ({
+    title: 'Specialization patch not found',
+    note: `Cannot find specialization ${expectedSpecializationName} for abstract ${patchPath}. Try creating the missing patch in your project or install a library which provides such one.`,
+    trace,
```
nkrkv: Sorry to be unclear here: “Try creating...” should be put in the `solution` field, not `note`

evgenykochetkov: Oh. Done 👍🏻

PR #1169, opened 2018-04-23
Fix quotes in the label

brusherru:
There is no issue.
The bug was introduced by my PR #1164


brusherru:
@evgenykochetkov you mean some flag that will turn off converting the value into literal? 🤔

But...
The former solution to partially reuse the settings of StringWidget, but replace the "component" property — looks like a crutch to avoid code duplication. Now the code is not just duplicated, but also contains differences.
Alternatively, we can encapsulate some widget properties (parts of config) in a separate map by data type. But I think that code will not be reduced and will not become more transparent.

What do you think?

brusherru:
@nkrkv check it out, please!

packages/xod-client/src/editor/components/inspectorWidgets/index.js
```diff
@@ -56,10 +56,35 @@ const WIDGET_MAPPING = {
...
+        ),
+        // Let user type `nan`, `+inf`, `-inf` values with wrong case
+        // and type `+Inf` without plus symbol — correct it automatically
+        R.cond([
+          [R.test(/^nan$/i), R.always('NaN')],
+          [R.test(/^\+?inf$/i), R.always('+Inf')],
+          [R.test(/^-inf$/i), R.always('-Inf')],
+          [R.T, R.identity],
+        ]),
+        ensureLiteral(PIN_TYPE.NUMBER)
```
nkrkv: A simple lowercasing before the `cond` will make things better: no need for regex testing, a `42E+3` will be converted to `42e+3` automatically.

nkrkv:
:bug: 

Literal values `.5` and `5.` should be valid numbers. Currently they are not.

:bug: 

I can enter any trash into a number widget and press `<TAB>`. The trash will be preserved. I expect reset to 0.

brusherru:
@nkrkv fixed!
Also, I left a comment about valid numbers.

packages/xod-project/src/utils.js
```diff
@@ -146,3 +146,38 @@ export const resolveLinkNodeIds = R.curry((nodeIdMap, links) =>
...
+// =============================================================================
+// RegExp parts for `isValidNumberDataValue`
+// =============================================================================
+
+// Optional sign
+const optSignRegExp = '(-|\\+)?';
+// Digits or digits+dot or dot+digits or digits+dot+digits
+const digitsWithDotRegExp = '(\\d+|\\d+\\.|\\.\\d+|\\d+\\.\\d+)';
+// Optional scientific notation with exponential part
+const scientificPartRegExp = '((e-|e\\+|e|\\d+)\\d+)?';
```
nkrkv: Can’t get with an idea how it could hurt, but what `\\d+` in the inner parens does? Scientific notation always starts with `e`, isn’t it? And you have `optSignRegExp`, so shouldn’t it be `(e${optSignRegExp}\\d+)?}`?

packages/xod-project/src/utils.js
```diff
@@ -146,3 +146,38 @@ export const resolveLinkNodeIds = R.curry((nodeIdMap, links) =>
...
+// Composited number regexp
+const numberRegExp = `^${optSignRegExp}${digitsWithDotRegExp}${scientificPartRegExp}$`;
+// NaN
+const nanRegExp = '^NaN$';
+// Inf, +Inf, -Inf
+const infRegExp = '^(-|\\+)?Inf$';
```
nkrkv: Can be more explicit: `^${optSignRegExp}Inf$`

PR #1171, opened 2018-04-26
Migrate bound values into bound literals

brusherru:
There is no separate issue, but it could be a part (or replacement) of #1164

### DIFF
--- a/packages/xod-func-tools/src/monads.js
+++ b/packages/xod-func-tools/src/monads.js
@@ -103,12 +103,10 @@ export const explodeMaybe = def(
   }
 );
 
-export const catMaybies = def(
-  'catMaybies :: [Maybe a] -> [a]',
-  R.compose(
-    R.map(explodeMaybe('Expected Just, but got Nothing.')),
-    R.filter(Maybe.isJust)
-  )
+// :: Iterable Maybe a -> Iterable a
+export const catMaybies = R.compose(
+  R.map(explodeMaybe('Expected Just, but got Nothing.')),
+  R.filter(Maybe.isJust)
 );
 
 /**
@@ -189,4 +187,9 @@ export const leftIf = def(
   'leftIf :: (a -> c) -> (a -> b) -> a -> Either b c',
   (condition, leftFn, val) =>
     condition(val) ? Either.of(val) : Either.Left(leftFn(val))
+);
+
+export const maybeProp = def(
+  'maybeProp :: String -> StrMap a -> Maybe a',
+  R.compose(Maybe, R.prop)
 );
--- a/packages/xod-project/src/extractBoundInputsToConstNodes.js
+++ b/packages/xod-project/src/extractBoundInputsToConstNodes.js
@@ -1,6 +1,6 @@
 import * as R from 'ramda';
 
-import { explodeEither } from 'xod-func-tools';
+import { explodeEither, maybeProp, catMaybies } from 'xod-func-tools';
 
 import * as Pin from './pin';
 import * as Node from './node';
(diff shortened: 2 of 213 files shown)


packages/xod-project/src/extractBoundInputsToConstNodes.js
```diff
@@ -113,12 +113,15 @@ const getMapOfExtractablePinPaths = def(
...
-            : CONST_NODETYPES[pinType]
-      )
+      R.map(([pinType, pinValue]) => {
+        if (pinType === PIN_TYPE.PULSE) {
+          return R.compose(
+            foldMaybe(null, R.prop(R.__, PULSE_CONST_NODETYPES)),
+            ensureLiteral
+          )(pinType, pinValue);
+        }
+        return CONST_NODETYPES[pinType];
```
nkrkv: Hard to understand.

FP vs plain old JS turbulence detected. Looking at the two top functions of the composition I see you’re trying to (a) convert pinType + pinValue to paths which can fail, then (b) reject failures. Nulls/maybes mixed which leads to hard readability. What if:

```js
R.compose(
  R.map(catMaybies)
  R.map(stuffToRejectNothings)
  R.map(R.map((...) => pinType === PIN_TYPE.PULSE
    ? maybeProp(PULSE_CONST_NODETYPES, ensureLiteral(...))
    : maybeProp(CONST_NODETYPES, pinType)))
```

packages/xod-project/src/flatten.js
```diff
@@ -732,7 +727,7 @@ const rekeyBoundValues = R.curry((boundValues, node) => {
     })(boundValues);
   }
 
-  return R.prop('boundValues', node);
+  return R.prop('boundLiterals', node);
```
nkrkv: Why not `getAllBoundValues`?

packages/xod-project/src/flatten.js
```diff
@@ -783,12 +778,19 @@ export const extractPatches = R.curry(
...
+                'boundLiterals',
+                rekeyBoundValues(boundValues, node),
+                node
+              )
           ),
           R.converge(extractPatches(project, leafPaths), [
             R.compose(getPrefixedId(prefix), Node.getNodeId),
-            R.prop('boundValues'),
-            R.compose(getPatchByPath(project), Node.getNodeType),
+            R.prop('boundLiterals'),
```
nkrkv: Use API?

packages/xod-project/src/legacy.js
```diff
@@ -33,12 +40,117 @@ export const ensureLiteral = def(
...
+  )(node);
+});
+
+// :: [Pin] -> Node -> Node
+const migrateNodeBoundValues = R.curry((node, pins) => {
+  const pinKeys = R.map(Pin.getPinKey, pins);
+  return R.compose(
+    // and then convert all other bound values
+    R.reduce(convertBoundValueToBoundLiteral, R.__, pins),
+    // omit bound values to unexisting pins
```
nkrkv: nonexistent

packages/xod-project/src/legacy.js
```diff
@@ -33,12 +40,117 @@ export const ensureLiteral = def(
...
+    Patch.upsertNodes(R.__, patch),
+    R.map(node =>
+      R.compose(
+        // Remove `boundValues` completely if it's empty
+        R.when(
+          R.pipe(R.prop(LEGACY_BOUND_VALUES_PROP), R.isEmpty),
+          R.dissoc(LEGACY_BOUND_VALUES_PROP)
+        ),
+        R.ifElse(
+          Maybe.isNothing,
```
nkrkv: `R.ifElse(Maybe.isNothing` is a fold of a Maybe, isn’t it?

packages/xod-func-tools/src/monads.js
```diff
@@ -190,3 +188,8 @@ export const leftIf = def(
   (condition, leftFn, val) =>
     condition(val) ? Either.of(val) : Either.Left(leftFn(val))
 );
+
+export const maybeProp = def(
```
evgenykochetkov: 👍🏻

PR #1173, opened 2018-04-26
Adjust documentation index page look-n-feel and structure

nkrkv:
An attempt to make the docs index more comprehensive and more structured:

xod documentation xod


docs/README.md
```diff
@@ -2,89 +2,140 @@
 title: XOD Documentation
 ---
 
+<style>
+/* Force linked headers to be black, not blue */
```
brusherru: If it's a general rule(s) — maybe move it into `styles.css`?

nkrkv: No, it’s not a general rule. It can break some complex layouts done with Semantic UI. I mean, it could be in `style.css`, but a thorough research is required. I would like to postpone it until the moment we’d like to make some other headers black again.

brusherru: 👌

PR #1175, opened 2018-04-26
Bind values to generic pins

### LINKED ISSUE #1159: As a xoder I want to bind values to generic pins so that I don’t have to use many constant nodes and switch specializations
### Rationale

Currently, if a pin has a generic type I can’t bind a value on it. Inspector shows “generic type” and the input is disabled. It forces me to use `constant-xxx` or switch specializations even in trivial cases like `if-else` routing. Such behavior makes XOD UX worse than it was before generics introductions and breaks the tutorial structure.

Trivial workflows should be as smooth as possible. To make it possible we can allow binding on generic type pins. The type could be deduced from the bound literal value. And in case of conflicts, the pins involved in the error and the node itself could be colored red with error text available in a tooltip.

### Prerequisites

#1158 

### Acceptance criteria

- [x] I can place an `if-else` node and see Inspector allows binding on `T` and `F` pins
- [x] I can bind `42` to `T` and `F` with `R` will resolve to Number
- [x] I can re-bind `"Hello"` to `T` and `F` with `R` will resolve to String
- [x] I can bind `True` to `F` and `T` with `F` and the node itself will cause an error, the same as when link types are in conflict
- [x] I can re-bind `Ouch` to `F` and `F` will cause an error¹

¹
```markdown
**Bad literal value**
@/main -> @/foo -> if-else: Value `Ouch` bound to pin `F` is invalid.
If you meant a string, surround it with double quotes: `"Ouch"`.
```

evgenykochetkov:
Closes #1159


packages/xod-project/src/project.js
```diff
@@ -548,6 +549,75 @@ export const validateLinkPins = def(
   }
 );
 
+export const getInvalidBoundNodePins = def(
+  'getInvalidBoundNodePins :: Project -> Patch -> Node -> StrMap (Either Error a)',
```
nkrkv: Can’t easily understand what does the function return. A comment would help a lot (we document exported funcs, yep? :smiley: ).

In particular, what’s hidden behind `a`?

packages/xod-project/src/utils.js
```diff
@@ -76,6 +79,36 @@ export const defaultValueOfType = def(
...
+
+    if (isAmong(['True', 'False'], literal))
+      return Either.of(CONST.PIN_TYPE.BOOLEAN);
+
+    if (isAmong(R.values(CONST.INPUT_PULSE_PIN_BINDING_OPTIONS), literal))
+      return Either.of(CONST.PIN_TYPE.PULSE);
+
+    if (R.test(/^".*"$/gi, literal)) return Either.of(CONST.PIN_TYPE.STRING);
+
+    if (!isNaN(parseFloat(literal))) return Either.of(CONST.PIN_TYPE.NUMBER);
```
nkrkv: `NaN` which is NaN is a valid literal value. Could you reuse what @brusherru did for validating number literals?

packages/xod-project/src/utils.js
```diff
@@ -76,6 +79,36 @@ export const defaultValueOfType = def(
...
+
+    return fail('BAD_LITERAL_VALUE', { value: literal });
+  }
+);
+
+export const getTypeFromLiteralUnsafe = def(
+  'getTypeFromLiteralUnsafe :: DataValue -> DataType',
+  R.pipe(getTypeFromLiteral, explodeEither)
```
nkrkv: Can’t understand how can it work?  `explodeEither` expects a potential error message, but it is not passed anywhere. Am I miss something?

evgenykochetkov: Nope, it does not:
https://github.com/xodio/xod/blob/d074a15245503c2926975250deaf6843728018d8/packages/xod-func-tools/src/monads.js#L116-L121

`explodeMaybe` does.

nkrkv: Oh, I forgot an `Either` contains error out of the box. OK, don’t bother.

packages/xod-project/src/utils.js
evgenykochetkov: Fixed!

packages/xod-project/src/project.js
evgenykochetkov: Done

PR #1179, opened 2018-04-27
Bump xcode version in CircleCI config

evgenykochetkov:


evgenykochetkov:
Well, it created way more problems than it solved :(

PR #1192, opened 2018-05-04
Add new type `Byte`

### LINKED ISSUE #1185: Add Byte data type
### Rationale

The byte is a fundamental type when doing data exchange which we have to emulate with `Number`. Number semantics is different: it can be fractional, it eats more memory, it is not limited in range, and so on.

Let’s make a new basic type `Byte` which represents a datum made of eight bits.

Mainly, the values for bytes are set with hexadecimal values. In most programming languages they are usually defined with `0x` prefix, however, in most datasheets, they are represented with an `h` suffix. For the sake of compactness let’s use the `h` suffix, but allow `0x` in the input and make it automatically convert. Sometimes decimal and binary representations are convenient. They could use suffixes too: `d` and `b`. To achieve backward compatibility we should allow values without a suffix: the existing bound values will be picked up correctly, and suffixes in new values will be enforced by the inspector widget. To put in summary, the valid literals for a byte value are:

- `7h`
- `07h`
- `77h`
- `7d`
- `77d`
- `1000b`
- `7`
- `0h`
- `0d`
- `0b`

To assist a user in avoiding bad values for a byte let’s normalize user input like so:

Input | Normalized literal value | Why
----- | ---------- | ---
2Ah  | 2Ah
2ah  | 2Ah
ah    | 0Ah | let’s always use 2 digits for hex
0xa  | 0Ah | used in other langs
aaah | FFh | overflow
wazzaa | 00h | invalid
-2 | 00h | underflow
0  | 0d
15 | 15d | auto-add hex suffix
15d | 15d | explicit decimal
10001000b | 10001000b
0b10001000 | 10001000b | used in other langs
1000b | 00001000b | let’s always use 8 digits for bin

The underlying C++ type for a `Byte` is `uint8_t`. Any XOD byte literal renders to C++ in the `0x7F` form.

A byte can be implicitly cast to `String` (shows up as `0Fh`) and not any other type. To convert a byte to the number, for example, a user should use conversion nodes from `xod/bits` which we’ll implement later.

### Acceptance criteria

- [x] I can add `input-byte`, `output-byte`, and `constant-byte` nodes to my patch
- [x] I can bind value `2Fh` to the `constant-byte`
- [x] I can provide any input string for the bound value and it will be normalized like shown in the table above
- [x] When I transpile, the `2Fh` value is rendered as `uint8_t xxxx = 0x2F` in C++
- [x] I can link the `constant-byte` to `L1` of a `text-lcd` and it will display `2Fh`

### How to implement

1. Add new PIN_TYPE and a default value (`xod-project`)
2. Add new `constant-byte` in `xod/core` library
3. Add new type in the `WIDGET_TYPE` map (`xod-client/editor/constants.js`)
4. Add config for a new widget type and normalization function (`xod-client/editor/components/inspectorWidgets/index.js`)
5. Update `getTypeFromLiteral` (`xod-project/utils.js`)
6. Implement a function that converts any `Byte` literal into C++ byte. Use it in transpilation (`xod-arduino`)

brusherru:
It closes #1185.


brusherru:
>> nth-input(byte) is probably not needed

Removed it. But did not deleted other specializations, cause these nodes already released and probably should be marked as deprecated.

PR #1194, opened 2018-05-08
Fix crash when constructing message about wrong variadic pin types

evgenykochetkov:
Fixes a bug that was [reported on our forum](https://forum.xod.io/t/xod-0-20-1-issue/751)

Bug could be reproduced by placing `variadic-1` node in a patch like this one:
[troublesome-variadic.xodball.zip](https://github.com/xodio/xod/files/1983331/troublesome-variadic.xodball.zip)

### DIFF
--- a/packages/xod-project/src/patch.js
+++ b/packages/xod-project/src/patch.js
@@ -1138,14 +1138,14 @@ export const computeVariadicPins = def(
     }
 
     const outputs = listOutputPins(patch);
-    const outputsCount = outputs.length;
+    const outputCount = outputs.length;
 
-    if (outputsCount === 0) {
+    if (outputCount === 0) {
       return fail('VARIADIC_HAS_NO_OUTPUTS', { trace: [patchPath] });
     }
 
     const inputs = listInputPins(patch);
-    const inputsCount = inputs.length;
+    const inputCount = inputs.length;
     const arityStep = R.compose(
       explodeMaybe(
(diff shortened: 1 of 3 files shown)


nkrkv:
Cannot understand how it works. What changed actually? :roll_eyes: 

This bug opens a serious problem IMO: a fail in the formatting of a failure can lead to an absolute crash. Does it make sense to wrap the formatting in a try/catch clause and fall back to the default formatting on a failure? Or does it sound like fixing 1 out of 1000 problems where something could go wrong? In other words: can we place a few guards to totally prevent the white screens?

packages/xod-project/src/patch.js
```diff
@@ -1137,35 +1137,35 @@ export const computeVariadicPins = def(
       return fail('TOO_MANY_VARIADIC_MARKERS', { trace: [patchPath] });
     }
 
-    const outputs = listOutputPins(patch);
-    const outputsCount = outputs.length;
+    const outputs = R.compose(Pin.normalizePinLabels, listOutputPins)(patch);
```
brusherru: Why do we need to call `normalizePinLabels` here?
It looks like this function needs only output pins count. Am I right?

nkrkv: They are necessary below for computing `pinLabelsOfNonEqualPinTypes`

evgenykochetkov: Yep. But that creates another problem and is out of scope of this fix, so I reverted it back.

PR #1197, opened 2018-05-08
Add support of Byte type in tabular testing

### LINKED ISSUE #1187: Support byte type in tabular tests
### Rationale

We have the byte type in XOD, so it should be supported in tabtests. The literals for tabtests are same: the values which end with `h`, `d`, or `b`.

### Prerequisites

#1185 

### Acceptance criteria

- [ ] I can use byte literals in table tests like I use numbers

brusherru:
It closes #1187

### DIFF
--- a/packages/xod-tabtest/src/TabData.re
+++ b/packages/xod-tabtest/src/TabData.re
@@ -9,13 +9,27 @@ module Value = {
     | Number(float)
     | Boolean(bool)
     | String(string)
+    | Byte(int)
     | Invalid(string);
   let numberRegex = [%re {|/^[+-]?(?=.)*\d*(?:\.\d+)?$/|}];
   let stringRegex = [%re {|/^".*"$/|}];
+  let byteRegex = [%re {|/^[0-9a-f]{2}h|[0,1]{8}b|\d{1,3}d$/i|}];
   let unquote = str =>
     str
     |> Js.String.replaceByRe([%re "/^\"/"], "")
     |> Js.String.replaceByRe([%re "/\"$/"], "");
+  let init = (str: string) : string =>
+    String.sub(str, 0, String.length(str) - 1);
+  let byteStringToInt = str =>
+    switch (str) {
+    | bin when Re.test(bin, [%re {|/b$/|}]) =>
+      "0b" ++ init(bin) |. int_of_string |. (x => Byte(x))
+    | hex when Re.test(hex, [%re {|/h$/|}]) =>
+      "0x" ++ init(hex) |. int_of_string |. (x => Byte(x))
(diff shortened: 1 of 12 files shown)


packages/xod-tabtest/src/TabData.re
```diff
@@ -20,12 +22,16 @@ module Value = {
     switch (str) {
     | "" => Empty
     | "true" => Boolean(true)
+    | "True" => Boolean(true)
     | "false" => Boolean(false)
+    | "False" => Boolean(false)
```
nkrkv: There’s a shorter form:

```reason
| "true"
| "True" => Boolean(true)
| "false"
| "False" => Boolean(false)
```

packages/xod-tabtest/src/Tabtest.re
```diff
@@ -170,6 +171,7 @@ module TestCase = {
     | Boolean(true) => "true"
     | Boolean(false) => "false"
     | NaN => "NAN"
+    | Byte(x) => string_of_int(x)
```
nkrkv: I guess this line is not necessary since `Byte(x)` will be correctly handled by the last clause.

packages/xod-tabtest/src/TabData.re
```diff
@@ -9,23 +9,41 @@ module Value = {
...
+    };
   let parse = str =>
     switch (str) {
     | "" => Empty
-    | "true" => Boolean(true)
-    | "false" => Boolean(false)
+    | "true"
+    | "True" => Boolean(true)
+    | "false"
+    | "False" => Boolean(false)
```
nkrkv: If we deal with bytes in the PR, why touch booleans? Yes, there’s now an inconsistency between the XOD IDE and XOD tabtests, but I’m sure it should be addressed separately and there should be only a single variant allowed.

packages/xod-tabtest/src/TabData.re
```diff
@@ -9,23 +9,41 @@ module Value = {
...
   let numberRegex = [%re {|/^[+-]?(?=.)*\d*(?:\.\d+)?$/|}];
   let stringRegex = [%re {|/^".*"$/|}];
+  let byteRegex = [%re {|/^[0-9a-f]{2}h|[0,1]{8}b|\d{1,3}d$/i|}];
   let unquote = str =>
     str
     |> Js.String.replaceByRe([%re "/^\"/"], "")
     |> Js.String.replaceByRe([%re "/\"$/"], "");
+  let _init = (str: string) : string =>
+    String.sub(str, 0, String.length(str) - 1);
+  let _byteStringToInt = str =>
```
nkrkv: No reason to prepend with the underscore. `Value` is already a utility module.

PR #1211, opened 2018-05-15
Tweak performance of type deduction

brusherru:
It tweaks IDE performance, that was reduced in the v0.20.0 by adding generic types and type deduction.


packages/xod-project/src/patch.js
```diff
@@ -830,7 +830,9 @@ const sendDeferNodesToBottom = def(
  */
 export const getTopology = def(
   'getTopology :: Patch -> Either Error [NodeId]',
-  patch => R.compose(R.map(sendDeferNodesToBottom(patch)), toposortGraph)(patch)
+  R.memoizeWith(JSON.stringify, patch =>
```
nkrkv: If I understand correctly, it will blow RAM usage eventually. See https://github.com/ramda/ramda/blob/v0.25.0/source/memoizeWith.js the `cache` dictionary is filled up as we get slight patch variations and nothing cleans the cache then.

`JSON.stringify` is not a speed champion either.

Does the memoization in this place bring notable performance improvements?

packages/xod-project/src/project.js
```diff
@@ -673,9 +673,16 @@ export const validatePatchContents = def(
...
+      R.any(Pin.isGenericPin),
+      Patch.listPins
+    );
+
     return checkNodeTypes(patch)
       .chain(checkedPatch => {
-        const deducedPinTypes = deducePinTypes(checkedPatch, project);
+        const deducedPinTypes = checkIsPatchGeneric(checkedPatch)
+          ? deducePinTypes(checkedPatch, project)
+          : {};
```
nkrkv: I guess it would be better to incorporate the check *into* `deducePinTypes`. Otherwise, we’ll have code duplication or optimization misses.

packages/xod-project/src/typeDeduction.js
```diff
@@ -192,26 +193,58 @@ const omitIntermediateAmbiguousPins = (
     )
   );
 
+// :: [a] -> Map a *
+const toMap = R.pipe(R.map(a => [a, true]), R.fromPairs);
```
nkrkv: Isn’t it a [`Set`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set)?

evgenykochetkov: I'd add a comment explaining why this is needed.
Something along the lines of "Because R.has is faster than R.find"

packages/xod-project/src/patch.js
evgenykochetkov: Using `WeakMap` to store cache could be cool.
There are probably a dozen of libraries on npm that do this🤔

PR #1213, opened 2018-05-16
Tweak performance of type deduction

brusherru:
It tweaks IDE performance, that was reduced in the v0.20.0 by adding generic types and type deduction.

The same as https://github.com/xodio/xod/pull/1211
But based on another branch (0.20.x)

### DIFF
--- a/packages/xod-project/src/typeDeduction.js
+++ b/packages/xod-project/src/typeDeduction.js
@@ -10,6 +10,8 @@ import {
   fail,
   failOnNothing,
   prependTraceToError,
+  toSet,
+  inSet,
 } from 'xod-func-tools';
 
 import * as Link from './link';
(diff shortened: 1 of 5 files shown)


packages/xod-project/src/typeDeduction.js
```diff
@@ -192,26 +193,62 @@ const omitIntermediateAmbiguousPins = (
     )
   );
 
+// We use `toSet` and `inSet` in `deducePinTypes`
+// cause it is faster than `isAmong` function.
+// :: [a] -> Set [a]
+const toSet = a => new Set(a);
+// :: a -> Set [a] -> Boolean
```
nkrkv: I think the signature `Set [a]` is wrong. Should it be `Set a`?

And I think it’s a better idea to move the usage note comment to the place of actual usage. Imagine you’d promote the two funcs to `xod-func-tools`.

PR #1216, opened 2018-05-17
Custom types

### LINKED ISSUE #1189: As a xoder I want to define own data types so that I can operate on higher level objects
### Rationale

The vast majority of programming languages have mechanisms to define custom data types besides built-in ones. It allows building new layers of powerful abstractions to empower programs and libraries. We should add such mechanism in XOD to later be able to express IoT programs, for example.

The way to add a new data type is to make a patch with a `xod/patch-nodes/output-self` terminal on it. If a xoder places such terminal on a `color` patch, that defines a new type with name `color` in the program. The underlying C++ data type for `color` must be defined by the xoder. Along with `State` and `evaluate` he defines a type with the name `Type` which will be used by XOD to carry values of `foo`:

```cpp
/* A potential implementation of the `color` constructor  */

struct State {
    // this is a regular state related to the patch (i.e. constructor in this case)
};

struct Type {
    uint8_t r;
    uint8_t g;
    uint8_t b;
};

void evaluate(Context ctx) {
    uint8_t r = getValue<input_R>(ctx) * 255.f;
    uint8_t g = getValue<input_G>(ctx) * 255.f;
    uint8_t b = getValue<input_B>(ctx) * 255.f;
    Type obj = { r, g, b };
    emitValue<output_OUT>(ctx, obj);
}
```

Using the `output-self` node on a patch automatically generates terminal nodes of the corresponding type: `input-color` and `output-color`.

The new type don’t cast implicitly to any other type.

Consumer C++ nodes who choose to use a custom type for some terminals, refer to the required type using fully-qualified identifiers:

```cpp
void evaluate(Context ctx) {
    bob__fun_lib__color::Type myColor;
    // or
    auto myColor2 = getValue<input_IN>(ctx);
    // or
    using Color = bob__fun_lib__color::Type;
    Color color3;
    Color color4;
}
```

When a node uses a custom type for its input, the field for that pin in the inspector is disabled with a “complex type” gag. If a custom type pin is not bound on the moment of transpilation, a new constructor node is created and its output is fed into the unbound pin. So, effictively for the `color` example we’ll get the black color on each unbound color pin.

### Acceptance criteria

- [ ] I can place an `output-self` on my patch `foo`
- [ ] When I do, the `input-foo` and `output-foo` are automatically generated in the project browser
- [ ] The items to rename and delete the generated terminals are disabled
- [ ] I can remove the `output-self` and the generated terminal nodes will be removed automatically
- [ ] I can place two `output-self` on the `foo` and that’s OK
- [ ] I can place an extra output number terminal and that’s still OK
- [ ] I can place `input-foo` and `output-foo` on a patch `bar`
- [ ] I can place `bar` on `qux` and see the corresponding bound-value widgets disabled with the “complex type” text
- [ ] I can place a few more `bar`s and add links between them
- [ ] I’m not allowed to link `bar` and any other type pin

As a practical proof, make an example library to work with colors which utilizes the new mechanism to express color values. There should be nodes:

- [ ] `color` — constructor, takes R/G/B ∈ [0; 1]
- [ ] `to-rgb` — destructor, decomposes a color into R/G/B ∈ [0; 1]
- [ ] `to-bytes` — destructor, provides byte values for each of the R, G, B components
- [ ] `from-hsv` — alternative constructor H ∈ [0; 1); S, V ∈ [0; 1]
- [ ] `to-hsv` — alternative destructor H ∈ [0; 1); S, V ∈ [0; 1]
- [ ] `rotate` — gets a color and a shift, changes the hue by offseting on the shift
- [ ] `to-string` — color formatter returning a string in the web form (`#FD4007`)

### TODO

- Should we save the generated terminals in xodballs?
- What if a xoder already has a node with name `input-foo` and he place the `output-self` on `foo`?

evgenykochetkov:
Closes #1189

Also contains a commit about extracting removing unused patches from `flatten` to a (much simpler) separate step, which affected a lot of fixtures.


packages/xod-arduino/src/utils.js
```diff
@@ -0,0 +1,31 @@
...
+export const kebabToSnake = R.replace(/-/g, '_');
+
+// foo(number,string) -> foo__number__string
+export const sanitizeTypeSpecification = R.compose(
+  R.replace(/\(|,/g, '__'),
+  R.replace(')', '')
+);
+
+export const createPatchNames = def(
+  'createPatchNames :: PatchPath -> { owner :: String, libName :: String, patchName :: String }',
```
nkrkv: Not quite clear what does it do at a first sight. A comment with example and more precise name can help.

I’m not sure whether the function is useful on its own. It returns a quite strange intermediate structure. Will it be better if it will convert to a namespace name as the last step so that the intermediate type goes away?

packages/xod-func-tools/src/utils.js
```diff
@@ -22,3 +22,17 @@ export const enumerate = R.curry((separator, lastSeparator, items) => {
     R.join(lastSeparator)
   )(lastItems);
 });
+
+export const memoizeOnlyLast = f => {
```
nkrkv: :+1: 

Shorter name suggestion: `cache`

packages/xod-project/src/builtInPatches.js
```diff
@@ -44,14 +49,21 @@ const getTerminalPins = R.curry((direction, type) => {
...
   return R.objOf(pinKey, pin);
 });
 
+export const getCustomTypeTerminalPins = R.converge(getTerminalPins, [
+  getTerminalDirection,
+  getTerminalDataType,
+]);
+
 const BUILT_IN_MARKERS = [
+  OUTPUT_SELF_PATH,
```
nkrkv: :thinking: output-self is not a marker. `PINS_FOR_BUILT_IN_MARKERS` and `PINS_OF_PATCH_NODES` will work incorrectly

packages/xod-project/src/extractBoundInputsToConstNodes.js
```diff
@@ -57,6 +56,7 @@ const getNodePinValues = def(
     R.map(
       node =>
         R.compose(
+          R.reject(R.equals(INPUT_PULSE_PIN_BINDING_OPTIONS.NEVER)),
```
nkrkv: What’s so special in `Never`?

packages/xod-project/src/extractBoundInputsToConstNodes.js
```diff
@@ -139,6 +146,7 @@ const getMapOfPathsToPinKeys = def(
         R.compose(
           constPinKey => [constPath, constPinKey],
           Pin.getPinKey,
+          // TODO: add more logic here about 'output-self'
```
nkrkv: What logic?

packages/xod-project/src/project.js
```diff
@@ -178,7 +181,35 @@ export const BUILT_IN_PATCHES = R.compose(
...
+  R.any(R.equals(CONST.OUTPUT_SELF_PATH)),
+  R.map(Node.getNodeType),
+  Patch.listNodes
+);
+
+// :: Project -> Map PatchPath Patch
+const getPatches = memoizeOnlyLast(project => {
+  // TODO: better name.
+  // It must be clear that those are not built-in or auto-generated patches
+  const assignedPatches = R.prop('patches', project);
```
nkrkv: genuinePatches? :)

packages/xod-project/src/project.js
```diff
@@ -206,7 +237,16 @@ export const listPatches = def(
 export const listPatchesWithoutBuiltIns = def(
   'listPatches :: Project -> [Patch]',
   R.compose(
-    R.reject(R.compose(isAmong(BUILT_IN_PATCH_PATHS), Patch.getPatchPath)),
+    R.reject(
+      // TODO: extract to separate util?
+      R.compose(
```
nkrkv: Another question is why it is called `...WithoutBuiltIns` but returns without builtins *and* terminals? Maybe, rename to `listGenuinePatches` and the util `isGenuinePatchPath`?

packages/xod-project/src/project.js
```diff
@@ -178,7 +181,35 @@ export const BUILT_IN_PATCHES = R.compose(
 // :: String -> Boolean
 export const isPathBuiltIn = R.flip(R.contains)(BUILT_IN_PATCH_PATHS);
 
-const getPatches = R.compose(R.merge(BUILT_IN_PATCHES), R.prop('patches'));
+const isConstructorPatch = R.compose(
```
brusherru: It should be in `patch.js`

packages/xod-project/src/project.js
```diff
@@ -178,7 +181,35 @@ export const BUILT_IN_PATCHES = R.compose(
...
+  R.any(R.equals(CONST.OUTPUT_SELF_PATH)),
+  R.map(Node.getNodeType),
+  Patch.listNodes
+);
+
+// :: Project -> Map PatchPath Patch
+const getPatches = memoizeOnlyLast(project => {
+  // TODO: better name.
```
brusherru: I'm bad in naming, but I think it's time to rename it and remove todo.

evgenykochetkov: > I'm bad at naming, but I think it's time to rename it and remove todo.

Yep, that TODO was more of "help me name it better", not "let's leave it like this and rename some time later"

> My proposals: original / genuine / real patches.

`genuine` sounds fine👍🏻

packages/xod-project/src/builtInPatches.js
evgenykochetkov: If it's a terminal, than it's a *very* special one.
- We do not have `self` type
- Nothing can be linked/bound to `output-self`
- It does not have an `input` counterpart

I think it kind of *is* a marker 🤔

nkrkv: Good points. Initially, I thought about a future where we can derive new types from tuples by linking them to an `output-self`. But, yes, we are not even close to it. So, OK, let it be a marker now.

evgenykochetkov:
@nkrkv @brusherru 
Check it out again, please

PR #1228, opened 2018-05-21
Run tabular tests when doing CI

### LINKED ISSUE #1227: Run tabtests in CI
We have tabular tests but they only can be run manually. They will stale if not run regularly.
Running all tabtest for the standard library during a regular CI job is required.

nkrkv:
- Closes #1227 
- Adds support to test a whole XOD project, not a particular patch
- Adds `xodc tabtest --output-dir=/path/to/dir` switch
- Adds `xodc tabtest --no-build` switch

### DIFF
--- a/packages/xod-tabtest/src/Tabtest.re
+++ b/packages/xod-tabtest/src/Tabtest.re
@@ -249,7 +249,7 @@ module TestCase = {
   };
 };
 
-let generateSuite = (project, patchPathToTest) : XResult.t(t) => {
+let generatePatchSuite = (project, patchPathToTest) : XResult.t(t) => {
   let patchUnderTestOpt = Project.getPatchByPath(project, patchPathToTest);
   let tsvOpt = patchUnderTestOpt |. Option.flatMap(Patch.getTabtestContent);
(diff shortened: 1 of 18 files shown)


packages/xod-tabtest/src/Tabtest.re
```diff
@@ -266,9 +266,15 @@ let generateSuite = (project, patchPathToTest) : XResult.t(t) => {
   | (Some(patchUnderTest), Some(tsv)) =>
     let bench = Bench.create(project, patchUnderTest);
     let probes = bench.probes;
-    let benchPatchPath = "@/tabtest-bench";
+    let benchPatchPath = "tabtest-" ++ patchPathToTest;
+    let safeBasename =
+      PatchPath.getBaseName(patchPathToTest)
+      |> Js.String.replace("(", "__")
+      |> Js.String.replace(")", "");
```
evgenykochetkov: What about patches with more than one generic type(`foo(number,string)` etc)?

nkrkv: Oops, commas should also be replaced. Will fix it now.

PR #1240, opened 2018-05-23
Add byte-number conversion nodes

### LINKED ISSUE #1186: Add byte-number conversion utilities to xod/bits
### Rationale

Bytes are only used to store and transfer data. On the other hand, numbers are more applicative. We need a sane way to convert between numbers and bytes. Let’s add a few nodes to `xod/bits`:

- `i8-to-number`
- `u8-to-number`
- `i16-to-number`
- `u16-to-number` (same as `xod/core/word-to-number`)
- `i32-to-number`
- `u32-to-number`
- `number-to-i8`
- `number-to-u8`
- `number-to-i16`
- `number-to-u16`
- `number-to-i32`
- `number-to-u32`

Here, `i` means signed and `u` is unsigned. The number denotes number of bits from the byte-data side. For example, `number-to-i32` have a single number input and four byte outputs. The number is considered to be integer: a fractional part is truncated if present in the input.

To serialize numbers as is, let’s implement a couple of additional nodes:

- `ieee754-to-number` — takes four bytes and returns a single number
- `number-to-ieee754`

### Prerequisites

#1185, #1187 

### Acceptance criteria

- [ ] The nodes listed are implemented in `xod/bits`
- [ ] The `xod/core/word-to-number` is deprecated
- [ ] All new nodes are documented
- [ ] All new nodes have table tests

brusherru:
It will close #1186


workspace/__lib__/xod/bits/i16-to-number/patch.cpp
```diff
@@ -0,0 +1,13 @@
...
+struct State {
+};
+
+{{ GENERATED_CODE }}
+
+void evaluate(Context ctx) {
+    uint8_t b1 = getValue<input_B1>(ctx);
+    uint8_t b0 = getValue<input_B0>(ctx);
+    int16_t bytes = ((b1 << 8) | b0);
+    Number result = bytes;
```
nkrkv: The outer parens are not necessary. I suggest being more explicit on why the `Number result = bytes` statement exists. You perform casting here, so:

```cpp
Number result = (Number)bytes;
// or
auto result = (Number)bytes;
// or
auto result = static_cast<Number>(bytes);
```

The same for other implementations. Just be consistent.

workspace/__lib__/xod/bits/number-to-i32/patch.cpp
```diff
@@ -0,0 +1,17 @@
+
+struct State {
+};
+
+{{ GENERATED_CODE }}
+
+void evaluate(Context ctx) {
+    int32_t val = getValue<input_IN>(ctx);
+    uint8_t b3 = ((val >> 24) & 0xFF);
```
nkrkv: I guess `& 0xFF` is not required here as the highest bits are always 0 after the shift.

packages/xod-client/src/core/styles/components/PatchDocs.scss
```diff
@@ -42,6 +42,7 @@ $PinInfo-padding-bottom: 10px;
       &.number { color: $color-datatype-number }
       &.string { color: $color-datatype-string }
       &.generic { color: $color-datatype-generic }
+      &.byte { color: $color-datatype-byte }
```
nkrkv: Has nothing to do with the PR topic, I think.

brusherru: Yep, but... Okay, I'll fix it in another one :)

brusherru:
@nkrkv @evgenykochetkov, check out `i32-to-number` descriptions, please.

workspace/__lib__/xod/bits/i32-to-number/patch.xodp
```diff
@@ -0,0 +1,59 @@
+{
+  "description": "Converts four bytes into integer-valued number. Result of the conversion will be between -2147483647.00 and 2147483647.00.",
```
nkrkv: An important point about the sign is a bit implicit. Also, are you sure that the lower bound is 214748364**7** and not 214748364**8**? Also, one article is missing.

> Converts 32-bit signed integer presented as four bytes to a number. A result of the conversion will be between -2147483648.00 and 2147483647.00.

packages/xod-arduino/test-cpp/run-tests.dSYM/Contents/Info.plist
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
evgenykochetkov: I think this should be added to `.gitignore`

brusherru:
@nkrkv check it out, please

brusherru:
@nkrkv done!

PR #1251, opened 2018-05-28
Extract math nodes

### LINKED ISSUE #1237: Extract xod/math
### Rationale

Let’s move math beyond the elementary arithmetics out of `xod/core` so that it can be more lightweight.

### Details

A new standard library `xod/math` should appear and the nodes from `xod/core` moved there should be deprecated with a typical “Use `xod/math/foo` instead`. The move gives a chance to apply a few renames and label adjustments.

The plan is summarized below.

Current | New | Comment
-- | -- | --
xod/core/absolute | xod/math/abs | Rename
xod/core/acos | xod/math/acos
xod/core/asin | xod/math/asin
xod/core/atan | xod/math/atan
xod/core/ceil | xod/math/ceil
xod/core/clip | xod/math/clip | Remove output label
~xod/core/constrain~ |  | Just deprecate: use clip instead
xod/core/cos | xod/math/cos
xod/core/cube | xod/math/cube
xod/core/discretize | xod/math/discretize
xod/core/floor | xod/math/floor
xod/core/log-10 | xod/math/log-10
xod/core/log-bx | xod/math/log-bx
xod/core/log-e | xod/math/log-e
xod/core/map-clip-range | xod/math/map-clip | Rename, remove output label
xod/core/map-range | xod/math/map | Rename, remove output label
xod/core/pi | xod/math/pi
xod/core/pow | xod/math/pow
xod/core/round | xod/math/round
xod/core/sin | xod/math/sin
xod/core/sqrt | xod/math/sqrt
xod/core/square | xod/math/square
xod/core/tan | xod/math/tan

When copying be careful while splitting compositional references: many of `@`’s are going to be replaced with `xod/core`.

To provide better migration UX we can add `README.md` to renamed nodes with a section “Also known as” and the old name mentioned there. Having done that, the node will appear in the quick search by its old name.

### Acceptance criteria

- [x] `xod/math` is in IDE with nodes listed
- [x] The original nodes are deprecated
- [x] Descriptions of nodes are adjusted where necessary to math new names/labels
- [x] All `docs/**/*.md` are grepped for references and adjusted to use new paths
- [x] `welcome-to-xod` contains no obsolete references to math nodes
- [x] Link to `xod/math` is added to the Standard Library section of the Reference
- [x] I can quick-search for `map-ra` and the `xod/math/map` will appear as the first result

brusherru:
It closes #1237

### DIFF
--- a/workspace/__lib__/xod/math/map-clip/README.md
+++ b/workspace/__lib__/xod/math/map-clip/README.md
@@ -1,0 +1,1 @@
+Previously known as `xod/core/map-clip-range`.
--- a/workspace/__lib__/xod/math/map/README.md
+++ b/workspace/__lib__/xod/math/map/README.md
@@ -1,0 +1,1 @@
+Previously known as `xod/core/map-range`.
--- a/docs/guide/analog-sensor-node/sharp-irm.step1.xodball
+++ b/docs/guide/analog-sensor-node/sharp-irm.step1.xodball
@@ -131,7 +131,7 @@
         },
         "HkbVixief": {
           "id": "HkbVixief",
-          "type": "xod/core/map-range",
+          "type": "xod/math/map",
           "position": {
             "x": 102,
             "y": 204
--- a/docs/guide/analog-sensor-node/sharp-irm.step2.xodball
+++ b/docs/guide/analog-sensor-node/sharp-irm.step2.xodball
@@ -58,7 +58,7 @@
         },
         "HkWRBRxigf": {
           "id": "HkWRBRxigf",
-          "type": "xod/core/map-range",
+          "type": "xod/math/map",
           "position": {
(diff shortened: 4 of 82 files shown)


workspace/__lib__/xod/math/map-clip/README.md
```diff
@@ -0,0 +1 @@
+Also known as `xod/core/map-clip-range`.
```
nkrkv: Previously known as ...

workspace/__lib__/xod/math/map/README.md
```diff
@@ -0,0 +1 @@
+Also known as `xod/core/map-range`.
```
nkrkv: Previously known as ...

PR #1255, opened 2018-05-29
Make toolchain download and installation process more durable

### LINKED ISSUE #1205: Durable toolchain download and install
The issue appears on the forum from time to time. Manual recovering is cumbersome. The flaw affects newcomers and effectively ruins the first impression.

### Steps to reproduce

1. Make a fresh install of the desktop IDE (no packages or tools installed)
2. Connect a board and try to upload an empty patch
3. Close IDE quickly before it has a chance to complete 100 MB download
4. Open IDE again and try to upload again

### Expected behavior

- The incomplete toolchain download/install is restarted or resumed

### Actual behavior

- Depending on the stage when it was interrupted: an ENOENT error about `boards.txt`, or hanging

### How to fix

The problems are in `doesXxxxExist` in `packages/xod-arduino-deploy/src/packageManager.js`. They use directory existence as a sign the toolchain was successfully installed. However, the directory is created right at the start of big archives download so in case of an interruption XOD IDE cannot differentiate between complete and incomplete toolchain install.

The algorithm should be changed to be more durable:

* Use the full path to a marker file `INSTALLED` when checking whether a toolchain is installed or not
* Use the full path to an archive to check whether the required tool bundle is present on FS or should be downloaded
* Download archives into files with temporary `.part` extension like web-browsers do
* When a download is complete, rename the file to drop the `.part` extension
* When unarchiving is complete, touch the `INSTALLED` marker file in the destination directory

evgenykochetkov:
Closes #1205


evgenykochetkov:
@nkrkv @brusherru check it out again, please.

PR #1264, opened 2018-06-01
Fix transpiling analog ports into another C++ literals

brusherru:
There is no issue, but it could be a problem for some boards...

Transpile analog ports into `A#` instead of `PIN_A#`, cause some boards do not have defined `PIN_A#`. Need a proof? See `pins_arduino.h` for Arduino Due.

### DIFF
--- a/packages/xod-arduino/src/templates.js
+++ b/packages/xod-arduino/src/templates.js
@@ -131,14 +131,10 @@ const cppByteLiteral = def(
 // E.G.
 // 3 -> 3
 // D13 -> 13
-// A3 -> PIN_A3
+// A3 -> A3
 const cppPortLiteral = def(
   'cppPortLiteral :: String -> String',
-  R.cond([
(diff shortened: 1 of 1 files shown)


packages/xod-arduino/src/templates.js
```diff
@@ -131,14 +131,10 @@ const cppByteLiteral = def(
...
-// A3 -> PIN_A3
+// A3 -> A3
 const cppPortLiteral = def(
   'cppPortLiteral :: String -> String',
-  R.cond([
-    [R.test(/^D\d+$/i), R.tail],
-    [R.test(/^A\d+$/i), R.pipe(R.tail, R.concat('PIN_A'))],
-    [R.T, R.identity],
-  ])
+  R.cond([[R.test(/^D\d+$/i), R.tail], [R.T, R.identity]])
```
nkrkv: Now it can be simplified to `R.when(R.test(...), R.tail)`

PR #1270, opened 2018-06-05
Extract I2C-related nodes of standard library to `xod/i2c`

### LINKED ISSUE #1253: Extract xod/i2c
### Rationale

Currently, a basic support for I²C is implemented with `xod/core/i2c-*` nodes. There’re some asymmetrical things like `i2c-request-bytes-6` with no nodes to request fewer bytes, `i2c-send-bytes-2` without more bytes, etc. Furthermore, only hardware I2C0 is supported. Let’s make a simpler and more straightforward `xod/i2c` using custom types to later support alternative I2C interfaces like software I2C.

### Nodes

Common changes

- All new nodes should use the byte type for address pins and data transfer.
- All new nodes should take `I2C` of type `xod/i2c/i2c` as the first input input
- A call to `begin()` should be done once, in the constructor node, not in each `evaluate`

Old Node | New Node | Changes
--- | --- | ---
i2c-begin-transmission | begin-transmission
i2c-end-transmission | end-transmission
i2c-read | read | remove `TOUT`, add `ERR`
i2c-request | request
i2c-request-bytes-6 | | remove
i2c-send-byte | | remove
i2c-send-bytes-2 | | remove
i2c-write | write

### Acceptance criteria

- [x] The `xod/i2c` is there
- [x] `i2c-*` nodes in `xod/core` are deprecated

### Out of scope

- Software I2C support

nkrkv:
Closes #1253

### DIFF
--- a/workspace/__lib__/xod/i2c/request/patch.xodp
+++ b/workspace/__lib__/xod/i2c/request/patch.xodp
@@ -1,0 +1,76 @@
+{
+  "description": "Requests N bytes from a given slave I²C device",
+  "nodes": [
+    {
+      "description": "Slave device I²C address to request from",
+      "id": "BkRw4E3kX",
+      "label": "ADDR",
+      "position": {
+        "x": 68,
+        "y": 0
+      },
+      "type": "xod/patch-nodes/input-byte"
+    },
+    {
+      "description": "I2C interface object",
+      "id": "Hk1dRQh17",
+      "label": "I2C",
+      "position": {
+        "x": 0,
+        "y": 0
+      },
+      "type": "@/input-i2c"
+    },
+    {
+      "description": "Pulses if request failed. E.g., due to an invalid `ADDR`.",
+      "id": "S1eJu0Q2JQ",
+      "label": "ERR",
+      "position": {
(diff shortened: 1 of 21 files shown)


workspace/__lib__/xod/i2c/request/patch.xodp
```diff
@@ -0,0 +1,76 @@
...
+        "__out__": "01h"
+      },
+      "description": "Number of bytes to request",
+      "id": "r1NONEn1m",
+      "label": "N",
+      "position": {
+        "x": 136,
+        "y": 0
+      },
+      "type": "xod/patch-nodes/input-byte"
```
evgenykochetkov: Is there a reason this is an `input-byte` besides https://github.com/arduino/ArduinoCore-avr/blob/b7c607663fecc232e598f2c0acf419ceb0b7078c/libraries/Wire/src/Wire.h#L61-L63 ?
IMHO, "**Number** of bytes to request" should be a number.

nkrkv: Oops, no. Good catch. Should be a number, I will fix it

PR #1274, opened 2018-06-08
Use port and byte pins in common hardware

### LINKED ISSUE #1254: Lift xod/common-hardware to use ports and bytes
### Rationale

We’ve introduced byte and port type, however, `xod/common-hardware` emulate them through numbers. Also, #1241 and #1253 bring new ways to perform GPIO and I²C interfacing. The standard `xod/common-hardware` should use this stuff to be a good example for others and become more effective.

There are programs in the world which depend on the nodes that we are going to change. We should minimize the damage caused by the breaking change. We can preserve terminal IDs, changing the type of terminal only.

### Prerequisites

#1241, #1253 

### Acceptance criteria

In the context of `xod/common-hardware`:

- [ ] Projects using leds and buttons from 0.20.3 load and work as in 0.20.3
- [ ] All nodes are transformed to use new `xod/gpio` and `xod/i2c`
- [ ] All nodes use `Byte` for I2C addresses
- [ ] All nodes use `Port` for ports
- [ ] No node uses deprecated functionality internally
- [ ] All nodes have explicit `UPD`, `ERR`, `DONE`
- [ ] The new pins are documented
- [ ] The `welcome-to-xod` has no alignment and overlap issues related to the appearance of new pins

evgenykochetkov:
Closes #1254

### DIFF
--- a/workspace/__lib__/xod/common-hardware/adxl335-accelerometer/patch.xodp
+++ b/workspace/__lib__/xod/common-hardware/adxl335-accelerometer/patch.xodp
@@ -24,6 +24,17 @@
       }
     },
     {
+      "id": "BkfwhKHl7",
+      "input": {
+        "nodeId": "HkrZrcIt-",
+        "pinKey": "SkuhqCqym"
+      },
+      "output": {
+        "nodeId": "BkdA4cUFW",
+        "pinKey": "__out__"
+      }
+    },
+    {
       "id": "By27I9UYZ",
       "input": {
         "nodeId": "H1EVB58Y-",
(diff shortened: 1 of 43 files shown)


evgenykochetkov:
Also, @brusherru suggested yesterday to mark `hc-sr04-ultrasonic-time` an an utility. WDYT?

nkrkv:
@evgenykochetkov I think we should keep the `common-hardware` public API as is for now. It sounds OK, but let’s think again very soon when we’ll split the library into many `xod-dev/lalala` libs.

PR #1276, opened 2018-06-08
Update screenshots for guide

evgenykochetkov:


brusherru:
`docs/guide/variadics/overview.patch.png` was better 😬 
Cause it had a cursor (which illustrates how to change arity) and monochrome comment.
I understood that is generated by the tool, but nevertheless.

tools/update-docs-screenshots.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
```
knopki:
```
#!/bin/sh
export SHOT="$PWD/tools/screenshot-xodball"
find ./docs -iname "update-screenshots.sh" -execdir ./update-screenshots.sh \;
```
But not all versions of `find` has option execdir.

nkrkv:
@brusherru We have to sacrifice something to achieve automation. The *content* should not depend on the cursor in screenshots.

PR #1277, opened 2018-06-09
Upgrade puppeteer to v1.4.0

evgenykochetkov:


packages/xod-client-browser/test-func/pageObjects/ProjectBrowser.js
```diff
@@ -43,6 +43,9 @@ class ProjectBrowser extends BasePageObject {
   }
 
   async addNodeViaContextMenu(groupName, itemName) {
+    // Hover before Click is workaround over strange error:
+    // https://github.com/GoogleChrome/puppeteer/issues/1769
+    await this.elementHandle.hover();
     const item = await this.scrollToPatchGroupItem(groupName, itemName);
 
     await item.click();
```
nkrkv: Supporting the common workaround, shouldn’t it be preceded with `await item.hover()`?

PR #1284, opened 2018-06-13
Make snack bar messages persistent by default

### LINKED ISSUE #1234: Make error snack bar messages persistent by default
### Rationale

Most error messages are important enough to keep them visible until a xoder explicitly acknowledges he read it (click close). By default now all messages are transient. Changing the behavior for errors and keeping for successes will likely improve UX.

Nevertheless, some errors should be kept transient. For example:

- an error caused by an invalid link attempt should be transient.

Other cases do not come into my mind immediately, but they will become apparent with real-life use.

### How to implement

Update parameter `persistent` of `addMessage` in `packages/xod-client/src/messages/actions.js` to be a part of `opts` so that it can be `true`/`false`/absent. Absent means auto-detect: `true` for errors and `false` in other cases.

Review all `addMessage` calls to ensure they are compatible with the new signature.

Explicitly make linking errors transient.

brusherru:
It closes #1234

### DIFF
--- a/packages/xod-client/src/messages/actions.js
+++ b/packages/xod-client/src/messages/actions.js
@@ -1,21 +1,23 @@
+import * as R from 'ramda';
 import * as ActionType from './actionTypes';
 import { MESSAGE_TYPE } from './constants';
 import { STATUS } from '../utils/constants';
 
 const getTimestamp = () => new Date().getTime();
 
-export const addMessage = (
-  type,
-  messageData,
-  persistent = false,
-  id = null
-) => ({
+const getDefaultPersistencyByType = R.equals(MESSAGE_TYPE.ERROR);
+
+export const addMessage = (type, messageData, id = null) => ({
   type: ActionType.MESSAGE_ADD,
   payload: messageData,
   meta: {
     id,
     type,
-    persistent,
+    persistent: R.propOr(
+      getDefaultPersistencyByType(type),
+      'persistent',
+      messageData
+    ),
     timestamp: getTimestamp(),
     status: STATUS.STARTED,
   },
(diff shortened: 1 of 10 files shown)


packages/xod-client/src/messages/actions.js
```diff
@@ -1,21 +1,17 @@
...
-  id = null
-) => ({
+export const addMessage = (type, messageData, id = null) => ({
   type: ActionType.MESSAGE_ADD,
   payload: messageData,
   meta: {
     id,
     type,
-    persistent,
+    persistent: R.propOr(true, 'persistent', messageData),
```
evgenykochetkov: Could it default to `true` only if `type` is `MESSAGE_TYPE.ERROR`?

brusherru: I thought about it and even did it at first. But when I began to place persistent flags in message objects, I realized that there is confusion. So I decided to do it persistent by default for all types of messages.

brusherru:
@nkrkv @evgenykochetkov just fixed it. Check it out, please.

packages/xod-client/src/messages/actions.js
```diff
@@ -1,21 +1,32 @@
...
 
 const getTimestamp = () => new Date().getTime();
 
-export const addMessage = (
-  type,
-  messageData,
-  persistent = false,
-  id = null
-) => ({
+const getDefaultPersistencyByType = type => {
```
evgenykochetkov: why not just `const getDefaultPersistencyByType = R.equals(MESSAGE_TYPE.ERROR)`?

packages/xod-client/src/messages/actions.js
```diff
@@ -1,21 +1,32 @@
...
+  switch (type) {
+    case MESSAGE_TYPE.ERROR:
+      return true;
+    case MESSAGE_TYPE.CONFIRMATION:
+    case MESSAGE_TYPE.NOTIFICATION:
+    default:
+      return false;
```
nkrkv: The whole function can be simplified to `type === MESSAGE_TYPE.ERROR`.

PR #1285, opened 2018-06-13
Rewrite pragma directive lookup in ReasonML

nkrkv:
Upcoming features depend on pragma parsing, so I’m merging the portion of work I’ve done a long time ago so we can evolve it and benefit from strong-typing, ADTs, etc.


packages/xod-tabtest/package.json
```diff
@@ -14,7 +14,7 @@
   "main": "lib/js/src/Tabtest_Js.bs.js",
   "devDependencies": {
     "@glennsl/bs-jest": "^0.4.2",
-    "bs-platform": "^2.2.3"
```
evgenykochetkov: Nice!

PR #1295, opened 2018-06-19
Update tutorial screenshots

gabbapeople:
- New tutorial GIF's and screenshots.


evgenykochetkov:
Also, I am a bit concerned about the addition of ~20 slightly modified copies of tutorial project packed as xodballs — that could be tedious to update. @brusherru @nkrkv WDYT?

nkrkv:
Agree with @evgenykochetkov about the copies. Suggest keeping the required patch only in each xodball. Or alternatively, have one modified xodball at `./docs/tutorial/`

PR #1302, opened 2018-06-26
w5500 support, plus utility libraries for internet stuff and streaming

### LINKED ISSUE #1287: As a xoder I want to connect my W5500 to a network so that I’m online and ready
### Rationale

W5500 is a central IC in the official and many other Arduino Shields for ethernet connectivity. Let’s make it possible to connect a W5500 to the home network.

### User story

A user places a `w5500-device` node from the new `xod-dev/w5500` library. It takes an input `SPI` of type `xod/uart/spi` which wraps Arduino’s `SPIClass` and a `CS` port.

By default `xod/uart/spi` refers to `SPI` object.

Then he links the output device to a `xod-dev/w5500/connect-lan`. Besides control pulses, the node takes `SSID` and `PASS`, and outputs `IP`.

IP is an instance of `xod/net/ip-address` (stored as four bytes) and can be formatted with `xod/net/format-ip`.

### Acceptance criteria

- [ ] I can use the nodes described to make a program which connects to a home network on boot
- [ ] The IP received with DHCP is shown on an LCD
- [ ] The program works with the official Ethernet Shield 2

evgenykochetkov:
Will close #1287 and #1289

For now some functionality(and a lot of documentation) is still missing, but it's already reviewable :)

`xod-dev/w5500/example-get-your-ip` is a good entry point

### DIFF
--- a/workspace/__lib__/xod-dev/w5500/project.xod
+++ b/workspace/__lib__/xod-dev/w5500/project.xod
@@ -1,0 +1,6 @@
+{
+  "description": "Supprot for ethernet shields that use Wiznet W5500 chipset.",
+  "license": "AGPL-3.0",
+  "name": "w5500",
+  "version": "0.22.0"
+}
--- a/workspace/__lib__/xod/net/ip-address/patch.cpp
+++ b/workspace/__lib__/xod/net/ip-address/patch.cpp
@@ -1,0 +1,17 @@
+struct State {
+};
+
+// because most other IPAdress classes can cast from/to uint32_t
+using Type = uint32_t;
+
+{{ GENERATED_CODE }}
+
+void evaluate(Context ctx) {
+    uint32_t octet1 = getValue<input_IN1>(ctx);
+    uint32_t octet2 = getValue<input_IN2>(ctx);
+    uint32_t octet3 = getValue<input_IN3>(ctx);
+    uint32_t octet4 = getValue<input_IN4>(ctx);
+
+    Type ip = octet4 << 24 | octet3 << 16 | octet2 << 8 | octet1;
+    emitValue<output_OUT>(ctx, ip);
+}
--- a/workspace/__lib__/xod/net/mac-address/patch.cpp
+++ b/workspace/__lib__/xod/net/mac-address/patch.cpp
@@ -1,0 +1,21 @@
+
+struct State {
+    uint8_t mac[6];
+};
+
+using Type = uint8_t*;
+
+{{ GENERATED_CODE }}
+
+void evaluate(Context ctx) {
+    auto state = getState(ctx);
+
+    state->mac[0] = getValue<input_IN1>(ctx);
+    state->mac[1] = getValue<input_IN2>(ctx);
+    state->mac[2] = getValue<input_IN3>(ctx);
+    state->mac[3] = getValue<input_IN4>(ctx);
+    state->mac[4] = getValue<input_IN5>(ctx);
+    state->mac[5] = getValue<input_IN6>(ctx);
+
+    emitValue<output_OUT>(ctx, state->mac);
+}
--- a/workspace/__lib__/xod/stream/accumulate-string/patch.cpp
+++ b/workspace/__lib__/xod/stream/accumulate-string/patch.cpp
@@ -1,0 +1,37 @@
+
+struct State {
+    char* buff;
+    char* cursor;
+    size_t cap;
+    CStringView view;
+};
+
+{{ GENERATED_CODE }}
+
+void evaluate(Context ctx) {
+    auto state = getState(ctx);
+
+    if (isSettingUp()) {
+        // save initial cap to ignore possiple input changes during program execution
+        state->cap = getValue<input_CAP>(ctx);
+        state->buff = new char[state->cap + 1]; // +1 to make room for terminal '\0'
+        state->view = CStringView(state->buff);
+    }
+
+    if (isSettingUp() || isInputDirty<input_RST>(ctx)) {
+        memset(state->buff, '\0', state->cap + 1);
+        state->cursor = state->buff;
+    }
+
+    if (isInputDirty<input_PUSH>(ctx)) {
+        if (state->cursor >= &state->buff[state->cap]) {
+            emitValue<output_FULL>(ctx, 1);
+            return;
+        }
+
+        *state->cursor = getValue<input_CHAR>(ctx);
+        state->cursor++;
+        emitValue<output_STR>(ctx, XString(&state->view));
+        emitValue<output_UPD>(ctx, 1);
+    }
+}
--- a/packages/xod-client-electron/arduino-libraries/Ethernet2/CONTRIBUTIONS.txt
+++ b/packages/xod-client-electron/arduino-libraries/Ethernet2/CONTRIBUTIONS.txt
@@ -1,0 +1,33 @@
+---------------
+DHCP.cpp
+
+// DHCP Library v0.3 - April 25, 2009
+// Author: Jordan Terrell - blog.jordanterrell.com
+
+---------------
+DNS.cpp
+
+// Arduino DNS client for WizNet5100-based Ethernet shield
+// (c) Copyright 2009-2010 MCQN Ltd.
+// Released under Apache License, version 2.0
+
+---------------
+Ethernet2: 
+ modified 12 Aug 2013
+ by Soohwan Kim (suhwan@wiznet.co.kr)
+
+---------------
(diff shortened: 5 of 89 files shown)


workspace/__lib__/xod-dev/w5500/project.xod
```diff
@@ -0,0 +1,3 @@
+{
+  "name": ""
```
nkrkv: Metadata should be populated

workspace/__lib__/xod/stream/accumulate-string/patch.cpp
```diff
@@ -0,0 +1,38 @@
...
+            emitValue<output_FULL>(ctx, 1);
+            return;
+        }
+
+        state->buff[state->cursor] = getValue<input_CHAR>(ctx);
+        state->cursor++;
+        emitValue<output_OUT2>(ctx, 1);
+    }
+
+    state->view = CStringView(state->buff);
```
nkrkv: No need to rebuild view again and again. Better do it once in the constructor of `State`:

```cpp
struct State {
    uint8_t* buff;
    size_t cursor;
    CStringView view;

    State() : view(buff) {}
};
```

workspace/__lib__/xod/stream/accumulate-string/patch.cpp
```diff
@@ -0,0 +1,38 @@
...
+    size_t cap = getValue<input_LEN>(ctx);
+
+    if (isSettingUp()) {
+        state->buff = new uint8_t[cap];
+    }
+
+    if (isSettingUp() || isInputDirty<input_RST>(ctx)) {
+        state->cursor = 0;
+        for (int i = 0; i < cap; i++) {
+            state->buff[i] = '\0';
```
nkrkv: Better use `memset`:

```cpp
memset(state->buff, 0, cap);
```

workspace/__lib__/xod/stream/accumulate-string/patch.cpp
```diff
@@ -0,0 +1,38 @@
+
+struct State {
+    uint8_t* buff;
+    size_t cursor;
```
nkrkv: You can make the node slightly more efficient if replace the meaning of `cursor` from “index” to “pointer”. That is, make `cursor` a `char*` pointing to the exact position in `buff`.

workspace/__lib__/xod/stream/accumulate-string/patch.cpp
```diff
@@ -0,0 +1,38 @@
+
+struct State {
+    uint8_t* buff;
```
nkrkv: Are we talking about strings? `char*` is more appropriate type here.

workspace/__lib__/xod/stream/pulse-on-sequence(byte,string)/patch.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+{{ GENERATED_CODE }}
+
+char charAt(XString str, int pos) {
+    auto it = str.iterate();
+    for (int i = 0; i < pos; i++) {
+        ++it;
+    }
+
+    return (char)*it;
```
nkrkv: Is the `(char)` cast necessary? IIRC `*it` returns a char on its own.

workspace/__lib__/xod/net/mac-address/patch.cpp
```diff
@@ -0,0 +1,21 @@
+
+struct State {
+    uint8_t mac [6] = {};
```
nkrkv: No need for `{}` initializer. Please, do not place a space before `[6]`.

evgenykochetkov:
@brusherru @nkrkv check it out again, please

nkrkv:
Here are my observations and complaints:

### format-http-request

I think `METH` is a more intuitive abbrev than `MTD`. I used to think it is ubiquitous but actually failed to find a proof. Another opinion is required.

Internally it uses `format-integer` with a zero precision. Isn’t it a good moment to introduce `format-integer` which might be implemented via `itoa` and be a bit more efficient and compact?

### socket-number

Is just `socket` a better name? Then, pins of such type can be named `SOCK` which is a bit more intuitive than `SN` in my opinion.

### accumulate-string

Suggest renaming `LEN` to `CAP` since we are talking more about capacity than about a desired *length*. `CAP` might be a universal abbrev for many similar cases.

The unnamed outputs confused me a bit. Especially the unnamed pulse. Does it triggers if the buffer is full or not?! It looks like a `DONE` in other cases actually. But the `ERR` is named `FULL`. Maybe we should label them as `STR`/`OK`/`FULL`?

### parse-integer

To be consistent with others, suggest re-labeling:

- `C` → `CHAR`
- `NC` → `PUSH`

### pass-from

Just a bore... Should we name it `skip-until` because some functional programming libraries name things this way. Cargo cult, not sure.

### pass-from-sequence

- `SQN` → `SEQ`? Not sure

### \*-connect-\*

I see a naming clash between:

- `connect-dhcp`
- `connect-with-local-ip`
- `client-connect`
- `is-connected`

They all use the same word to denote different connections. I have no obvious idea for a resolution. Given that an average user “connects to WiFi” and then “opens a site” maybe they should be `connect-dhcp` / `open-client`.

A more formal variant is to include `lan` or `wan` word: `connect-lan-dhcp` / `connect-wan`

Not sure what will work best.

### client-connect

- `SN` → `SOCK`?
- `INIT` → `CONN`?

### has-available-byte

To be consistent with other nodes the outputs should be `{ OUT :: Boolean, DONE :: Pulse }`. Imagine, you lift the node so that it returns the number of bytes available rather than the fact. `Y`/`N` will not work in this case. However, I see how you use them and they are convenient with the current interface. So, I’m not sure whether we should change it or no.

More opinions are required.

### is-connected

The same

### print

Isn’t *print* is too vague? Printer prints things but an ethernet chip does not. Should it be `write-string` instead to align with `write-byte`? Ethernet chip does not write anything though as well.

Should the `LINE` be `STR`? We do not impose should it be a line actually or some text piece in the middle of an abstract character stream.

### request

It knows it is about HTTP. Should it be `http-request` then?

### read-response

It’s too generic to know whether it reads a response of a part of some exchange. Should it be something like `read-string-stream` instead?

brusherru:
`client-connect` -> `open-tcp`?

About `is-connected` it is not clear do we check connection with "internet" or connection with a server. So in the second case, it could have a name `is-socket-open` or `is-tcp-connected` or something like this. What do you think?

About pins of `has-available-byte` and `is-connected`.
The proposed pins look reasonable, but in our case, this solution will require additional nodes like `gate`. So I do not see much profit in this now. 🤔 

About `read-string-stream`. I think it's a general name, but we're acting here with a response from the server via our device. Maybe better to name it `receive-stream` or just `receive`?

In the rest, agree with @nkrkv comment.

workspace/__lib__/xod/net/ip-address/patch.cpp
```diff
@@ -0,0 +1,18 @@
...
+
+void evaluate(Context ctx) {
+    auto octet1 = getValue<input_IN1>(ctx);
+    auto octet2 = getValue<input_IN2>(ctx);
+    auto octet3 = getValue<input_IN3>(ctx);
+    auto octet4 = getValue<input_IN4>(ctx);
+
+    uint8_t bytes[4] = { octet1, octet2, octet3, octet4 };
+
+    emitValue<output_OUT>(ctx, *reinterpret_cast<Type*>(bytes));
```
brusherru:
```
warning: dereferencing type-punned pointer will break strict-aliasing rules [-Wstrict-aliasing]
   emitValue<output_OUT>(ctx, *reinterpret_cast<Type*>(bytes));
```

Use `union`? @nkrkv ?

workspace/__lib__/xod-dev/w5500/client-connect/patch.cpp
```diff
@@ -0,0 +1,31 @@
...
+    dump(serverName, serverNameBuff);
+    serverNameBuff[len] = '\0';
+
+    if (client.connect(serverNameBuff, port)) {
+        emitValue<output_DONE>(ctx, 1);
+    } else {
+        emitValue<output_ERR>(ctx, 1);
+    }
+
+    emitValue<output_SN>(ctx, client.getSocketNumber());
```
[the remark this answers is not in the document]
evgenykochetkov: Yep. It will return `MAX_SOCK_NUM`

evgenykochetkov:
Only documentation remain, and since @brusherru's work depends on `xod/net` etc, we can merge this and add missing docs as a separate PR

PR #1317, opened 2018-07-03
Char literal for byte type

evgenykochetkov:
Now in addition to hex, dec and bin literals, you can bind chars into `byte` inputs in the form `'a'` or `'\n'`. Should be very handy for processing strings.

### DIFF
--- a/docs/reference/data-types/README.md
+++ b/docs/reference/data-types/README.md
@@ -279,6 +279,14 @@ In some inputs IDE or CLI can normalize almost valid literals like lower-cased
       <td class="right aligned"><code>006d</code></td>
       <td>The leading zeros are allowed but may be omitted</td>
     </tr>
+    <tr>
+      <td class="right aligned"><code>'a'</code></td>
+      <td>A character surrounded by single quotes is translated to a byte value equal to its ASCII code</td>
+    </tr>
+    <tr>
+      <td class="right aligned"><code>'\n'</code></td>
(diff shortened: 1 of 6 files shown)


evgenykochetkov:
Added missing tests for `xod-project`'s `getTypeFromLiteral ` 😬

evgenykochetkov:
@nkrkv does [the addition to data types reference](https://github.com/xodio/xod/pull/1317/files#diff-66f611a2699b9e65b366fbcdb839c1f5) look ok? I have some doubts about it.

docs/reference/data-types/README.md
```diff
@@ -279,6 +279,14 @@ In some inputs IDE or CLI can normalize almost valid literals like lower-cased
       <td class="right aligned"><code>006d</code></td>
       <td>The leading zeros are allowed but may be omitted</td>
     </tr>
+    <tr>
+      <td class="right aligned"><code>'a'</code></td>
+      <td>A character surrounded by single quotation marks</td>
```
nkrkv: Maybe:

> A character surrounded by single quotes is translated to a byte value equal to its ASCII code

docs/reference/data-types/README.md
```diff
@@ -279,6 +279,14 @@ In some inputs IDE or CLI can normalize almost valid literals like lower-cased
...
+    </tr>
+    <tr>
+      <td class="right aligned"><code>'\n'</code></td>
+      <td>Escaped character</td>
```
nkrkv: A backslashed character specifies a \[control ASCII character\](https://en.wikipedia.org/wiki/Control_character) like line feed, carriage return, tab, etc.

PR #1330, opened 2018-07-05
More consistent snackbar messages

### LINKED ISSUE #955: Use headlinese for snackbar message titles
### Rationale

Currently, the message/error/info titles are not consistent. Some are verbose other are short. With #940 we got message notes where we can be verbose, but to make the popups expressive we should make all titles very short and precise.

Let’s use [headlinese](https://en.wikipedia.org/wiki/Headlinese) style for it. See https://english.stackexchange.com/questions/198869/style-of-technical-warnings-why-is-it-acceptable-to-omit-verbs

In particular, we should follow the rules:

> Forms of the verb "to be" and articles (a, an, the) are usually omitted.

> Most verbs are in the simple present tense, e.g. "Governor signs bill", while the future is expressed by an infinitive, with to followed by a verb, as in "Governor to sign bill".

We also should get rid of exclamations because they are superfluous and non-informative.

nkrkv:
Closes #955


packages/xod-client/src/project/messages.js
```diff
@@ -9,36 +9,40 @@ export const SUCCESSFULLY_PUBLISHED = {
...
   },
   [LE.SAME_NODE]: {
-    title: 'Canʼt create link between pins of the same node!',
-    persistent: false,
-  },
-  [LE.UNKNOWN_ERROR]: {
-    title: 'Canʼt create link',
-    note: 'Unknown error!',
-    persistent: false,
+    title: 'No-no-no',
```
evgenykochetkov: 👍🏻

PR #1342, opened 2018-07-09
Add abstract http-request patches to xod/net

evgenykochetkov:
We discussed that it is needed, but forgot to add it 😬

Also adds `http-get` and `http-post` that simply wrap `http-request `.


workspace/__lib__/xod/net/http-request/patch.xodp
```diff
@@ -0,0 +1,151 @@
...
+      "label": "METH",
+      "position": {
+        "x": 203,
+        "y": 0
+      },
+      "type": "xod/patch-nodes/input-string"
+    },
+    {
+      "boundLiterals": {
+        "__out__": "\"\""
```
brusherru: Oops, looks like default string literals aren't removed from xodball 🙈

evgenykochetkov:
Also, 
- renamed `xod-dev/esp8266/inet` and `xod-dev/w5500/inet` to avoid name clashing when resolving types
- renamed `xod-dev/w5500/ethernet-shield-device` to `xod-dev/w5500/w5500-device`
- added `xod-dev/w5500/ethernet-shield`

PR #1346, opened 2018-07-10
Install toolchains with antivirus without SSL errors

### LINKED ISSUE #1291: Fetch Error: unable to verify the first certificate during deploy/upload to Arduino.
A weird SSL bug which affects newcomers and makes XOD look like malware which gladly reports itself :smiley: 

### Environment

- Scope: desktop IDE
- Version: 0.21.0
- OS: Windows 10

### Steps to reproduce

On a fresh install, try to upload to trigger toolchain download

### Expected behavior

The toolchain is installed, the upload goes on

### Actual behavior

Error:

> Fetch Error: request to https://storage.googleapis.com/releases.xod.io/packages/win32-avr-tools.tar.gz failed, reason: unable to verify the first certificate”, during deploy/upload to Arduino.

### Details

The bug was reported by several users with Windows machines. In one particular case, it was due to Kaspersky enabled.

The hypothesis is that something wrong with cert for `googleapis.com` (which we have no influence on) *and* the CA data on Windows is somewhat stale (a user did not update for a long time).

### How to fix

Try to use the latest CA list explicitly: https://stackoverflow.com/questions/31673587/error-unable-to-verify-the-first-certificate-in-nodejs

brusherru:
It closes #1291.

### DIFF
--- a/packages/xod-arduino-deploy/src/packageManager.js
+++ b/packages/xod-arduino-deploy/src/packageManager.js
@@ -1,3 +1,5 @@
+import https from 'https';
+import http from 'http';
 import * as R from 'ramda';
 import path from 'path';
 import fse from 'fs-extra';
@@ -7,6 +9,16 @@ import bz2 from 'unbzip2-stream';
 import { tapP } from 'xod-func-tools';
 
 import * as Utils from './utils';
+
+// =============================================================================
+//
+// HTTP/HTTPS agents
+//
+// =============================================================================
+const httpsAgent = new https.Agent({
+  rejectUnauthorized: false,
+});
+const httpAgent = new http.Agent({});
 
 // =============================================================================
 //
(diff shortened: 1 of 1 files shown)


packages/xod-arduino-deploy/src/packageManager.js
```diff
@@ -8,6 +10,16 @@ import { tapP } from 'xod-func-tools';
 
 import * as Utils from './utils';
 
+// =============================================================================
+//
+// HTTP/HTTPS agents
+//
+// =============================================================================
+const httpsAgent = new https.Agent({
+  rejectUnauthorized: false,
```
nkrkv: That’s not a solution, but rather a security hole.

The problem should be solved by fixing the CA/valid cert lists.

brusherru:
Alright, I tried some other ways to solve it, but nothing helps.
Let's try to write to Google Support (cause third certificate points on 404: http://ocsp.pki.goog/GTSGIAG3 ) or attach it to our domain and our certificate.

PR #1322, opened 2018-07-04
Support ESP8266 as Internet provider

brusherru:
It closes #1286

PR #1347, opened 2018-07-10
Wrap output pin values with pragmas only once

brusherru:
It should be in PR #1322 🙈

### DIFF
--- a/packages/xod-arduino/platform/program.tpl.cpp
+++ b/packages/xod-arduino/platform/program.tpl.cpp
@@ -10,14 +10,18 @@
 namespace xod {
 
 // Define/allocate persistent storages (state, timeout, output data) for all nodes
-{{#each nodes}}
-{{~ mergePins }}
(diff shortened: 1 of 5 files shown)


packages/xod-arduino/platform/program.tpl.cpp
```diff
@@ -10,14 +10,17 @@
...
-{{~ mergePins }}
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wmissing-field-initializers"
+{{#each nodes}}
+{{~ mergePins }}
 {{#each outputs }}
 {{ decltype type value }} node_{{ ../id }}_output_{{ pinKey }} = {{ cppValue type value }};
 {{/each}}
+{{/each}}
 #pragma GCC diagnostic pop
```
nkrkv: Please, add a blank line right above `#pragma GCC diagnostic pop`

PR #1360, opened 2018-07-13
Comment out xod pragmas

### LINKED ISSUE #1323: Comment out XOD pragmas while transpiling
### Rationale

XOD pragmas carry meta information for the transpiler and not a compiler. GCC complains when sees a `#pragma XOD ...` directive:

>warning: ignoring #pragma XOD dirtieness [-Wunknown-pragmas]

We might comment out such pragmas when generating code so that the warning goes away.

### Acceptance criteria

- [ ] Any `#pragma XOD lala` found in `patch.cpp` is converted to `//#pragma XOD lala` in the generated code

### How to implement

- Search and replace with a trivial regex should be enough

brusherru:
It closes #1323

### DIFF
--- a/packages/xod-arduino/src/transpiler.js
+++ b/packages/xod-arduino/src/transpiler.js
@@ -94,6 +94,11 @@ const arrangeTPatchesInTopologicalOrder = def(
     )(project)
 );
 
+export const commentXodPragmas = def(
+  'commentXodPragmas :: String -> String',
+  R.replace(/^\s*#\s*pragma\s+XOD\s/gm, '//#pragma XOD ')
+);
+
 const convertPatchToTPatch = def(
   'convertPatchToTPatch :: Patch -> TPatch',
   patch => {
@@ -149,7 +154,7 @@ const convertPatchToTPatch = def(
       {
         outputs,
         inputs,
-        impl,
+        impl: commentXodPragmas(impl),
         patchPath,
       },
     ]);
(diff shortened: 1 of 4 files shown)


packages/xod-arduino/src/transpiler.js
```diff
@@ -94,6 +94,11 @@ const arrangeTPatchesInTopologicalOrder = def(
     )(project)
 );
 
+const commentXodPragmas = def(
+  'commentXodPragmas :: String -> String',
+  R.replace(/#\s*pragma XOD/, '//#pragma XOD')
```
nkrkv: The following should not be commented out:

```cpp
#pragma           XODULAR
```

Note multiple spaces and a custom continuation

brusherru:
@nkrkv check it out, please

packages/xod-arduino/src/transpiler.js
```diff
@@ -94,6 +94,11 @@ const arrangeTPatchesInTopologicalOrder = def(
     )(project)
 );
 
+const commentXodPragmas = def(
+  'commentXodPragmas :: String -> String',
+  R.replace(/#\s*pragma\s+XOD\s/, '//#pragma XOD ')
```
nkrkv: :smiling_imp: 

Will it also comment the following?

```cpp
Serial.println("#pragma XOD hello")
```

It must not. Suggest prefixing with the beginning of the line + possible whitespace before `#`.

By the way, I see the small issue is big enough to put a few unit tests on it.

PR #1361, opened 2018-07-13
Fix dev:* scripts

evgenykochetkov:
Before:
- run `yarn dev:browser`
- change something in `xod-project`
- nothing happens

Now, changes are correctly picked up.


evgenykochetkov:
Unfortunately, this still does not apply to Reason files. `lerna` does not seem to play well with `bsb -w`.
But that's a separate issue.

PR #1392, opened 2018-07-30
The Buses

### LINKED ISSUE #1353: As a xoder I want to create data buses so that I can get rid of repetitive links clutter
### Rationale

Quite often a scenario appears when a single output should be linked to very many inputs spread all over the patch. It leads to visual noise where the important links get lost:

screenshot from 2018-07-12 15-40-14

The problem arises when we make:

- An imperative chain over a single object like I2C
- ERR convergence
- Get I²C ADDR from an input terminal
- Link RST terminal to all RST’s of the internal nodes
- Link a common enabled/disabled flip-flop to shared hardware resource nodes

### User story

Let’s consider a possible refactoring of the example above to make it simpler visually.

1. I delete all outgoing links from the I2C terminal.
2. I add `xod/patch-nodes/to-bus` and label the node as `I2C`
3. I link `to-bus` input to the terminal output
4. I add one `xod/patch-nodes/from-bus` per deleted link right above nodes with I2C inputs
5. I link `from-bus` outputs to corresponding `I2C` inputs
6. I label all `from-bus`’es as `I2C`

As a result, I get semantically the same program with the same code.

### Acceptance criteria

- [ ] I can add `to-bus` and `from-bus` nodes
- [ ] Bus nodes with the same label act as a link network
- [ ] Empty label is a valid label, just one of the possible variants
- [ ] Pins of `to-bus` nodes are grey when placed
- [ ] Pins of `to-bus` get a type-color of the upstream pin when linked
- [ ] A `from-bus` node and its output pin are marked as an error, transpilation fails¹ if no matching `to-bus` found
- [ ] Pins of `from-bus` get a type-color of matching `to-bus`
- [ ] If an implicit cast for a link between `from-bus` and a downstream pin does not exist, the link is marked with an error like it would in other scenarios
- [ ] If multiple `to-bus` with the same label created, they are marked with an error along with all matching `from-bus`’es, transpilation fails²
- [ ] If a `to-bus` with a matching `from-bus` floats (not linked), they are marked with an error, transpilation fails³
- [ ] Graph cycles caused by chaining several buses are not allowed. The same error as for regular graph cycles might be shown.

¹
```markdown
**No bus source**
@/main -> @foo: | Bus 'FOO' does not exist |.
Create a `to-bus` node with label 'FOO' to define the required bus.
```

²
```markdown
**Multiple bus sources**
@/main -> @foo: | Bus 'FOO' has multiple conflicting sources |
Delete or rename one of `to-bus` nodes so that the bus get a single source of data.
```

³
```markdown
**Bus floats**
@/main -> @foo: | Bus 'FOO' source is not linked anywhere |
Link the `to-bus` with label `FOO` to an output pin of another node
```

### Out of scope

- Custom look for bus nodes (#1364)
- Hotkeys (#1365)

### How to test

Prepare two equivalent patches with buses and with direct linkage. Verify that they have an equivalent normal form.

## How to implement

- Make a function (`jumperizeBuses`) which replaces all matching bus pairs with a `xod/patch-nodes/jumper` node
- Do development-time validation and type deduction after the jumperization
- `jumper` is basically no-op and should be thrown away at the latter stages of the graph normalization

### LINKED ISSUE #1364: Outstanding look for bus nodes
### Rationale

The bus nodes are so special that deserve their own look. Making a parallel to electronics, let it be an isosceles triangle. For `to-bus` it points down and for `from-bus` it points up. It will not take up the whole slot in height. The spare vertical space can be used to render the bus label.

### Acceptance criteria

- [ ] Buses render as triangles on the patch board
- [ ] Buses render as triangles in helpboxes

### Out of scope

Adjusting the blue shadow anyhow.

### How to implement

As it done for terminal nodes.

evgenykochetkov:
Closes #1353,
Closes #1364


packages/holes/src/Holes_Map_String.rei
```diff
@@ -27,3 +27,7 @@ let innerJoin: (t(k), t('v)) => t('v);
 /** Merges two maps extending the first map with contents of the second. If a key
     presents in both maps, the value of the second map wins. */
 let mergeOverride: (t('v), t('v)) => t('v);
+
+let fromDict: Js.Dict.t('v) => t('v);
+
+let toDict: t('v) => Js.Dict.t('v);
```
nkrkv: Every symbol in a `*.rei` should have a doc-string, even if it is quite simple.

packages/xod-project/src/Buses.rei
```diff
@@ -0,0 +1,12 @@
...
+  A list of tuples with `to-bus` node as the first element
+  and a list of corresponding `from-bus` nodes as the second
+*/
+type matchingBusNodes = list((Node.t, list(Node.t)));
+
+let getMatchingBusNodes: Patch.t => matchingBusNodes;
+
+let jumperizePatch: (Patch.t, matchingBusNodes) => Patch.t;
+
+let jumperizePatchRecursively: (Project.t, Patch.path) => Project.t;
```
nkrkv: All `*.rei` symbols have to be documented :sob:

packages/xod-project/src/Link.re
```diff
@@ -1,7 +1,43 @@
...
 
+type id = string;
+
 [@bs.module ".."]
 external create :
   (~toPin: Pin.key, ~toNode: Node.id, ~fromPin: Pin.key, ~fromNode: Node.id) =>
   t =
   "createLink";
+
+[@bs.module ".."] external getId : t => id = "getLinkId";
```
nkrkv: Chris Carter would call it `[@bs.outthere]` :alien:

packages/xod-project/src/TypeDeduction.re
```diff
@@ -0,0 +1,110 @@
...
+    };
+
+  let dissoc: (t, Node.id, Pin.key) => t = (drs, nodeId, pinKey) => {
+    let updatedNodeTypes =
+        drs
+        |. Belt.Map.String.getWithDefault(nodeId, Belt.Map.String.empty)
+        |. Belt.Map.String.remove(pinKey);
+
+    if (Belt.Map.String.isEmpty(updatedNodeTypes) ) {
+      drs |. Belt.Map.String.remove(nodeId);
```
nkrkv: Bet it will be more clear to just `open` the `Belt`… For the whole module

packages/xod-project/src/TypeDeduction.re
```diff
@@ -0,0 +1,110 @@
...
+              )
+            );
+
+          deductionResults
+          |. DeductionResultsMap.dissoc(jumperNodeId, jumperOutPinKey)
+          |. copyResultToFromBusNodes;
+        };
+      },
+    );
+  };
```
nkrkv: EOL is missing.

:open_mouth: and a corresponding `*.rei` is missing too

packages/xod-project/src/Buses.re
```diff
@@ -0,0 +1,88 @@
+open Belt;
+
+type matchingBusNodes = list((Node.t, list(Node.t)));
+
+let getMatchingBusNodes = patch => {
+  let nodes = Patch.listNodes(patch);
+
+  let toBusNodesByLabel =
+    nodes
+    |. List.keep(n => Node.getType(n) == "xod/patch-nodes/to-bus")
```
brusherru: Better use constant or function from JS-part of `xod-project`

packages/xod-project/src/Buses.re
```diff
@@ -0,0 +1,88 @@
...
+    /* having multiple `to-bus` nodes with the same label is forbidden,
+       so exclude them from type resolution */
+    |. Map.String.keep((_, ns) => List.length(ns) == 1)
+    |. Holes.Map.String.keepMap(List.head);
+
+  if (Map.String.isEmpty(toBusNodesByLabel)) {
+    []; /* short-curcuit for optimization */
+  } else {
+    nodes
+    |. List.keep(n => Node.getType(n) == "xod/patch-nodes/from-bus")
```
brusherru: The same one.

packages/xod-project/src/project.js
```diff
@@ -1074,6 +1075,7 @@ const getDependenciesMap = (project, patchPaths, depsMap) => {
   const currentPatchDeps = R.compose(
     R.map(Node.getNodeType),
     Patch.listNodes,
+    // TODO: make safer?
```
brusherru: Sounds good :)

PR #1393, opened 2018-07-31
Bump node-sass to 4.9.2

evgenykochetkov:


nkrkv:
Bumping `node-sass` alone is not enough to switch to Node 10. For example, `serialport` offends too. I’m trying to provide a complete solution in https://github.com/xodio/xod/compare/chore-node-10

evgenykochetkov:
Well, this particular PR is more about "running `yarn dev:browser` without issues on node 10" than "completely switching to node 10" :)

PR #1395, opened 2018-08-02
Load built-in patches from a special .xodball instead of generating them programmatically

evgenykochetkov:
- it breaks circular dependencies in `xod-project`'s code
- it allows to easier document built-in patches


packages/xod-project/tools/loadBuiltinPatches.js
```diff
@@ -0,0 +1,44 @@
+const fs = require('fs');
+const path = require('path');
+const R = require('ramda');
+
+const { explodeEither } = require('xod-func-tools');
+const XP = require('..');
+
+const targetPath = path.resolve(__dirname, '../builtinPatches.json');
+const xodballPath = path.resolve(__dirname, '../built-in-patches.xodball');
```
nkrkv: I think it would be more clear if the build artifact basename matches the source name. For example, `builtin-patches.xodball` and `builtin-patches.json`.

Also, we have a directory dedicated to build artifacts. It is `dist/`. So it is a good idea to put the JSON there. Yes, it will imply importing with `../dist/builtin-patches.json`, but in case of an error it communicates much better than the absence of `builtinPatches`. A bunch of special ignore/clear hacks will also disappear if moved to `dist/`

PR #1396, opened 2018-08-03
Optimize IDE hinting

### LINKED ISSUE #1357: Optimize IDE hinting
Some manipulations with patches in XOD IDE is much slower than they might be because on *any* change in the `project` Redux branch many re-validations and type deductions take place. We can improve it by slightly changing the architecture.

The idea is rather than keeping validation/type deduction information along with every particular node in the state we can store it in a separate Redux branch. Let’s call it `hints`: after all, the errors and deduced types are just *hints* for a xoder which don’t matter anything for XOD until he tries to deploy.

After the split, we can *postpone* validation and type deduction to be done after vital updates such as node positioning are done. To achieve it, we move the responsibility of re-validation from selectors to a centralized Redux hinting middleware. If it sees a change in the project, it can update the state with a requirement for re-validation (`redux-loop` seems like a good fit). After the vital updates are done the re-validation requirements are satisfied.

The hinting middleware can make short-circuits to skip re-validations completely or granularly. For example, if it detects a simple movement of a non-terminal node, a re-validation may be skipped entirely. Or, if a link is deleted a check for dead links may be skipped, but other checks and deductions should still take place.

We also have some inconsistencies on who validates the program integrity and when. It causes redundant validity checks in (a) actions → (b) assoc/dissoc in xod-project → (c) re-validation on program change. We can get rid of this redundancy by removing the integrity checks from (a) and (b), leaving only (c).

### Outline

Putting task summary:

- [ ] Move validation and deduction results to a separate hinting branch in Redux
- [ ] Validate through a middleware, not selectors
- [ ] Make short-circuits in middleware for the most innocent operations¹
- [ ] Do not check for integrity breaks in actions
- [ ] Do not check for integrity breaks in assoc/dissoc nodes and links in `xod-project`

¹ Innocent and frequent operations on nodes:

- Movement (non-terminal, non-marker)
- Giving a label (non-terminal)
- Binding without changing a literal type
- Updating C++ code
- Updating a description

brusherru:
It closes #1357 🙌


packages/xod-client/src/hinting/middleware.js
```diff
@@ -0,0 +1,45 @@
...
+    ? validateProject(action, newProject, nextDeducedTypes, prevErrors)
+    : prevErrors;
+  const willUpdateErrors = notEquals(prevErrors, nextErrors);
+
+  if (willUpdateDeducedTypes && willUpdateErrors) {
+    next(updateHinting(nextDeducedTypes, nextErrors));
+  } else if (willUpdateDeducedTypes) {
+    next(updateDeducedTypes(nextDeducedTypes));
+  } else if (willUpdateErrors) {
+    next(updateErrors(nextErrors));
```
nkrkv: Why `next` here and not `store.dispatch`? Yeah, the actions are intrinsic to the middleware, but conceptually should be visible to any middleware in the chain.

packages/xod-client/src/hinting/typeDeduction.js
```diff
@@ -0,0 +1,89 @@
...
+
+// =============================================================================
+//
+// Type deduction
+//
+// =============================================================================
+
+// Run type deduction only for some actions
+// :: Project -> Action -> Boolean
+export const shallDeduceTypes = R.curry((project, action) =>
```
nkrkv: I see a conceptual flaw with `shallDeduceTypes` implementation. It’s a whitelist instead of a blacklist and thus fragile. That is, it returns `false` (will not deduce types) by default and returns `true` only in some cases foreseen by a developer. It seems like a big source of bugs. I did not test it yet, but I’m quite sure from what I see, the following actions will not update deduced types:

- Node deletion
- Link deletion

They are just out of my head, who knows how many actions are actually forgotten.

Better if the function will return `true` (will deduce types) by default and `false` in some short-circuit cases. Please, emphasize it in comments as well.

packages/xod-client/src/hinting/utils.js
```diff
@@ -0,0 +1,66 @@
+import * as R from 'ramda';
```
nkrkv: Please, no more modules with a name `utils`/`helpers`/`tools` and other “one size fits all”. Name it after the function it carries (`errorCollectors`?).

packages/xod-client/src/hinting/validators.internal.js
```diff
@@ -0,0 +1,486 @@
...
+);
+
+// :: { links: Map LinkId { errors: Map ErrorType [Error] } } -> Boolean
+const haveNoLinkErrors = R.either(
+  propEmpty('links'),
+  R.compose(R.all(haveNoErrors), R.values, R.prop('links'))
+);
+
+// :: PatchErrors -> Boolean
+const haveNoAnyErrors = R.allPass([
```
nkrkv: No one will say “have no any errors”. Should be “haveNoErrors”. Ah, it’s in conflict with another function. That function should get a sensible name then.

packages/xod-client/src/hinting/validators.internal.js
```diff
@@ -0,0 +1,486 @@
...
+// :: Action -> Project -> Map PatchPath DeducedPinTypes -> Map PatchPath PatchErrors -> ErrorsUpdateData
+export const generalValidator = (
+  action,
+  project,
+  allDeducedPinTypes,
+  prevErrors
+) => {
+  const maybePatchPath = getActingPatchPath(action);
+  if (Maybe.isJust(maybePatchPath)) {
+    const patchPath = explodeMaybe('IMPOSSIBLE ERROR', maybePatchPath);
```
nkrkv: Quite an ugly construct. Why not `bimap`?

packages/xod-client/src/hinting/validators.internal.js
```diff
@@ -0,0 +1,486 @@
...
+  ) => {
+    const validateFn = validatePatches(
+      [R.always({}), ...nodeValidators],
+      [R.always({}), ...pinValidators],
+      [R.always([]), ...linkValidators]
+    );
+
+    const maybePatchPath = getActingPatchPath(action);
+    if (Maybe.isJust(maybePatchPath)) {
+      const patchPath = explodeMaybe('IMPOSSIBLE ERROR', maybePatchPath);
```
nkrkv: `bimap`?

packages/xod-client/src/hinting/validators.internal.js
```diff
@@ -0,0 +1,486 @@
...
+// :: PatchErrors -> Boolean
+const haveNoAnyErrors = R.allPass([
+  haveNoErrors,
+  haveNoNodeErrors,
+  haveNoLinkErrors,
+]);
+
+// =============================================================================
+//
+// API
```
nkrkv: The API functions short doc comments are much desired :pray:  I can’t understand all that nuances of generally/not-generally, policies, `callFnIfExist` O_o

packages/xod-client/src/hinting/validators.js
```diff
@@ -0,0 +1,230 @@
...
+/**
+ * A map of short-circuit validations and exceptions to the general rules
+ * of validation.
+ * For example of exception case see `PAT.PATCH_ADD`.
+ *
+ * Indexed by ActionTypes.
+ * Should return `ErrorsUpdateData` { policy: UPDATE_ERRORS_POLICY, errors: Map PatchPath PatchErrors }
+ *
+ * Better and simplies way to produce this type is apply `set*Policy` function
+ * as a last step of custom validation.
```
nkrkv: **A** better and **simpler** way to produce this type is **to** apply `set*Policy` function as **the** last step of **a** custom validation.

packages/xod-client/src/hinting/validators.js
```diff
@@ -0,0 +1,230 @@
...
+ * There are three policy rules:
+ * `setMergePolicy`      - will merge new errors deeply inside previvous errors.
+ *                         It's handy when you check only one thing, not whole
+ *                         patch or patches.
```
nkrkv: **previous** … **the** whole

packages/xod-client/src/hinting/validators.js
```diff
@@ -0,0 +1,230 @@
...
+ * `setAssocPolicy`      - will overwrite errors only for listed patch paths.
+ *                         Use it when you validates some patches for all
+ *                         possibly errors. E.G. `validateLocalPatches`.
```
nkrkv: you **validate** … all **possible** errors

packages/xod-client/src/hinting/validators.js
```diff
@@ -0,0 +1,230 @@
...
+ * `setOverwritePolicy`  - will overwrite all errors with new ones.
+ *                         Use it only when you validates whole project,
```
nkrkv: **the** whole

packages/xod-client/src/hinting/validators.js
```diff
@@ -0,0 +1,230 @@
...
+ *                         including library patches.
+ *
+ * Signature for each function in this map:
+ * :: Action -> Project -> Map PatchPath DeducedPinTypes -> Map PatchPath PatchErrors -> ErrorsUpdateData
+ *
+ * But there is a pack of utility functions, that make validation easier:
+ *
+ * To validate only one Patch, that was referenced in the Action, call:
+ * validateChangedPatch :: [NodeValidateFn] -> [PinValidateFn] -> [LinkValidateFn] ->  Action -> Project -> Map PatchPath DeducedPinTypes -> Map PatchPath PatchErrors -> Map PatchPath PatchErrors
```
nkrkv: Line wrapping here and below will make the docs more readable:

```
validateChangedPatch ::
  [NodeValidateFn] ->
  [PinValidateFn] ->
  [LinkValidateFn] ->
  Action ->
  Project ->
  Map PatchPath DeducedPinTypes ->
  Map PatchPath PatchErrors ->
  Map PatchPath PatchErrors
```

packages/xod-client/src/hinting/validators.js
```diff
@@ -0,0 +1,230 @@
...
+      R.map(libName => {
+        // TODO: Replace with `R.takeWhile` from newer Ramda
+        const index = libName.indexOf('@');
+        return libName.substring(0, index !== -1 ? index : libName.length);
+      }),
+      R.keys,
+      R.path(['payload', 'projects'])
+    )(action);
+
+    const isAmongToInstalledLibs = R.compose(R.anyPass, R.map(R.startsWith))(
```
nkrkv: “among to”? I guess you’ve meant just “among”

packages/xod-func-tools/src/lists.js
```diff
@@ -56,3 +56,11 @@ export const uniqLists = def(
     []
   )
 );
+
+/**
+ * Concatenates list of lists into one list
+ */
+export const concatLists = def(
```
nkrkv: Isn’t it a `R.chain`?

evgenykochetkov: That's `R.unnest`

packages/xod-client/src/hinting/typeDeduction.js
brusherru: > but I’m quite sure from what I see, the following actions will not update deduced types

It will because of action type called "BULK_DELETE_ENTITIES" is listed.

I understand concerns about the whitelist, but I'm afraid that in this case almost all actions, except those listed, will be blacklisted. Is it OK?

nkrkv: I think it is OK. In this case, the worst thing we get is a slow-down. In the current implementation, the worst thing is buggy behavior. I’d prefer the first. Being real, only actions from `./packages/xod-client/src/project/actionTypes.js` can lead to re-deduction (because only they change the `project` branch) and there are less than 50 actions there. So white-listing a few dozens of actions should not be a big deal: start with copy-paste and remove those which can affect the type deduction.

packages/xod-client/src/hinting/validators.internal.js
evgenykochetkov: `Maybe` is not a bifunctor😔
Possibly, `foldMaybe` will do?

packages/xod-func-tools/src/objects.js
```diff
@@ -143,3 +143,10 @@ export const sameKeysetBy = def(
     return true;
   }
 );
+
+// :: [StrMap [a]] -> StrMap [a]
+export const mergeAllWithConcat = R.reduce(R.mergeWith(R.concat), {});
+
+// ObjectOrArray :: [a] | StrMap ObjectOrArray
+// :: [StrMap ObjectOrArray] -> StrMap ObjectOrArray
+export const mergeAllDeepWithConcat = R.reduce(R.mergeDeepWith(R.concat), {});
```
evgenykochetkov: That's a _very_ specific function. And it's name basically matches the implementation.
Do we really need to bloat our API with it?

packages/xod-client/src/hinting/validators.internal.js
```diff
@@ -0,0 +1,486 @@
...
+  XP.validatePinLabels
+);
+
+// TODO: Use validator from xod-project after refactoring
+// :: Patch -> Project -> Map NodeId (Map ErrorType [Error])
+export const getDeadRefErrorMap = (patch, project) =>
+  R.compose(
+    R.map(
+      R.compose(
+        R.objOf('checkPatchExists'),
```
evgenykochetkov: There are quite a lot of "magic" strings like this. Maybe they should be extracted to some constant?
Or, since most of them are named after validation functions, use those functions `name` property?

brusherru:
@nkrkv @evgenykochetkov check it out, please! :)

packages/xod-client/src/hinting/typeDeduction.js
```diff
@@ -1,21 +1,74 @@
...
+    R.propSatisfies(
+      R.complement(
+        isAmong([
+          PAT.PROJECT_OPEN_REQUESTED,
+          PAT.PROJECT_PUBLISH_REQUESTED,
+          PAT.PROJECT_PUBLISH_CANCELLED,
+          PAT.PROJECT_PUBLISH_START,
+          PAT.PROJECT_PUBLISH_SUCCESS,
+          PAT.PROJECT_PUBLISH_FAIL,
+          PAT.PROJECT_OPEN_WORKSPACE,
```
nkrkv: Do the actions from the head ever affect the `project` branch in the Redux state? Namely:

```
          PAT.PROJECT_OPEN_REQUESTED,
          PAT.PROJECT_PUBLISH_REQUESTED,
          PAT.PROJECT_PUBLISH_CANCELLED,
          PAT.PROJECT_PUBLISH_START,
          PAT.PROJECT_PUBLISH_SUCCESS,
          PAT.PROJECT_PUBLISH_FAIL,
          PAT.PROJECT_OPEN_WORKSPACE,
```

I mean, if they do not change `project`, no reason to list them here because the check will never be run.

packages/xod-client/src/hinting/typeDeduction.js
```diff
@@ -1,21 +1,74 @@
...
+          PAT.COMMENT_RESIZE,
+          PAT.COMMENT_SET_CONTENT,
+          PAT.BULK_MOVE_NODES_AND_COMMENTS,
+          PAT.PATCH_ADD,
+          PAT.PATCH_DELETE,
+          PAT.PATCH_DESCRIPTION_UPDATE,
+          PAT.PATCH_NATIVE_IMPLEMENTATION_UPDATE,
+          PAT.PATCH_HISTORY_UNDO,
+          PAT.PATCH_HISTORY_REDO,
+          PAT.SAVE_ALL,
```
nkrkv: Ditto: `SAVE_ALL` does not affect the project

packages/xod-client/src/hinting/typeDeduction.js
```diff
@@ -33,57 +86,71 @@ export const shallDeduceTypes = R.curry((project, action) =>
...
-      'type'
-    ),
   ])(action)
 );
 
-// :: Project -> Action -> Maybe DeducedPinTypes
+/**
+ * Returns deduced pin types for only one patch, that are
+ * changed by Action.
+ * Returns Maybe, cause Action could not contain PatchPath.
```
nkrkv: *because Action cannot*

packages/xod-client/src/hinting/validation.internal.js
```diff
@@ -0,0 +1,471 @@
...
+      R.compose(setOverwritePolicy, validateAllPatches)(
+        project,
+        allDeducedPinTypes,
+        null,
+        prevErrors
+      ),
+    // PatchPath exists -> Validate this Patch first
+    patchPath =>
+      R.compose(
+        R.compose(
```
nkrkv: `R.compose(` of `R.compose(` can be replaced with a single `R.compose` which removes one nesting level

packages/xod-client/src/hinting/validation.js
```diff
@@ -0,0 +1,333 @@
...
+  getVariadicMarkersErrorMap,
+  getDeadRefErrorMap,
+  validateBoundValues,
+  validateLinkPins,
+} from './validation.funcs';
+
+/**
+ * HOW IT WORKS AND HOW TO MAINTAIN IT
+ *
+ * First of all, middleware check do we need to validate anything,
```
nkrkv: checks

packages/xod-client/src/hinting/validation.js
```diff
@@ -0,0 +1,333 @@
...
+ * it calls `shallValidate` with passed into it Map of Functions
+ * `shallValidateFunctions`. If it found any function (key is an ActionType)
```
nkrkv: If it finds

packages/xod-client/src/hinting/validation.js
```diff
@@ -0,0 +1,333 @@
...
+ *
+ * I FOUND VERY SLOW CASE, HOW TO TWEAK PERFORMANCE?
+ * There is two ways:
+ * a). if some action could not break something in the Patch
+ *     or Project, or could break it only in the specific cases,
+ *     add a function (or just `R.F`) into `shallValidateFunctions`.
+ * b). if action could break only specific things, add a short-circuit
+ *     validate function into `shortValidators`.
+ *
+ * GOOD LUCK!
```
nkrkv: Cool reading :heart_eyes:

packages/xod-client/src/hinting/validation.js
```diff
@@ -0,0 +1,333 @@
...
+ */
+
+/**
+ * A map of functions, that accepts new project and action
+ * to check shall we need to run any validations on this changes.
```
nkrkv: “shall we run” or “do we need to run”

this → these

packages/xod-client/src/hinting/validation.js
```diff
@@ -0,0 +1,333 @@
...
+} from './validation.internal';
+import {
+  getVariadicMarkersErrorMap,
+  getDeadRefErrorMap,
+  validateBoundValues,
+  validateLinkPins,
+} from './validation.funcs';
+
+/**
+ * HOW IT WORKS AND HOW TO MAINTAIN IT
```
evgenykochetkov: This is awesome 👍🏻

brusherru:
@nkrkv @evgenykochetkov thanks for the review! Just fixed.

brusherru:
@nkrkv @evgenykochetkov check out the last commit, please.

It fixes a bug, that occurred when changes in the Patch did not invalidate Patch, but invalidates other patches.
For example, open `xod/core/concat` and copy a rightmost terminal.

Now it will check changed Patch and all patches that depend on changed one.
It gains some extra time to validate changes. The more often a modified patch is used, the longer it will validate. Up to 1 second in the case with `concat` node.

After approve I'll squash this commit.

nkrkv:
OK, but a bunch of tests will make the system more solid

brusherru:
@nkrkv updated. Check it out, please

packages/xod-client/src/hinting/validation.funcs.js
```diff
@@ -57,36 +57,38 @@ export const getConstructorMarkersErrorMap = getMarkerNodesErrorMap(
...
+        R.compose(
+          R.map(R.of),
+          R.fromPairs,
+          R.map(R.pair(R.__, err)),
+          R.path(['payload', 'nodeIds'])
+        )(err),
+      R.always({})
+    ),
+    XP.validateBuses
+  )(patch);
```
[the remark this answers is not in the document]
brusherru: Added eslint rule that skips arguments starts with an underscore from `no-unused-vars` validation. So now these functions have a signature `(patch, _project)` that tells why do we need to write arguments more obviously.
(We need to pass only patch, cause in other way `hm-def` throws error in the tests with something like "Function accepts only one argument, but it called with two").

PR #1397, opened 2018-08-08
Shortcuts for buses

### LINKED ISSUE #1365: As a xoder I want to make buses with hotkeys to avoid manual labeling of many bus nodes
### Rationale

Giving the same label to a dozen of bus nodes is tiresome. We can make shortcuts to create and configure buses quickly.

### Acceptance criteria

- [ ] I can start linking and press `B` to place a bus node under the cursor
- [ ] A bus node created while linking automatically pulls a label from the linked pin
- [ ] I can select a link and press `B` to replace the link with a pair of to/from buses
- [ ] A bus made from a link takes a name of the output pin
- [ ] The `to-bus` is created one slot below the output pin
- [ ] The `from-bus` is created one slot above the output pin
- [ ] I can select another link from a pin already having a `to-bus` connected and press `B`. Instead of creating a new `to-bus`, the existing will be re-used
- [ ] I can select multiple links and nodes, then press `B` and the result will be equivalent to selecting each link one by one and converting it to a bus pair
- [ ] I can hit “Edit → Convert Selected Links to Buses” to do the same

### Out of scope

- Collision checking
- Name clash checking

evgenykochetkov:
Closes #1365
Also see xodio/xod-docs#52


nkrkv:
Has not hard-tested the functionality yet

nkrkv:
One thing nice to have which was not described in the issue is some special terminal handling.

### Steps to reproduce

1. Place an `input-pulse`
2. Label it `UPD`
3. Start linking its output and press B

### Desired behavior

The bus gets label `UPD`

### Actual behavior

The bus gets label `OUT`

PR #1399, opened 2018-08-09
Fix bug with dead reference error

brusherru:
There is no issue 😬 

**How to reproduce:**
1. Download xodball (it have dead references and links): [deadrefs.xodball.zip](https://github.com/xodio/xod/files/2274206/deadrefs.xodball.zip)
2. Open Project
3. Create new Patch "b", leave it empty
4. Open "main" patch
CRASH! See console.


brusherru:
> When was it introduced?
When we added the possibility to load a project with dead refs, dead links and etc 😬
Which means he's been around for a while. Apparently, I the first who caught this bug :)

PR #1400, opened 2018-08-10
Improve performance of `xod-patch-search`

brusherru:
There is no issue.

It reduces search time:
- by ~9% for common searches (like "number", "meter" and etc): 108ms -> 99ms
- by ~54% for search by a library ("lib:xod/core add"): 68ms -> 30ms
- by ~ 91% for search specializations ("when-either-ch(mber"): 74ms -> 6.79ms
- unchanged for search abstracts: 74ms -> 74ms

Also, it exports two new methods: `setCollection` and `getCollection, that will be used in a further optimization.

### DIFF
--- a/packages/xod-patch-search/src/index.js
+++ b/packages/xod-patch-search/src/index.js
@@ -92,20 +92,41 @@ const refineScore = R.curry((query, result) =>
   )(result)
 );
 
+// :: [String] -> String -> Number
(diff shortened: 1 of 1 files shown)


brusherru:
@nkrkv @evgenykochetkov check it, please.
It's a small PR, moving search index in the hinting will be in another one.

PR #1404, opened 2018-08-15
Optimize patch search and suggester

brusherru:
There is no issue.

It optimizes the patch search engine by:
- move patch search data into `hinting` branch of the state and update it by middleware, instead of selectors
- update patch search data only on some actions (like `PATCH_DESCRIPTION_UPDATE` or `PROJECT_OPEN`).
- and update search data partially if it possible (updates only patch description, rename patch or add the new one — update only its search data).
- make `searchPatches` function pure and memoize it once
- Suggester component shows only 20 best search results
- Suggester searches when User types at least two symbols

In numbers:
- **Every change** on the patch will not call updating patch search index: ~53ms -> 0ms
- Little changes (description, rename, add): ~53ms -> ~2.5ms
- Change of the entire set of the patches (open/create a project): ~53ms -> ~23.5ms
- Suggester show first results faster on ~100ms

### DIFF
--- a/packages/xod-client/src/project/selectors.js
+++ b/packages/xod-client/src/project/selectors.js
@@ -3,7 +3,7 @@ import { createSelector } from 'reselect';
 
 import * as XP from 'xod-project';
 import { foldMaybe, maybeProp } from 'xod-func-tools';
-import { createIndexFromPatches } from 'xod-patch-search';
+import { createPatchSearcher } from 'xod-patch-search';
 
 import {
   addNodePositioning,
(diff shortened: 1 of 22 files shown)


packages/xod-client/src/project/selectors.js
```diff
@@ -412,34 +410,17 @@ export const getRenderableSelection = createMemoizedSelector(
 // Suggester
 //
 
-// :: State -> { search: (String -> [SearchResult])}
-export const getPatchSearchIndex = createMemoizedSelector(
-  [listPatches, getCurrentPatchPath],
-  [R.equals],
-  (patches, maybeCurPatchPath) =>
+// :: _ -> ([PatchSearchData] -> String -> [SearchResult])
```
nkrkv: It’s just

```js
// :: [PatchSearchData] -> String -> [SearchResult]
```

PR #1408, opened 2018-08-16
Fix error hinting on bus label change

brusherru:
There is no issue.


packages/xod-client/src/hinting/validation.js
```diff
@@ -127,7 +127,11 @@ const shallValidateFunctions = {
           () => key !== 'label',
           // But if it's terminal or self Node — validate
           R.compose(
-            R.either(XP.isTerminalPatchPath, XP.isTerminalSelf),
+            R.anyPass([
```
nkrkv: Update the comment right above as well.

PR #1412, opened 2018-08-17
Scroll patch with mouse wheel or trackpad

### LINKED ISSUE #1388: Pan patch on scroll events
### Rationale

When using a laptop with a dumb touchpad, panning the patchboard can be tricky as it requires some acrobatics from your fingers. In contrast to desktop+mouse laptops support touchpad gestures existing to solve such problems. Notably, scrolling usually might be done with two-finger swiping.

Also, as noted by users, many of them expect that the mouse wheel will scroll the patch board as well.

XOD ignores the scrolling events and that makes it feel non-native. We can handle the vertical and horizontal scrolling events generated by a touchpad or mouse to make UI more intuitive.

### Acceptance criteria

* [ ] I can pan using the two-finger swipe gesture on a touchpad configured to use this gesture for scrolling
* [ ] I can scroll the mouse wheel to pan the patch board vertically
* [ ] I can hold the Shift key and scroll the mouse wheel and the patch board will scroll horizontally

brusherru:
It will close #1388

### DIFF
--- a/packages/xod-client/src/editor/containers/Patch/index.jsx
+++ b/packages/xod-client/src/editor/containers/Patch/index.jsx
@@ -8,6 +8,7 @@ import { connect } from 'react-redux';
 import { bindActionCreators } from 'redux';
 import ReactResizeDetector from 'react-resize-detector';
 import { noop } from 'xod-func-tools';
+import normalizeWheel from 'normalize-wheel';
 
 import * as EditorActions from '../../actions';
 import * as ProjectActions from '../../../project/actions';
@@ -198,6 +199,7 @@ class Patch extends React.Component {
     const { currentMode } = this.state;
     const modeHandler = MODE_HANDLERS[currentMode];
 
+    const wheel = normalizeWheel(event);
     return R.compose(
       this.dispatchOffsetUpdate,
       R.tap(() =>
@@ -205,8 +207,8 @@ class Patch extends React.Component {
       ),
       R.tap(newOffset => this.setState({ offset: newOffset })),
       R.evolve({
-        x: R.subtract(R.__, event.deltaX),
-        y: R.subtract(R.__, event.deltaY),
+        x: R.subtract(R.__, wheel.pixelX),
+        y: R.subtract(R.__, wheel.pixelY),
       })
     )(this.state.offset);
   }
--- a/packages/xod-client/src/editor/containers/Patch/index.jsx
+++ b/packages/xod-client/src/editor/containers/Patch/index.jsx
@@ -77,6 +77,9 @@ class Patch extends React.Component {
       offset: this.props.offset,
     };
 
+    // Store is shift key was pressed on the first triggered Wheel event
+    this.shiftKey = false;
+
     // Storage for mode data without forcing update of component
     // E.G. store here refs on the components
     this.storage = {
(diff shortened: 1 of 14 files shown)


brusherru:
@nkrkv fixed these issues:
- invert scrolling
- shift key for horizontal scrolling (Windows has the same behavior as a Linux, so I made an exception only for `macosx`)
- Fix twitching of Patch on doing something before `PATCH_UPDATE_OFFSET` was dispatched
- normalized scroll speed for all browsers/platforms

I did all these things in separate commits, you can check it out separately. Then I'll squash it all in the one commit.

packages/xod-client/src/editor/containers/Patch/index.jsx
```diff
@@ -196,18 +197,31 @@ class Patch extends React.Component {
 
     const { currentMode } = this.state;
     const modeHandler = MODE_HANDLERS[currentMode];
+    const normalizedWheel = normalizeWheel(event);
+
+    // Most OS does not detect horizontal scroll by wheel event by default.
+    // Only MacOS works fine. So we have to do a hack for proper horizontal
+    // scrolling with pressed Shift key on all other OS: just switch X and Y.
+    const wheel =
+      detector.os.name !== 'macosx' && event.shiftKey
```
nkrkv: This one feels fragile nevertheless. Can we go on without relying on OS name? For example, the following predicate looks pretty equivalent to me:

```js
event.shiftKey && normalizedWheel.pixelX === 0
```

In contrast to the current conditional, this will trigger a swap on macOS when a user hold down the Shift key and somehow scrolled strictly horizontally?! Is it ever possible? If so, what’s in the head of this xoder who holds the Shift key while scrolling with a gesture?

brusherru:
@nkrkv check it out, please

nkrkv:
Tested it on ThinkPad with a touchpad and trackpoint. Both work fine :+1: 

Let’s wait for feedback from @joepie91 if he has time for testing until the end of this week. If not, let’s merge and open a new issue if something is wrong.

joepie91:
I've been working with it a bit on my desktop today (ie. with a regular mouse), and there it worked fine. I'll probably have time to test it on my laptop tomorrow, and I'll report back then.

I do keep running into weird intermittent scrolling issues with the project explorer and inspector; I'm up to three *different* mystery issues now. The specific issues I've been running into:

1. No scrolling at all in the Inspector pane (#1423)
2. When scrolling rapidly and nearing the end of the content in the project explorer, it suddenly jumps to the end, skipping a part of the content. Same for scrolling towards the start.
3. When clicking an item when the content exceeds the project browser pane size, the pane jumps to a different scroll position.

Annoyingly, none of these issues are reliably reproducible, and I have no idea whether they're a consequence of the patch panning or a totally unrelated issue (eg. a wonky pane scrolling implementation). I don't think they should be a blocker, unless you have a reason to believe that it might be related to patch panning.

brusherru:
@joepie91 thanks for the reports and test on the desktop. Will you test it on your Thinkpad?

About project browser and inspector — we'll inspect it and try to fix with another pull request. Looks like the problem is in the "custom scroll" component 🤔

nkrkv:
@brusherru I think we should rebase & merge this PR. In case of any flaws, a new issue might be open.

joepie91:
Apologies for the delay :)

I've tested it on my laptop for a bit, and the scrolling works fine. It does get a bit laggy/jittery when I scroll quickly; I'm not sure whether that's due to the scrolling code, or whether it's just a general performance issue. Looking at my process manager, it starts eating a core when that happens.

joepie91:
Hmm, another issue: if you pan too far away from the contents, it can be difficult to find the contents again, given that it's an infinite canvas and there doesn't seem to be a 'return to center' button or anything. I guess that's a broader panning UX issue though, and not specific to this PR.

nkrkv:
Thanks for testing!

To find the origin press HOME or Ctrl+HOME. These commands are also available via the View menu.

PR #1421, opened 2018-08-20
Center tab contents (label and cross icon)

### LINKED ISSUE #1416: Tab contents are not centered vertically
Observable in 0.23.0, but seems the flaw is out there for months already

Look at the tab titles and close buttons in tabs:

screenshot from 2018-08-20 12-56-38

Their baseline is aligned to the tabline bottom, however in most UI’s the content should be vertically-centered to the tabline height. Let’s fix it to be a bit more intuitive.

brusherru:
It fixes #1416.


brusherru:
Now it should work fine.
Tested on MacOS Chrome, FF and on Ubuntu (in VM) Chrome and FF.

PR #1430, opened 2018-09-04
Optimize sortGraph

evgenykochetkov:
Massively improves toposorting time for larger patches.
For example, for a patch with 449 nodes connected by 561 links it went down from 1460165ms to 1429ms.

### DIFF
--- a/packages/xod-project/src/gmath.js
+++ b/packages/xod-project/src/gmath.js
@@ -6,14 +6,29 @@ export function findVertexesWithNoIncomingEdges(vertexes, edges) {
   return R.difference(vertexes, R.map(R.nth(1), edges));
 }
 
-export function hasIncomingEdges(vertex, edges) {
-  const edgeIncoming = R.compose(R.equals(vertex), R.nth(1));
-  return R.any(edgeIncoming, edges);
-}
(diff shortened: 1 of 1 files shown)


packages/xod-project/src/gmath.js
```diff
@@ -6,14 +6,29 @@ export function findVertexesWithNoIncomingEdges(vertexes, edges) {
   return R.difference(vertexes, R.map(R.nth(1), edges));
 }
 
-export function hasIncomingEdges(vertex, edges) {
-  const edgeIncoming = R.compose(R.equals(vertex), R.nth(1));
-  return R.any(edgeIncoming, edges);
-}
+// Packs edges list in a data structure optimized for use in `sortGraph`
+//
+// :: [[Int, Int]] -> EdgesSet
```
brusherru: It will be nice to write somewhere that `EdgeSet` is `StrMap StrMap Boolean` :)

evgenykochetkov: @brusherru it's a implementation detail that I'd like to keep opaque.

@nkrkv I think access time won't improve by much, but `dissocEdge` could become a bit faster. Also, implementation will get a little bit more ugly😁 I'll try it.

PR #1436, opened 2018-09-07
Add library `xod/datetime`

gabbapeople:
The XOD library that provides date and time operations.


workspace/__lib__/xod/datetime/unpack-datetime/patch.cpp
```diff
@@ -0,0 +1,57 @@
+#define LEAP_YEAR(Y) (((1970 + (Y)) > 0) && !((1970 + (Y)) % 4) && (((1970 + (Y)) % 100) || !((1970 + (Y)) % 400)))
```
nkrkv: Ditto

workspace/__lib__/xod/datetime/pad-with-zeroes/patch.cpp
```diff
@@ -0,0 +1,33 @@
...
+
+    char strNum[16];
+    dtostrf(num, 0, 0, strNum);
+
+    CStringView numView(strNum);
+    uint8_t lenStr = length(XString(&numView));
+
+    if (lenStr >= lenFull) {
+        strcpy(state->str, strNum);
+    } else {
```
nkrkv: I think the `if` clause is no longer required at all. The `else` branch will do the job for any case. Please, confirm.

workspace/__lib__/xod/datetime/project.xod
```diff
@@ -0,0 +1,3 @@
+{
+  "name": "Date and timestamp operations"
```
nkrkv: Bad metadata

workspace/__lib__/xod/datetime/unpack-datetime/patch.cpp
```diff
@@ -0,0 +1,60 @@
...
+    uint8_t month = 0, monthLength = 0;
+    days = 0;
+
+    for (month = 0; month < 12; month++) {
+        if (month == 1) { // february
+            if (isLeapYear(year)) {
+                monthLength = 29;
+            } else {
+                monthLength = 28;
+            }
```
nkrkv: Simpler expressed with:

```cpp
monthLength = isLeapYear(year) ? 29 : 28;
```

…and you’re calling `isLeapYear` on each iteration which is suboptimal. Suggest computing it once the `year` value settles down.

:open_mouth: And you’re using `uint8_t` to hold `year`. Does it even work while handling 2000-ths?

gabbapeople:
Also removed auto-generated type input and output.

gabbapeople:
@evgenykochetkov @brusherru
Guys pls check it, or prefer some test if you want.

brusherru:
**Some bugs detected:**
1. `datetime` with all inputs equals to `zero` passed through `format-timestamp` returns "2047-12-31 -0:-0:-0"
2. As you seen above, any "0" in minutes/hours is formats as "-0"
3. I tried to calculate how much days ago I was born using two datetimes and `diff-seconds`... It looks like I'm 107 years old 🤔 
4. I place two `datetime` nodes: "Now" and "Now + 2 min" with corresponding values and link them with `diff-seconds` and got `4294967300.00` as a result, but expected to get 120 seconds.
5. `datetime` with value "2018-09-10 15:30:0" linked to `add-seconds` with binded "120" seconds returns "2018-09-10 15:32:16"... Looks like there are 16 additional seconds 🤔 

P.S.
I think it's time to start writing tabtests :)

evgenykochetkov:
Also can confirm "-0" stuff.
`diff-seconds` works ok when first value is bigger then a second one. Otherwise it returns weird stuff like "107 years"

gabbapeople:
@brusherru Ok, relating to your comments:
1) -2) the cause of `-0` values errors was in `abs(0)` function in the `pads-with zeroes` node. I also added proper results for possible situations ( `0` value, `0` `W` width;  `0` value, `<0` `W` width, e.t.c) 
3) 4) Fixed. Now it outputs the expected result for both link orders.
5) Fixed. the `add-seconds`  node first calculates the result time in int32_t type and then outputs Number.

gabbapeople:
@evgenykochetkov @brusherru 🙏 Ty!

gabbapeople:
@nkrkv please review

workspace/__lib__/xod/datetime/datetime/patch.cpp
```diff
@@ -0,0 +1,55 @@
...
+        }
+    }
+
+    bool isLeapFlag = isLeapYear(year);
+    for (i = 1; i < mon; i++) {
+        if ((i == 2) && isLeapFlag) {
+            seconds += SECS_PER_DAY * 29;
+        } else {
+            seconds += SECS_PER_DAY * monthDays[i - 1];
+        }
```
nkrkv: Can be further simplified :smiling_imp: 

```cpp
bool leap = isLeapYear(year);
constexpr uint8_t february = 2;
for (i = 1; i < mon; ++i) {
    seconds += SECS_PER_DAY * (i == february && leap) ? 29 : monthDays[i - 1];
}
```

workspace/__lib__/xod/datetime/unpack-datetime/patch.xodp
```diff
@@ -0,0 +1,91 @@
...
+      "id": "HJLWQZQPm",
+      "label": "WDAY",
+      "position": {
+        "x": 170,
+        "y": 204
+      },
+      "type": "xod/patch-nodes/output-number"
+    },
+    {
+      "description": "Year value in range [1970, 2038]",
```
nkrkv: Should it be 2105?

workspace/__lib__/xod/datetime/unpack-datetime/patch.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+    uint8_t month = 0, monthLength = 0;
+    days = 0;
+
+    for (month = 0; month < 12; month++) {
+        if (month == 1) {
+            monthLength = isLeapFlag ? 29 : 28;
+        } else {
+            monthLength = monthDays[month];
+        }
```
nkrkv: The same simplification is possible

workspace/__lib__/xod/datetime/unpack-datetime/patch.cpp
```diff
@@ -0,0 +1,59 @@
...
+    emitValue<output_HOUR>(ctx, time % 24);
+    time /= 24;
+
+    emitValue<output_WDAY>(ctx, ((time + 4) % 7) == 0 ? 7 : ((time + 4) % 7));
+
+    uint16_t year = 0;
+    uint32_t days = 0;
+
+    while ((unsigned)(days += (isLeapYear(year) ? 366 : 365)) <= time) {
+        year++;
```
nkrkv: This one is tricky. Better use do/while

```cpp
do {
    days += isLeapYear(year) ? 366 : 365;
    ++year;
} while (days <= time);
```

workspace/__lib__/xod/datetime/unpack-datetime/patch.cpp
```diff
@@ -0,0 +1,59 @@
...
+    static const uint8_t monthDays[] = { 31, 28, 31, 30, 31, 30, 31, 31, 30, 31, 30, 31 };
+
+    emitValue<output_SEC>(ctx, time % 60);
+    time /= 60;
+    emitValue<output_MIN>(ctx, time % 60);
+    time /= 60;
+    emitValue<output_HOUR>(ctx, time % 24);
+    time /= 24;
+
+    emitValue<output_WDAY>(ctx, ((time + 4) % 7) == 0 ? 7 : ((time + 4) % 7));
```
nkrkv: Can it be simplified a bit?

```cpp
uint8_t zwday = (time + 4) % 7;
emitValue<output_WDAY>(ctx, zwday ? zwday : 7);
```

workspace/__lib__/xod/datetime/datetime/patch.cpp
gabbapeople: Without () it has another meaning, so
`seconds += SECS_PER_DAY * ((i == 2 && leap) ? 29 : monthDays[i - 1]);`

workspace/__lib__/xod/datetime/unpack-datetime/patch.cpp
gabbapeople: These circles have different results

gabbapeople: `    uint32_t days = isLeapYear(year) ? 366 : 365;

    do {
        days += isLeapYear(year) ? 366 : 365;
        ++year;
    } while (days <= time);
` 
It works fine like this. What variant is better?

gabbapeople: @nkrkv 
I think the 
```cpp
 while ((days += (isLeapYear(year) ? 366 : 365)) <= time) {
        year++;
    }
``` 
is shortest and the order of both incrementions is understandable

workspace/__lib__/xod/datetime/pad-with-zeroes/patch.cpp
```diff
@@ -0,0 +1,30 @@
...
+    auto state = getState(ctx);
+
+    auto num = (getValue<input_IN>(ctx) == 0) ? 0 : abs(getValue<input_IN>(ctx));
+    int8_t lenFull = (getValue<input_W>(ctx) < 0) ? 0 : min(15, getValue<input_W>(ctx));
+
+    char strNum[16];
+    dtostrf(num, 0, 0, strNum);
+
+    CStringView numView(strNum);
+    int8_t lenStr = length(XString(&numView));
```
nkrkv: :open_mouth: this can be really made more simple:

```cpp
int8_t lenStr = strlen(strNum);
```

No `CStringView`/`XString` magic is required

workspace/__lib__/xod/datetime/pad-with-zeroes/patch.cpp
```diff
@@ -0,0 +1,30 @@
...
+
+    size_t zeroCount = ((lenFull - lenStr) < 0) ? 0 : max(0, lenFull - lenStr);
```
nkrkv: Why the condition? `max` will guarantee a result no less than zero

gabbapeople:
@nkrkv Updated

workspace/__lib__/xod/datetime/unpack-datetime/patch.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+    uint8_t zwday = (time + 4) % 7;
+    emitValue<output_WDAY>(ctx, zwday ? zwday : 7);
+
+    uint16_t year = 0;
+    uint32_t days = 0;
+
+    while ((days += (isLeapYear(year) ? 366 : 365)) <= time) {
+        year++;
+    }
```
nkrkv: Can you address the complexity issue?

PR #1440, opened 2018-09-11
Fix screenshoter script for docs to run properly on MacOS

brusherru:
`readlink` haven't flag `-f` on MacOS.
And it is not needed unless we run it through a symlink.
So I just replaced it with `$0`.

### DIFF
--- a/tools/update-docs-screenshots.sh
+++ b/tools/update-docs-screenshots.sh
@@ -1,5 +1,6 @@
 #!/bin/sh
 
-export SHOT="$(dirname "$(readlink -f "$0")")/screenshot-xodball"
+SHOT="$(cd "$(dirname "$0")"; pwd)/screenshot-xodball"
+export SHOT
 find . -iname "update-screenshots.sh" \
   -execdir sh -c 'pwd && ./update-screenshots.sh' \;


knopki:
@nkrkv First I think `readlink` is not needed, then I test paths with spaces :)

@brusherru I commit some new version. Can you test?

I change `readlink` to `realpath` and change export as proposed in [SC2155](https://github.com/koalaman/shellcheck/wiki/SC2155)

brusherru:
@knopki there are no `realpath` in MacOS by default 😬

brusherru:
It works! Thanks!

tools/update-docs-screenshots.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/sh
 
-export SHOT="$(dirname "$(readlink -f "$0")")/screenshot-xodball"
+export SHOT="$(cd "$(dirname "$0")"; pwd)/screenshot-xodball"
```
nkrkv: OK, but can we have a little comment describing what the hell is going here?

knopki: Relative path to `tools/update-docs-screenshots.sh` ($0) -> relative path of directory `tools` -> absolute path to directory `tools` -> absolute path to `tools/screenshot-xodball`.

knopki: Change to:
```bash
SHOT="$(cd "$(dirname "$0")"; pwd)/screenshot-xodball"
export SHOT
```
As proposed ad https://github.com/koalaman/shellcheck/wiki/SC2155

nkrkv: A good explanation for the comment I’m asking for :smiley:

PR #1447, opened 2018-09-14
Make xod packages usable from npm again

evgenykochetkov:
- rename "holes" package to "belt-holes", because the name "holes" is already taken on npm
- fix publishing of files generated by bucklescript toolchains


.gitignore
```diff
@@ -5,7 +5,6 @@
 
 node_modules/
 packages/*/dist/
-packages/*/lib/
```
nkrkv: Why remove? Conventionally `lib` in a JS project contains build artifacts. IMO they should not be kept in Git.

nkrkv: Furthermore, suggest adding `*.bs.js` here to not duplicate it many times in local `.gitignore`’s

evgenykochetkov: Yeah, I agree. That was a silly thing to do. Moved it back to "global" `.gitignore`

PR #1456, opened 2018-09-27
ESP8266 compatibility

### LINKED ISSUE #1454: Make C++ runtime compatible with ESP8266
### Rationale

In the light of the upcoming universal toolchain support, it is required that XOD C++ runtime is compatible with the minor differences in cores’ implementation between the AVR platform and ESP8266.

Some adjustments might be required to make the transpiled code compile successfully for both platforms.

### Acceptance criteria

The following examples compile and work for an ESP8266 board:

- [ ] Basic LED blink
- [ ] An I²C source + I²C display
- [ ] UART communication with an Arduino Uno board
- [ ] Requesting own external IP from XOD cloud using the built-in Wi-Fi capabilities

evgenykochetkov:
- Makes sure C++ runtime is compatible with ESP8266
- Adds `xod-dev/esp8266-mcu` library for using "native" ESP8266 features(for now, it only supports built-in Wi-Fi capabilities and the API mimics existing `xod-dev/esp8266` and `xod-dev/w5500` libraries)

Closes #1454

### DIFF
--- a/workspace/__lib__/xod-dev/esp8266-mcu/esp8266-mcu-inet/patch.cpp
+++ b/workspace/__lib__/xod-dev/esp8266-mcu/esp8266-mcu-inet/patch.cpp
@@ -1,0 +1,19 @@
+
+// clang-format off
+{{#global}}
+#include <ESP8266WiFi.h>
+{{/global}}
+// clang-format on
+
+struct State {
+};
+
+using Type = ESP8266WiFiClass*;
+
+// clang-format off
+{{ GENERATED_CODE }}
+// clang-format on
+
+void evaluate(Context ctx) {
+    emitValue<output_OUT>(ctx, &WiFi);
+}
--- a/workspace/__lib__/xod-dev/esp8266-mcu/send/patch.cpp
+++ b/workspace/__lib__/xod-dev/esp8266-mcu/send/patch.cpp
@@ -1,0 +1,29 @@
+
+struct State {
+};
+
+// clang-format off
+{{ GENERATED_CODE }}
+// clang-format on
+
+void evaluate(Context ctx) {
+    if (!isInputDirty<input_SND>(ctx))
+        return;
+
+    auto client = getValue<input_SOCK>(ctx);
+    auto msg = getValue<input_MSG>(ctx);
+
+    size_t lastWriteSize;
+
+    for (auto it = msg.iterate(); it; ++it) {
+        lastWriteSize = client->write((char)*it);
+        if (lastWriteSize == 0) {
+            emitValue<output_ERR>(ctx, 1);
+            return;
+        }
+    }
+
+    client->flush();
+
+    emitValue<output_DONE>(ctx, 1);
+}
--- a/workspace/__lib__/xod-dev/esp8266-mcu/socket/patch.cpp
+++ b/workspace/__lib__/xod-dev/esp8266-mcu/socket/patch.cpp
@@ -1,0 +1,31 @@
+
+// clang-format off
+{{#global}}
+#include <ESP8266WiFi.h>
+{{/global}}
+// clang-format on
+
+using Type = WiFiClient*;
+
+struct State {
+    Type client;
+};
+
+// clang-format off
+{{ GENERATED_CODE }}
+// clang-format on
+
+void evaluate(Context ctx) {
+    auto state = getState(ctx);
+
+    if (isSettingUp()) {
+        // Normally, WiFiClient would be constructed on a
+        // pre-allocated buffer using a placement new operator.
+        // (xod-dev/esp8266/esp8266-device is a nice example)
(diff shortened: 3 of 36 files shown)


workspace/__lib__/xod-dev/esp8266-mcu/send/patch.cpp
```diff
@@ -0,0 +1,27 @@
...
+void evaluate(Context ctx) {
+    if (!isInputDirty<input_SND>(ctx))
+        return;
+
+    auto client = getValue<input_CLNT>(ctx);
+    auto msg = getValue<input_MSG>(ctx);
+
+    size_t lastWriteSize;
+
+    for (auto it = msg->iterate(); it; ++it) {
```
brusherru: Produces warning:
`warning: 'xod::List<T>* xod::List<T>::operator->() [with T = char]' is deprecated`
So just replace with `msg.iterate()`

workspace/__lib__/xod-dev/esp8266-mcu/esp8266-mcu-inet/patch.cpp
```diff
@@ -0,0 +1,17 @@
...
+// clang-format on
+
+struct State {
+};
+
+using Type = ESP8266WiFiClass*;
+
+{{ GENERATED_CODE }}
+
+void evaluate(Context ctx) {
```
brusherru: `warning: unused parameter 'ctx'`

workspace/__lib__/xod-dev/esp8266-mcu/tcp-client/patch.cpp
```diff
@@ -0,0 +1,17 @@
...
+// clang-format on
+
+using Type = WiFiClient*;
+
+struct State {
+};
+
+{{ GENERATED_CODE }}
+
+void evaluate(Context ctx) {
```
brusherru: `warning: unused parameter 'ctx'`

workspace/__lib__/xod-dev/esp8266-mcu/send/patch.cpp
evgenykochetkov: Thanks! Fixed.

workspace/__lib__/xod-dev/esp8266-mcu/esp8266-mcu-inet/patch.cpp
evgenykochetkov: Removed `evaluate` entirely, so if someone will use this node directly the error will be produced at compile time. @nkrkv, do you think it is ok?

evgenykochetkov:
Things left to do:
- [ ] rename `tcp-client` node to `socket`, `CLNT` pins to `SOCK`
- [ ] instead of missing `evaluate`s provide meaningful defaults for `esp8266-mcu-inet` and `socket`
- [ ] add null checks to nodes that deal with `socket`

evgenykochetkov:
@nkrkv @brusherru please check out last two commits

workspace/__lib__/xod/gpio/pwm-write/patch.cpp
```diff
@@ -3,6 +3,12 @@ struct State {
 
 {{ GENERATED_CODE }}
 
+#ifdef PWMRANGE
+const Number pwmRange = PWMRANGE;
+#else
+const Number pwmRange = 255.0;
```
nkrkv: Suggest using `constexpr` instead of `const` to guarantee zero RAM cost

workspace/__lib__/xod-dev/esp8266-mcu/socket/patch.cpp
```diff
@@ -0,0 +1,26 @@
...
+
+// clang-format off
+{{ GENERATED_CODE }}
+// clang-format on
+
+void evaluate(Context ctx) {
+    auto state = getState(ctx);
+
+    if (isSettingUp()) {
+        state->client = new WiFiClient();
```
nkrkv: Can we use placement new instead? It will make RAM consumption report more accurate. That is, the `State` might only contain `uint8_t mem[sizeof(WiFiClient)]`.

evgenykochetkov: The way it is used now, `evaluate` will only be called if someone forgot to link a "proper" `socket`(created with `open-tcp`). So if we use placement new, `mem` will almost always be unused.

nkrkv: Ah… that’s a “bad” client, got it. OK.

nkrkv: Can you leave a comment in the code explaining that should not normally happen?

PR #1469, opened 2018-10-15
Show proper message about missing arduino-cli

brusherru:


packages/xod-client-electron/src/arduinoDependencies/messages.js
```diff
@@ -38,4 +38,10 @@ export default {
     solution: 'Now you are able to upload the program',
     persistent: true,
   }),
+  ARDUINO_CLI_NOT_FOUND: ({ path }) => ({
+    title: 'Arduino-cli not found',
```
nkrkv: `arduino-cli` is a name in this case, so start with lowercase

packages/xod-client-electron/src/arduinoDependencies/messages.js
```diff
@@ -38,4 +38,10 @@ export default {
...
+    note: `Tried to find arduino-cli by the path: ${path}`,
```
nkrkv: :scroll: ~by~ on

As far as I see, there are chances the message will be quite meaningless: “Tried to find arduino-cli by the path: arduino-cli”

packages/xod-client-electron/src/arduinoDependencies/messages.js
```diff
@@ -38,4 +38,10 @@ export default {
...
+    solution:
+      'If you run XOD IDE in the DEV mode you have to set the path to the arduino-cli in your $PATH or in the env variable "ARDUINO_CLI"',
```
nkrkv: We can pass `isDev` as an argument and adjust the phrase accordingly.

```
isDev
  ? 'When running in the development mode `arduino-cli` should be available on $PATH or explicitly set with $XOD_ARDUINO_CLI environment variable'
  : 'This is a bug, report it to XOD developers'
```

Yes, I overlooked it in the original comment, but we should scope our environment variables with `XOD_` prefix.

packages/xod-client-electron/src/app/arduinoCli.js
```diff
@@ -241,6 +241,10 @@ export const create = sketchDir =>
     await migrateArduinoPackages();
     await ensureExtraTxt();
 
+    if (!await fse.pathExists(arduinoCliPath)) {
+      throw createError('ARDUINO_CLI_NOT_FOUND', { path: arduinoCliPath });
```
nkrkv: Will it ever be executed if the following branch takes place?

```js
const getArduinoCliPath = () =>
  new Promise((resolve, reject) => {
    const arduinoCliBin =
      os.platform() === 'win32' ? 'arduino-cli.exe' : 'arduino-cli';

    //...

   which(arduinoCliBin, (err, cliPath) => {
      if (err) {
        reject(err);
        return;
      }
      // ...
    });
  });

```

PR #1474, opened 2018-10-16
Provide default name for xodball saved from browser IDE

evgenykochetkov:
A very small tweak.

The previous behavior caused some pain when creating quick projects in browser IDE, because it would save it as just `.xodball`, which was not correctly  associated with XOD IDE.


evgenykochetkov:
Changed it to `my-project` 👍🏻

PR #1482, opened 2018-10-18
Do not rely on NUM_DIGITAL_PINS being defined

### LINKED ISSUE #1477: Do not rely on NUM_DIGITAL_PINS being defined
See a [post on the forum](https://forum.xod.io/t/stm32f103c-blue-pill-port-literals/1239). Some boards/variants do not define the `NUM_DIGITAL_PINS` macro. Thereafter, compilation fails for some core nodes on these boards. Namely:

```bash
$ rg NUM_DIGITAL_PINS 
workspace/blink/__fixtures__/arduino.cpp
1238:    bool err = (port > NUM_DIGITAL_PINS - 1);

workspace/__lib__/xod/common-hardware/servo/patch.cpp
19:    if (port > NUM_DIGITAL_PINS - 1) {

workspace/__lib__/xod/gpio/digital-read/patch.cpp
11:    bool err = (port > NUM_DIGITAL_PINS - 1);

workspace/__lib__/xod/gpio/digital-write/patch.cpp
11:    bool err = (port > NUM_DIGITAL_PINS - 1);

workspace/__lib__/xod/gpio/pwm-write/patch.cpp
17:    bool err = (port > NUM_DIGITAL_PINS - 1);
```

### How to fix

We can introduce a new C++ API function `isValidDigitalPort`:

```cpp
bool isValidDigitalPort(uint8_t port) {
#ifdef NUM_DIGITAL_PINS
    return port > NUM_DIGITAL_PINS - 1;
#else
    return true;
#endif
}
```

and use it in the mentioned places.

evgenykochetkov:
Fixes #1477

### DIFF
--- a/packages/xod-arduino/platform/runtime.cpp
+++ b/packages/xod-arduino/platform/runtime.cpp
@@ -198,6 +198,14 @@ bool isTimedOut(const ContextT* ctx) {
     return detail::isTimedOut(ctx->_node);
 }
 
+bool isValidDigitalPort(uint8_t port) {
+#ifdef NUM_DIGITAL_PINS
+    return port < NUM_DIGITAL_PINS;
+#else
+    return true;
+#endif
+}
+
 } // namespace xod
 
 //----------------------------------------------------------------------------
--- a/packages/xod-arduino/platform/runtime.cpp
+++ b/packages/xod-arduino/platform/runtime.cpp
@@ -206,6 +206,14 @@ bool isValidDigitalPort(uint8_t port) {
 #endif
 }
 
+bool isValidAnalogPort(uint8_t port) {
+#ifdef NUM_ANALOG_INPUTS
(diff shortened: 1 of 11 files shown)


evgenykochetkov:
Not sure about placement of `isValidDigitalPort`, suggestions are welcome

brusherru:
How about adding the same function for analog pins?
For example, this guy with "Blue pill" on the forum could have the same error if he will try to use `analog-read` node, and I'm sure in this package `NUM_ANALOG_INPUTS` is not defined too...

So, let's add
```
bool isValidAnalogPort(uint8_t port) {
#ifdef NUM_ANALOG_INPUTS
    return (port >= A0 || port < A0 + NUM_ANALOG_INPUTS - 1);
#else
    return true;
#endif
}
```

And tweak `xod/gpio/analog-read` with this function.

brusherru:
Oh, yes. My bad.

PR #1493, opened 2018-10-22
Sharp Infrared Range Meters Library

gabbapeople:
PR refers to the [XEN-L005](https://github.com/xodio/xod-xens/tree/master/XEN-L005)

- Transfers range meter nodes for Sharp devices to the new `xod-dev/sharp-irm` library;
- Deprecates the previous Sharp nodes at the `xod/common-hardware` library.
- For new nodes, adds the analog reference voltage adjustment.


gabbapeople:
Nodes are tested:
for 5v tested on Arduino Uno board.
for 3.3v tested on Amperka WiFi Slot.

PR #1494, opened 2018-10-22
Fix log position of compile/upload tabs on first show

brusherru:


brusherru:
@nkrkv fixed. Check it out, please!

packages/xod-client/src/utils/components/Autoscroll.jsx
```diff
@@ -0,0 +1,98 @@
+/*
+  Edited fork of `autoscroll-react` from
+  https://github.com/xodio/autoscroll-react.git#cbb3422461a28b40ca082d6a5f33f667508c3288
```
nkrkv: Nope. The original should be the original original. And a note on what’s been changed:

- onScrollFromBottom
- scrollDown

PR #1496, opened 2018-10-22
Make xod/core/clock tick even if EN is always dirty

evgenykochetkov:
In patches like this
screen shot 2018-10-22 at 20 19 27
`clock` would never start because `EN` is always dirty.

[clock-err.xodball.zip](https://github.com/xodio/xod/files/2502427/clock-err.xodball.zip)


evgenykochetkov:
Tweaked the implementation so that original behavior of RST is preserved

PR #1512, opened 2018-10-29
Make sure generated C++ is compatible with STM32 boards

evgenykochetkov:
- Add ability to define PORTs as `PA13`, `PB8` etc
- Don't redefine `dtostrf`, [the built-in one](https://github.com/stm32duino/Arduino_Core_STM32/blob/master/cores/arduino/avr/dtostrf.c) is fine

### DIFF
--- a/packages/xod-client/src/utils/normalizePort.js
+++ b/packages/xod-client/src/utils/normalizePort.js
@@ -1,12 +1,16 @@
 import * as R from 'ramda';
-import { DEFAULT_VALUE_OF_TYPE, PIN_TYPE } from 'xod-project';
+import {
+  DEFAULT_VALUE_OF_TYPE,
+  PIN_TYPE,
+  isValidPortLiteral,
+} from 'xod-project';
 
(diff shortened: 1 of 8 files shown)


packages/xod-client/src/utils/normalizePort.js
```diff
@@ -6,7 +6,7 @@ const LEADING_CHAR = 'D';
 // :: DataValue -> DataValue
 export default input =>
   R.unless(
-    R.test(/^(A|D)\d+/i),
+    R.test(/^(P[A-F]|A|D)\d+$/),
```
nkrkv: I worry a bit about the regex. It is complex and _almost_ matches the one used in `packages/xod-project/src/utils.js:220`: `R.test(/^(P[A-F]|A|D)\d{0,3}$/g, literal)`. I suggest extracting the test to a function like `isValidNumberDataValue` we have already.

evgenykochetkov: Done

PR #1524, opened 2018-11-01
xod/bits/shift-* works with int32_t

### LINKED ISSUE #1510: shift-left and shift-right should work with 32-bit numbers
[Reported on the forum](https://forum.xod.io/t/xod-0-25-1-issues/1287). The problem was always there but pointed just recently.

The correct behavior for the nodes should be:

### shift-left

Performs a logical left shift of `IN` by `N` bits.

| Dir | Pin | Type   | Description |
|-----|-----|--------|-------------|
| in  | IN  | number | Input number. Interpreted as a signed 32-bit interger |
| in  | N   | number | Shift bits count. Truncated to an integer. Negative values round to zero |
| out | OUT | number |             |

### shift-right

Performs a logical right shift of `IN` by `N` bits.

| Dir | Pin | Type   | Description |
|-----|-----|--------|-------------|
| in  | IN  | number | Input number. Interpreted as a signed 32-bit interger |
| in  | N   | number | Shift bits count. Truncated to an integer. Negative values round to zero |
| out | OUT | number |             |

---

The C++ implementations should be adjusted accordingly to deal with `int32_t` rather than `uint8_t`.

A tabular test is much desired to enforce the behavior

brusherru:
It fixes #1510


workspace/__lib__/xod/bits/shift-left/patch.cpp
```diff
@@ -7,7 +7,15 @@ struct State {
...
+    int32_t x = (int32_t)getValue<input_IN>(ctx);
+    int32_t n = (int32_t)getValue<input_N>(ctx);
+    int32_t b = 0;
+    if (n < 0) {
+      b = 0;
+    } else if (n > 31) {
+      b = 31;
+    } else {
+      b = n;
+    }
```
evgenykochetkov: https://www.arduino.cc/reference/en/language/functions/math/constrain/ ?

brusherru: I did not use `constrain` to avoid bugs on some other platforms that did not define it.

PR #1531, opened 2018-11-08
Improve tabtests engine, cover `xod/math` with tabtests and fix/add some nodes

brusherru:
Improvements in `xod-tabtest`:
- support `NaN`, `Inf`, `+Inf`, `-Inf` literals in tests
- support approximate numbers in the tests (`3.1416~`). It makes possible to test some computations which return a float, like `sin`, `sqrt` and etc

Added nodes:
- `is-finite`
- `is-nan`

Fixed nodes:
- `discretize` — it did not work correctly if the ports `Tn` were not in ascending order and for NaN always returned a value from the first port. Now for NaN returns NaN.
- `map` — the same thing with NaN


workspace/__lib__/xod/math/is-finite/patch.xodp
```diff
@@ -0,0 +1,29 @@
+{
+  "description": "Determines if the given number has finite value i.e. it is normal, subnormal or zero, but not infinite or NaN",
```
nkrkv: :scroll: Determines if the given number has a finite value, i.e., it is normal, subnormal, or zero, but not infinite or NaN

workspace/__lib__/xod/math/pi/patch.test.tsv
```diff
@@ -0,0 +1,2 @@
+OUT
+3.14159265359
```
nkrkv: :rofl:

workspace/__lib__/xod/math/sin/patch.test.tsv
```diff
@@ -0,0 +1,11 @@
+RAD	OUT
+0	0
+1	0.841471~
+-1	-0.841471~
+180	-0.801153~
+270	-0.176046~
+360	0.958916~
+-520	0.997795~
```
nkrkv: What’s the point to test against 180, 270, 360 _radians_? I would expect values like π/2, π, 3π/2, 2π here

workspace/__lib__/xod/math/tan/patch.test.tsv
```diff
@@ -0,0 +1,12 @@
+RAD	OUT
+0	0
+30	-6.40533~
+-30	6.40533~
+45	1.61~
+90	-1.995~
+180	1.338~
+-180	-1.338~
```
nkrkv: Ditto. Unit mismatch

workspace/__lib__/xod/math/acos/patch.test.tsv
```diff
@@ -0,0 +1,8 @@
+IN	OUT
+0	0
+-1	-1.57
+1	1.57
```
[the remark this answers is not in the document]
nkrkv: If so, our CI is misconfigured at must be fixed. Because now I see it green.

packages/xod-tabtest/test/TabData_jest.re
```diff
@@ -64,6 +65,11 @@ describe("TSV parser", () => {
       |. Map.String.set("Byte", Byte(13))
       |. Map.String.set("String", String("World")),
       Map.String.empty
+      |. Map.String.set("Number", ApproxNumber(-1.243, 3))
+      |. Map.String.set("Boolean", Boolean(false))
+      |. Map.String.set("Byte", Byte(255))
+      |. Map.String.set("String", String("!")),
```
evgenykochetkov:
```
Map.String.fromArray([|
    ("Number", ApproxNumber(-1.243, 3)),
    ("Boolean", Boolean(false)),
    ("Byte", Byte(255)),
    ("String", String("!")),
|]);
```
could be a bit cleaner

PR #1536, opened 2018-11-09
xod-cli: tabtests in separate directories, clean before run

knopki:
Remove the output tabtest directory if `--clean` flag is set.
Why not do this by default? Because Pete can set output directory to ~.
Special feature for @brusherru


knopki:
@brusherru  I like your idea.
New implementation:
If `--output-dir` is not defined, create a project's subdirectory and empty it before tests.

PR #1544, opened 2018-11-14
Support XStrings in tabtests and make bulletproof `formatNumber`

brusherru:
Changes:
- Support XStrings
- Move XString formatter for Catch2 into a `cpplib` directory at the repo root
- Make possible to compare XStrings using `==` operator
- Remove `dtostrf` function and add a new one `dtoa`, which works fine on any platform (tested on Wifi_slot, STM32 Nucleo401RE, Arduino Leonardo, Arduino Due, MacOS)
- Fix `xod/core/pad-with-zeroes` node to work properly with NaN, Inf, OVF values
- Fix path to `xodc` executable (tool to update transpiled fixtures automatically was broken)
- Added tabtests for some nodes in the `xod/core`
- Replaced fragile `xod-deploy` functional tests of compiled code equality with checks of filesize (must be above 4000 bytes) and containing hex


cpplib/catch2utils/XStringFormat.inl
```diff
@@ -0,0 +1,32 @@
+template <typename T>
```
nkrkv: A short banner of the file purpose would be nice… along with a URL to the string maker (or how is it called) documentation page.

packages/xod-arduino/platform/dtoa.h
```diff
@@ -0,0 +1,150 @@
...
+
+static void strreverse(char* begin, char* end)
+{
+    char aux;
+    while (end > begin)
+        aux = *end, *end-- = *begin, *begin++ = aux;
+};
+
+size_t dtoa(xod::Number value, int prec, char* str)
+{
```
nkrkv: I think `dtoa` is not the best name. `d` stands for `double`, but `Number` is not equivalent to `double`. So, it really should be… why overcomplicate… `formatNumber`.

Also, I suggest enclosing the contents with `namespace xod { }` and header file guards. Prefixing `Number` with `xod::` will no longer be required. See `listFuncs.h` for the reference.

Also, would you pass the file through `clang-format`?

packages/xod-arduino/platform/dtoa.h
```diff
@@ -0,0 +1,150 @@
...
+    /* Hacky test for NaN
+	 * under -fast-math this won't work, but then you also won't
+	 * have correct nan values anyways.  The alternative is
+	 * to link with libmath (bad) or hack IEEE double bits (bad)
```
nkrkv: No longer hacky. The comment is wrong now and should be rewritten (or dropped)

packages/xod-arduino/platform/dtoa.h
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    if (isinf(value)) {
+        bool isNegate = value < 0;
+        strcpy(str, isNegate ? "-Inf" : "Inf");
+        return (size_t) isNegate ? 4 : 3;
+    }
+
+    /* if input is larger than thres_max, revert to exponential */
+    const xod::Number thres_max = (xod::Number)(0x7FFFFFFF);
```
nkrkv: Misguiding comment. Behavior has changed to output `OVF`

packages/xod-arduino/platform/types.h
```diff
@@ -0,0 +1,10 @@
+namespace xod {
```
nkrkv: A banner is missing

packages/xod-arduino/platform/types.h
```diff
@@ -0,0 +1,10 @@
...
+#if __SIZEOF_FLOAT__ == 4
+typedef float Number;
+#else
+typedef double Number;
+#endif
+typedef bool Logic;
+typedef unsigned long TimeMs;
+typedef uint8_t DirtyFlags;
+}
```
nkrkv: Should be:

```cpp
} // namespace xod
```

because not indented

packages/xod-arduino/src/templates.js
```diff
@@ -306,10 +308,12 @@ export const renderProject = def(
       preambleH,
       config,
       stlH,
+      typesH,
       listViewsH,
       memoryH,
       uartH,
       omitLocalIncludes(listFuncsH),
+      dtoaH,
```
nkrkv: rename?

packages/xod-arduino/test-cpp/dtoa.cpp
```diff
@@ -0,0 +1,64 @@
+
+// Uncomment for debug:
+// #include <iostream>
+#include "catch.hpp"
+#include "../platform/types.h"
+#include "../platform/listViews.h"
+#include "../platform/listFuncs.h"
+#include "../platform/dtoa.h"
+#include "XStringFormat.inl"
```
nkrkv: You’re looking for `XStringFormat.inl` elsewhere, should be:

```
#include <XStringFormat.inl>
```

And better put it before the local includes.

packages/xod-tabtest/package.json
```diff
@@ -3,7 +3,8 @@
   "version": "0.25.2",
   "scripts": {
     "refmt": "find src/ test/ -name '*.re*' -exec refmt --in-place {} +",
-    "build": "bsb -make-world",
+    "build:copy-catch2utils": "cpx \"../../cpplib/catch2utils/**\" \"./cpp/catch2utils\"",
```
nkrkv: Seems like copying `XStringFormat.inl` should not be here, but rather be delivered in the same way as `catch2.hpp` does.

Such patterns are error-prone and lead to surprises when you see behavior changes only after full rebuilds.

packages/xod-arduino/platform/formatNumber.h
```diff
@@ -0,0 +1,151 @@
+/**
+ * Provide `dtoa` cross-platform number to string converter function.
```
nkrkv:
```suggestion
 * Provide `formatNumber` cross-platform number to string converter function.
```

packages/xod-arduino/platform/formatNumber.h
```diff
@@ -0,0 +1,151 @@
...
+size_t formatNumber(Number value, int prec, char* str) {
+    if (isnan(value)) {
+        strcpy(str, "NaN");
+        return (size_t)3;
+    }
+
+    if (isinf(value)) {
+        bool isNegate = value < 0;
+        strcpy(str, isNegate ? "-Inf" : "Inf");
+        return (size_t)isNegate ? 4 : 3;
```
nkrkv: `isNegative`? Negate is a verb.

packages/xod-arduino/platform/formatNumber.h
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    /* if input is larger than thres_max return "OVF" */
+    const Number thres_max = (Number)(0x7FFFFFFF);
+
+    double diff = 0.0;
```
nkrkv: Should work over `Number`’s, not `double`

packages/xod-arduino/platform/formatNumber.h
```diff
@@ -0,0 +1,151 @@
...
+    /* we'll work in positive values and deal with the
+	   negative sign issue later */
+    int neg = 0;
+    if (value < 0) {
+        neg = 1;
+        value = -value;
+    }
+
+    int whole = (int)value;
+    double tmp = (value - whole) * powers_of_10[prec];
```
nkrkv: ditto

packages/xod-arduino/platform/formatNumber.h
```diff
@@ -0,0 +1,151 @@
+/**
```
nkrkv: Look at the resulting C++ for `blink.cpp`. I suggest prefixing any inlined source file with the following banner to be consistent:

```cpp
/*=============================================================================
 *
 *
 * Bla bla bla bla
 *
 *
 =============================================================================*/
```

Otherwise, the structure is lost a bit

packages/xod-arduino/platform/types.h
```diff
@@ -0,0 +1,13 @@
+//----------------------------------------------------------------------------
+// Types
+//----------------------------------------------------------------------------
```
nkrkv: A note about the banner, see nearby

packages/xod-arduino/tools/test-avr-size.sh
```diff
@@ -25,7 +25,7 @@ AVR Memory Usage
 ----------------
 Device: atmega328p
 
-Program:    1684 bytes (5.1% Full)
+Program:    1626 bytes (5.0% Full)
```
nkrkv: 60 bytes diet :+1:

cpplib/catch2utils/XStringFormat.inl
```diff
@@ -0,0 +1,43 @@
...
+ * {?} == {?}
+ *
+ * Catch2 string conversion docs:
+ * https://github.com/catchorg/Catch2/blob/master/docs/tostring.md
+ */
+
+template <typename T>
+struct XodStringMaker {
+    static std::string convert(T value) {
+        size_t len = length(value) + 2;
```
evgenykochetkov: Shouldn't this be `+ 3`? (extra one for leading '\0')?

brusherru:
@nkrkv @evgenykochetkov fixed!
Check it out, please!

PR #1548, opened 2018-11-15
New xod-cli commands and new package xod-deploy-bin

knopki:
Part of work for #884.
Created new package `xod-deploy-bin` from `xod-client-electron/src/app/arduinoCli.js`.
`xod-deploy-bin` used by `xod-client-electron` and `xod-cli`.
New commands in `xod-cli`: `boards`, `install:arch`, `compile`, `upload`.
Small bugfixes in `arduino-cli`.


packages/xod-cli/README.md
```diff
@@ -40,27 +40,29 @@ For example, instead of `--username` you can declare variable `XOD_USERNAME`.
...
 | --workspace  | -w    | XOD_WORKSPACE        |
 
 
 # Commands
 <!-- commands -->
 * [`xodc autocomplete [SHELL]`](#xodc-autocomplete-shell)
+* [`xodc boards [options]`](#xodc-boards-options)
+* [`xodc compile [options] [entrypoint]`](#xodc-compile-options-entrypoint)
 * [`xodc help [COMMAND]`](#xodc-help-command)
+* [`xodc install:arch [fqbn]`](#xodc-installarch-fqbn)
```
nkrkv: :thinking: can it be just `install`. Think of it as of a package manager: XOD lib, C++ lib, hardware package, does not matter much

packages/xod-cli/README.md
```diff
@@ -85,6 +87,68 @@ EXAMPLES
...
+```
+USAGE
+  $ xodc boards [options]
+
+OPTIONS
+  -V, --version         show CLI version
+  -h, --help            show CLI help
+  -q, --quiet           do not log messages other than errors
+  -w, --workspace=path  [default: ~/xod] use the workspace specified, defaults to $HOME/xod
+  --available           show available boards
```
nkrkv: `xodc boards`, as states above, “shows available boards.” What does this flag adds. Looks like with it we “show available available boards”.

Do you mean “show only boards connected to PC”?

packages/xod-cli/README.md
```diff
@@ -85,6 +87,68 @@ EXAMPLES
...
+         * ./path/to/proj.xodball main      # xodball + patch name
+         * ./path/to/proj/main/patch.xodp   # just full path to a patch
+         * main                             # a patch in the current project
+
+OPTIONS
+  -V, --version         show CLI version
+  -b, --board=fqbn      (required) target board identifier
+  -h, --help            show CLI help
+
+  -o, --output=path     save output binary at the path specified; could be a file path or directory; the same directory
```
nkrkv:
```suggestion
  -o, --output=path     save output binary at the path specified; can be a file path or directory; the same directory
```

nkrkv: “the same directory” is not clear to me

packages/xod-cli/README.md
```diff
@@ -85,6 +87,68 @@ EXAMPLES
...
+      be a XOD patch path. If either is omitted, it is inferred from the current
+      working directory or another argument. Examples:
+
+         * ./path/to/proj.xodball main      # xodball + patch name
+         * ./path/to/proj/main/patch.xodp   # just full path to a patch
+         * main                             # a patch in the current project
+
+OPTIONS
+  -V, --version         show CLI version
+  -b, --board=fqbn      (required) target board identifier
```
nkrkv:
```suggestion
  -b, --board=fqbn      (required) target board identifier (see `xodc boards` output)
```

packages/xod-cli/README.md
```diff
@@ -100,7 +164,27 @@ OPTIONS
...
+## `xodc install:arch [fqbn]`
+
+install toolchains
+
+```
+USAGE
+  $ xodc install:arch [fqbn]
+
+ARGUMENTS
+  FQBN  Board FQBN. `arduino:sam` for example.
```
nkrkv:
```suggestion
  FQBN  Board FQBN. `arduino:sam` for example. See `xodc boards` list for the full list.
```

packages/xod-cli/README.md
nkrkv: OK, after a talk, nevermind, leave `install:arch`

packages/xod-cli/README.md
knopki: Oops, overengineered. I removed these options. Just showing the table (as in the issue) with or without header (-q flag).

packages/xod-cli/README.md
knopki: The string is copied from issue :)
Actually, there is a bug. I always use this path as the directory path and copy here ino, elf and hex/bin. Do we really need the option to save only firmware here? If yes, what logic here? If the path exists and is a directory, then output all files to the directory, or if a path not exists/file, then save firmware to the path?

nkrkv: > The string is copied from issue :)

Regardless I made the issue the sentence is not clear to me :rofl:
To be honest, my bad, I missed the next line: “the same directory is used for intermediate build artifacts; defaults to `cwd`” is fine.


> Actually, there is a bug. I always use this path as the directory path and copy here ino, elf and hex/bin.

It is OK, I think. When I wrote about the path here I though about putting binaries for multiple platforms in a single directory. No longer actual. So, I suggest rewording to reflect the current state:

```
-o, --output=dir   save the result binary to the directory; the same directory is used
                   for intermediate build artifacts; defaults to `cwd`
```

packages/xod-cli/src/messages.js
```diff
@@ -6,6 +6,13 @@ export default {
       'The error has no formatter, which is a bug. Report the issue to XOD developers.',
   }),
 
+  ARDUINO_CLI_ERROR: ({ code, message }) => ({
+    title: `'arduino-cli' error, code ${code}`,
+    note: message,
+    solution:
+      '`arduino-cli` error. The original tool output message is above. If you believe it is a bug, report the problem to XOD developers.',
```
nkrkv: The `title` must not contain variables and should try to avoid punctuation. Solution should be a solution: not a doubt. So:

```js
title: 'Launch arduino-cli failed',
note: message,
solution: 'The error was not properly handled or formatted. This is a bug. Report it to XOD developers.'
```

Particular expected `arduino-cli` errors should be formatted accordingly: 'Compilation failed', 'Upload failed', etc (see IDE implementation)

brusherru:
There are scary things in the Build log in circle-ci on MacOS.
And it reproduces on my machine.

Also, it won't build at all while I'm using node v8.6.0. We're using node v.10+, so this is not necessary to support the older version of node. BUT it looks like these changes should not affect somehow on it. And probably we have a bug, that could bring some pain to us in the future 🤔

knopki:
```
xod-client-electron: WARNING in /var/home/sk/dev/xodio/xod/node_modules/child-process-promise/node_modules/cross-spawn/index.js
xod-client-electron: Module not found: Error: Can't resolve 'spawn-sync' in '/var/home/sk/dev/xodio/xod/node_modules/child-process-promise/node_modules/cross-spawn'
```
Warning issued by `webpack` because of the `spawn-sync` (dependency of the `cross-spawn` that is dependency of the `child-process-promise`) is not resolved. It's because `cross-spawn` doesn't list this dependency in `package.json`. But the version is not changed and I don't know why we don't have this warning before.
We can swap `child-process-promise` with `promisify-child-process` in `arduino-cli` and `xod-deploy-bin` but maybe it's not worth it (i'm scared of testing on windows).

knopki:
```
xod-client-electron: WARNING in /Users/distiller/project/node_modules/chokidar/lib/fsevents-handler.js
xod-client-electron: Module not found: Error: Can't resolve 'fsevents' in '/Users/distiller/project/node_modules/chokidar/lib'
```
Almost the same. Issued by `webpack` (why?) about optional dependency `fsevents` of `chokidar` of `cpx`.

knopki:
Also, we have many `node-gyp` errors on Macs, but we have them in `master` too.

knopki:
I need access to any Mac on Monday to visually understand why `puppeteer` test fails. :man_facepalming:

PR #1549, opened 2018-11-15
Prevent autoupdate on not supported platforms

knopki:
Proposed change: do not check for update on not supported platforms. This prevents exception and useless stack traces in the console.

### DIFF
--- a/packages/xod-client-electron/src/app/main.js
+++ b/packages/xod-client-electron/src/app/main.js
@@ -291,7 +291,11 @@ const onReady = () => {
       win.setTitle(newTitle);
     });
 
-    autoUpdater.checkForUpdates();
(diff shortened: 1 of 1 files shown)


packages/xod-client-electron/src/app/main.js
```diff
@@ -291,7 +291,9 @@ const onReady = () => {
       win.setTitle(newTitle);
     });
 
-    autoUpdater.checkForUpdates();
+    if (process.platform === 'win32' || process.platform === 'darwin') {
+      autoUpdater.checkForUpdates();
+    }
```
nkrkv: I would leave a comment here. Something like:

```
// On Linux XOD autoupdates are not supported. Use of OS package manager is encouraged there.
```

PR #1556, opened 2018-11-20
Improve undo/redo hotkeys reliability

### LINKED ISSUE #1530: Undo/Redo works instable on Windows/Linux
### Environment
- Scope: desktop IDE
- Version: 0.25.2
- OS: Windows 10 / Linux

Undo/redo works fine for moving nodes or adding new node from Project Browser, but when User adds a new node using `i` hotkey it makes undo/redo works strange (need to click hotkey few times).

Reported on forum: https://forum.xod.io/t/ctrl-z-ctrl-shift-z-no-works/1267

### Steps to reproduce
1. Open XOD IDE
2. Click `i` hotkey
3. Type "clock" and press "enter"
4. Move the node
5. Press Ctrl+Z

### Expected behavior
Node is moved back.

### Actual behavior
Nothing happened.
If click few more times it will do undo.
The same thing with "redo".

Also, if you click "undo" in menu item it will react instantly.

I did a small investigation, place `console.log` in accelerator function and in the Editor's command function. The last one was not called at all, accelerator function called only after few clicks on the hotkey.

evgenykochetkov:
Closes #1530

### DIFF
--- a/packages/xod-client-browser/package.json
+++ b/packages/xod-client-browser/package.json
@@ -18,7 +18,7 @@
     "react-dom": "^16.2",
     "react-event-listener": "^0.5.3",
     "react-fa": "^5.0.0",
-    "react-hotkeys": "^0.10.0",
+    "react-hotkeys": "^1.1.4",
     "react-redux": "^4.4.5",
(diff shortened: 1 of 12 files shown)


evgenykochetkov:
(makred as [WIP] because it should be merged after #1554)

PR #1557, opened 2018-11-20
Support pulses in tabtests

### LINKED ISSUE #1538: As tabtests author, I want to probe pulses so that I can cover stateful nodes
### Rationale

Currently, there is no way to test an effect which comes after sending a pulse to a node. Output pulses cannot be tested too. Pulses are the very basic feature and should be supported by the tabular test framework

### User story

Like I do for other data types, I can specify input and output values for pulse-type pins in a tabtest row. To do it, two literals exist: `pulse` and `no-pulse`.

Having the `pulse` value in an input cell means “the pulse was emitted in the transaction proceeding this row.” The `no-pulse` or an empty value mean no input pulse in this transaction.

For an output cell:

- `pulse` means the pulse was emitted in this transaction
- `no-pulse` means no pulse was emitted in this transaction
- an empty value means the pulse presence does not matter

### Acceptance criteria

- [x] I can use `pulse` and `no-pulse` literals as described above

brusherru:
It closes #1538


brusherru:
@nkrkv check it out, please!

workspace/__lib__/xod/core/delta/patch.xodp
```diff
@@ -10,17 +10,17 @@
...
+      "label": "REF",
       "position": {
         "x": 102,
         "y": 0
       },
       "type": "xod/patch-nodes/input-pulse"
     },
     {
-      "description": "Difference between the current `IN` value and the value it has when the previous `UPD` was send. Can be positive or negative. Always equals to 0 on the first update.",
+      "description": "Difference between the current `IN` value and the reference value which stored when `REF` received a pulse. Can be positive or negative. Always equals to 0 on the first update.",
```
nkrkv:
```suggestion
      "description": "The difference between `IN` and the reference value which was stored when `REF` received a pulse. Can be positive or negative. Always equals to 0 on the first update.",
```

brusherru:
@nkrkv done!

PR #1567, opened 2018-11-23
Support mocking time flow for tabtests

### LINKED ISSUE #1539: As tabtests author, I want to emulate time flow so that I can cover time-dependent nodes
### Rationale

Some nodes have an implicit dependency on time. They use `setTimeout`, `isTimedOut`, `millis`, `transactionTime`, or other nodes dependent on time flow. Such nodes should be testable.

### User story

When I create a tabular test I can add a special column which reflects the time flow. For the “pin name” a special token is used: `__time(ms)`. For the values, I use absolute timestamps expressed in milliseconds:

| __time(ms) | T   | SET   | RST | DONE     | ACT   |
|--------------|-----|-------|-----|----------|-------|
| 0            | 0.5 |       |     | no-pulse | False |
| 1000         |     | pulse |     | no-pulse | True  |
| 1499         |     |       |     | no-pulse | True  |
| 1500         |     |       |     | pulse    | False |

When the column is found, the framework start time flow simulation. For each transaction it stubs the following functions:

- `transactionTime`
- `millis`
- `micros`

To correctly handle `isTimedOut` nothing special might be added to the framework because having a proper `transactionTime` mock will automatically make `isTimedOut` work correctly. If I, as a tester, want to give my node a chance to evaluate multiple times between test cases, I can insert lines containing timespamps only:

| __time(ms) | T   | SET   | RST | DONE     | ACT   |
|--------------|-----|-------|-----|----------|-------|
| 0            | 0.5 |       |     | no-pulse | False |
| 1000         |     | pulse |     | no-pulse | True  |
| 1100         |     |       |     |  |   |
| 1300         |     |       |     |  |   |
| 1499         |     |       |     | no-pulse | True  |
| 1500         |     |       |     | pulse    | False |

### Acceptance criteria

- [ ] I can define time flow columns as described above
- [ ] `clock`, `delay`, `time-delta`, `fade`, `system-time` are covered with tabtests

brusherru:
It closes #1539.

Also, I reverted behavior of the `REF` port ( https://github.com/xodio/xod/commit/ec8674fd94a9ea550554e6373246751bbe3a3d61 ) of `xod/core/delta`, because it causes problems in `xod/core/delta-time`, `xod/core/pid-controller` and etc.
Please, pay attention to the changed description of `OUT` port.


packages/xod-tabtest/src/TabData.re
```diff
@@ -62,6 +63,12 @@ module Value = {
       byteStringToInt(byteString)
     | x => Invalid(x)
     };
+  let parseTime = str =>
+    switch (str) {
+    | numString when Re.test(numString, numberRegex) =>
```
nkrkv: I think using `numberRegex` is not correct. We should accept positive integers only

workspace/__lib__/xod/core/delta-time/patch.xodp
```diff
@@ -24,7 +24,7 @@
       }
     },
     {
-      "id": "SJVJkVcNM",
```
nkrkv: Why is this change necessary?

workspace/__lib__/xod/core/delta/patch.xodp
```diff
@@ -10,7 +10,7 @@
       "type": "xod/patch-nodes/not-implemented-in-xod"
     },
     {
-      "description": "Triggers memoization of the input value as the reference value.",
+      "description": "Triggers memoization of the input value as the reference value. ",
```
nkrkv: Dangling space

workspace/__lib__/xod/core/delta/patch.xodp
```diff
@@ -20,7 +20,7 @@
       "type": "xod/patch-nodes/input-pulse"
     },
     {
-      "description": "The difference between `IN` and the reference value which was stored when `REF` received a pulse. Can be positive or negative. Always equals to 0 on the first update.",
+      "description": "The difference between `IN` and the reference value which was stored by a pulse into `REF` port on the previous transaction. Can be positive or negative. Always equals to 0 on the first update.",
```
nkrkv:
```suggestion
      "description": "The difference between `IN` and the reference value which was stored by a `REF` pulse last time. Can be positive or negative. Always equals to 0 on the first update.",
```

workspace/__lib__/xod/core/delta-time/patch.xodp
brusherru: `delta-time` uses `system-time` and `delta` inside.
To make `RST` works properly we have to pulse `system-time` on `RST`.

brusherru: It's a links' ID ;)

packages/xod-tabtest/src/TabData.re
brusherru: Fixed.

brusherru:
@nkrkv oops. Overlooked this comment 🙈 
Fixed :)

PR #1575, opened 2018-11-28
Fix error on successful arduino libs installation

### LINKED ISSUE #1546: Misleading error when downloading Arduino library
### Environment

- Scope: desktop IDE
- Version: 0.25.2

### Steps to reproduce

- create a native patch and place any valid `#pragma XOD require` directive in the C++ implementation (`#pragma XOD require "https://github.com/adafruit/Adafruit-PN532"` from [our guide](https://xod.io/docs/guide/wrapping-arduino-libraries/) will do)
- try to upload patch that uses that node
- agree to install missing Arduino libraries
- this error message appears:
screenshot 2018-11-15 at 00 28 56

Notably, despite the error message, the library actually installs (second upload attempt succeeds without asking for library).

evgenykochetkov:
Closes #1546

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,5 +81,8 @@
     "webpack": "^3.10.0",
     "webpack-dev-server": "^2.9.6",
     "webpack-merge": "^4.1.1"
+  },
+  "resolutions": {
+    "**/ramda-fantasy/ramda":  "^0.24.1"
   }
 }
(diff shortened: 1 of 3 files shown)


evgenykochetkov:
@brusherru, `tapP` is fine. `tap` from ramda@0.25.0 was the culprit. And because 0.25.0 ended up in asar file despite the fact that we use 0.24.1 everywhere led to error being reproducible only in production build.

PR #1579, opened 2018-11-29
Add pin labels validation in tabtests

brusherru:
Now it throws errors if tabtest:
- does not have columns for some pins (missing pin labels)
- have redundant pins, that are missing in tested node
- have duplicated pin labels
And it works fine with "special" columns, like `__time(ms)`.

Also, I've added `List.contains` and `List.difference` into `belt-holes` package.

### DIFF
--- a/packages/xod-tabtest/src/TabData.re
+++ b/packages/xod-tabtest/src/TabData.re
@@ -12,9 +12,7 @@ module Value = {
     | String(string)
     | Byte(int)
     | Pulse(bool)
-    | Millis(int)
     | Invalid(string);
-  let uintRegex = [%re {|/^\d+$/|}];
   let numberRegex = [%re {|/^[+-]?(?=.)*\d*(?:\.\d+)?$/|}];
   let approxNumberRegex = [%re {|/^[+-]?(?=.)*\d*(?:\.\d+)?~$/|}];
   let stringRegex = [%re {|/^".*"$/|}];
@@ -62,12 +60,6 @@ module Value = {
       String(unquote(quotedString))
     | byteString when Re.test(byteString, byteRegex) =>
       byteStringToInt(byteString)
-    | x => Invalid(x)
-    };
-  let parseTime = str =>
-    switch (str) {
-    | timeString when Re.test(timeString, uintRegex) =>
-      Millis(int_of_string(timeString))
     | x => Invalid(x)
     };
 };
@@ -120,13 +112,9 @@ let parse = (tsvSource: string) : t =>
       | None => []
       | Some(firstLine) =>
         let header = tabSplit(firstLine);
-        let timeColumnIndex =
-          header |. List.toArray |> Js.Array.indexOf("__time(ms)");
         let lineToRecord = (line: string) : Record.t =>
           tabSplit(line)
-          |. List.mapWithIndex((i, el) =>
-               i === timeColumnIndex ? Value.parseTime(el) : Value.parse(el)
-             )
+          |. List.map(Value.parse)
           |. List.zip(header, _)
           |. Record.fromPairs;
         List.tailExn(lines)
--- a/packages/xod-tabtest/src/TabData.re
+++ b/packages/xod-tabtest/src/TabData.re
@@ -102,14 +102,18 @@ let emptyLinesRegEx = [%re {|/^\s+$/gm|}];
 let tabSplit = x =>
   Js.String.split("\t", x) |. List.fromArray |. List.map(Js.String.trim);
 
-let parse = (tsvSource: string) : t =>
+let listDataLines = (tsvSource: string) : list(string) =>
   tsvSource
   |> Js.String.replaceByRe([%re {|/\r/gm|}], "")
   |> Js.String.replaceByRe(commentsRegEx, "")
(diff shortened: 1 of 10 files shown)


packages/xod-tabtest/src/TabtestValidator.re
```diff
@@ -0,0 +1,58 @@
+open Belt;
```
evgenykochetkov: Opening a lot of modules at the same time could lead to name clashes and confusion. I suggest using short aliases like `module B = Belt;`

packages/belt-holes/src/BeltHoles_List.rei
```diff
@@ -10,3 +10,10 @@ let groupByString: (list('a), 'a => string) => Map.String.t(list('a));
     and grouping the results according to values returned. */
 let groupBy:
   (list('v), Map.id('k, 'c), 'v => 'k) => Map.t('k, list('v), 'c);
+
+/** Checks a list to contain some value */
+let contains: ('a, list('a)) => bool;
```
evgenykochetkov: Belt (and thus BeltHoles) use "`t`-first" arguments order. So it should be `let contains: (list('a), 'a) => bool;`

packages/xod-tabtest/src/TabData.re
```diff
@@ -102,14 +104,18 @@ let emptyLinesRegEx = [%re {|/^\s+$/gm|}];
 let tabSplit = x =>
   Js.String.split("\t", x) |. List.fromArray |. List.map(Js.String.trim);
 
-let parse = (tsvSource: string) : t =>
+let prepareTsv = (tsvSource: string) : list(string) =>
```
evgenykochetkov: Confusing name. Prepare for what?

packages/xod-tabtest/test/TabtestValidator_jest.re
```diff
@@ -0,0 +1,156 @@
...
+        fail(
+          "Different error messages\n"
+          ++ "Expected: \""
+          ++ expectedMsg
+          ++ "\""
+          ++ "\nActual:   \""
+          ++ message
+          ++ "\"",
+        )
+    | None => fail("fuck off")
```
evgenykochetkov: 😯

packages/xod-tabtest/src/TabData.rei
```diff
@@ -38,6 +39,12 @@ let map: (t, Record.t => 'v) => List.t('v);
 /** Maps all records in data with indexes like regular lists do */
 let mapWithIndex: (t, (int, Record.t) => 'v) => List.t('v);
 
+/** Returns a list of strings separated by `\n` without comments and empty lines */
+let prepareTsv: string => list(string);
+
+/** Returns a list of strings separated by tabs and trimmed white spaces */
```
nkrkv:
```suggestion
/** Returns a list of strings separated by tabs and with leading/trailing whitespace trimmed */
```

packages/xod-tabtest/src/TabtestValidator.rei
```diff
@@ -0,0 +1,13 @@
...
+
+/**
+ * Checks tabtest for correct pin labels and throws an error if tabtest has:
+ * - missing pin labels
+ * - redundant pin labels
+ * - duplicated pin labels
+ *
+ * Otherwise do nothing.
+ */
+let assertPinLabels: (Map.String.t(Node.id), TabData.t, string) => unit;
```
nkrkv: What are we asserting? Maybe, `validatePinLabels`?

:astonished: are we making an exception a part of normal flow?! Not FP. Should be a Result of an Error with code and message.

packages/xod-tabtest/src/TabtestValidator.rei
```diff
@@ -0,0 +1,13 @@
+open Belt;
+
+open XodProject;
+
+/**
+ * Checks tabtest for correct pin labels and throws an error if tabtest has:
+ * - missing pin labels
+ * - redundant pin labels
+ * - duplicated pin labels
```
nkrkv: Args are not documented

packages/xod-tabtest/test/TabtestValidator_jest.re
nkrkv: No fucks in code, please

packages/xod-tabtest/src/TabtestValidator.re
nkrkv: Although you’re right in general, `Belt` is kind of exception. I think it is OK to `open` the `Belt` root because it is stdlib and it matches the semi-official recommendation.

Other modules, yes, should be better aliased.

packages/xod-tabtest/src/TabData.re
nkrkv: I’m suggesting `listDataLines` below

packages/xod-tabtest/test/TabtestValidator_jest.re
brusherru: Oops 🙈

brusherru:
@evgenykochetkov @nkrkv fixed! Check it out, please.
Also, I've got rid of the special handling of `__time(ms)` in `TabData` module.

PR #1580, opened 2018-11-30
Bump react to v16.3 and react-dnd to v7.0

evgenykochetkov:


nkrkv:
@evgenykochetkov +description and merge?

evgenykochetkov:
This was a quick attempt to solve a drag-and-drop bug that @gabbapeople had encountered. One of changelog entries described a pretty similar bug. Turns out it was not needed.

PR #1582, opened 2018-12-03
update catch2 and fix tabtest namespacing

knopki:
Catch2 updated to version v2.5.0 (can compile to emscripten).
Catch2 configuration tweaks (fast compilation - 20% faster).
Fix problem with namespacing for XStrings.

### DIFF
--- a/packages/xod-tabtest/src/Tabtest.re
+++ b/packages/xod-tabtest/src/Tabtest.re
@@ -273,10 +273,9 @@ module TestCase = {
         "",
         source(nodeAliases),
         "",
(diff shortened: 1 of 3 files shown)


packages/xod-tabtest/src/Tabtest.re
```diff
@@ -353,4 +352,4 @@ let generateProjectSuite = project : XResult.t(t) =>
          accFiles,
          generatePatchSuite(project, patchPath),
        )
-     );
+     );
```
nkrkv: Missing EOL

Issue #937, opened 2017-12-05
Expand node width to fit the label if possible


nkrkv:
### Rationale

![xod-narrow-nodes](https://user-images.githubusercontent.com/146383/33610383-6eddc062-d9dc-11e7-9f89-a1ce5b53cbe8.png)

Now, the node width depends only on pin count. For single/double pin nodes often we see the label with ellipsis because it is too long. However, in most of the cases, the width of a node could be safely extended without any visual collisions to contain the whole label.

This is how text cells in spreadsheet applications behave. They overflow through cell boundaries until they collide with a meaningful data. We could implement the same mechanism

### Acceptance criteria

Consider a node with `bcd-to-decimal` label and single input/output. Depending on neighbors to the right it should grow to fit the label.

- [ ] If no obstacles, the width grows to fit the whole label

```
+--------------------+
|                    |
|   bcd-to-decimal   |
|                    |
+--------------------+
```

- [ ] If there’s an obstacle right next to the node, the width stays 1 slot

```
+------+------+
|      |      |
|bcd-t…| stop |
|      |      |
+------+------+
```

- [ ] If there’s an obstacle few slots away, the width grows until it can:

```
+-------------+------+
|             |      |
|bcd-to-decim…| stop |
|             |      |
+-------------+------+
```

PR #1585, opened 2018-12-07
Add tabtest viewer to IDE

### LINKED ISSUE #1540: As a xoder I want to view tabtests in IDE so that I can better understand what node does
### Rationale

We have the tabular tests framework for some time, however, the tests themselves are hardly accessible by a casual xoder. Test contents can make node functionality obvious at the first glance, that is, they are an important part of the node’s documentation. Tabtests should be as easy to view as it is done for C++ implementation.

### User story

Patches with tabular tests have a `xod/patch-nodes/tabtest` marker on it. Double-clicking on it drills me down to the contents like it would for `not-implemented-in-xod`.

### Acceptance criteria

- [ ] I can place a `tabtest` marker and it will pick up the `patch.test.tsv` as a drill-down content
- [ ] I can cut/copy/paste/delete the `tabtest` marker and it carries test contents with it (like `not-implemented-in-xod` does)
- [ ] All standard library nodes which already have tabtests got their `tabtest` marker to access the test

evgenykochetkov:
Closes #1540


nkrkv:
Looks nice, but I'm experiencing some strange behavior.

### Steps to reproduce

1. Drill down to `xod/math/abs` tabtests
2. Double click the empty cell to the right of 0 on the second line to invoke the input

### Expected behavior

- No jumps/resizes are done, the content starts editing inline literally

### Actual behavior

- The cell becomes ~50px wider
- The cell becomes ~4px taller

---

### Steps to reproduce

1. Drill down to `xod/math/abs` tabtests
2. Span the selection to the two cells with 0's on the second row
3. Press Ctrl+X

### Expected behavior

- The contents are moved to the clipboard
- The cells are erased (or will erase on paste)

### Actual behavior

- The whole table is replaced with

```
IN	OUT
"some input"	"expected output"
```

- Ctrl+Z does not work to undo the mess

packages/xod-project/src/flatten.js
```diff
@@ -85,7 +85,7 @@ const terminalOriginalDirectionLens = R.lensProp('originalDirection');
 const isLeafPatchWithImplsOrTerminal = def(
   'isLeafPatchWithImplsOrTerminal :: Patch -> Boolean',
   R.anyPass([
-    Patch.hasImpl,
+    R.both(Patch.isPatchNotImplementedInXod, Patch.hasImpl),
```
nkrkv: This check looks strange: if an apple is green and greenish.

Why can’t we have a single function to check for the implementation presence? That is, we should kill either `Patch.isPatchNotImplementedInXod` or `Patch.hasImpl`

packages/xod-client/src/editor/containers/Editor.jsx
```diff
@@ -149,33 +150,62 @@ class Editor extends React.Component {
...
+                      newContents
+                    )
+                  }
+                  isInDebuggerTab={type === TAB_TYPES.DEBUGGER}
+                  onClose={this.props.actions.closeAttachmentEditor}
+                  patchPath={currentPatchPath}
+                />
+              );
+            }
+
```
nkrkv: For consistency I would continue the block to be like:

```js
if (editedAttachment === XP.TABTEST_MARKER_PATH) {
    return <TabtestEditor ... />
} else if (editedAttachment === XP.NOT_IMPLEMENTED_IN_XOD_PATH) {
    return <CppImplementationEditor ... />
} else {
    // Should happen only if there’s a bug in XOD IDE
    throw new Error(`Unknown attachment type ${editedAttachment}`);
}
```

packages/xod-client/src/project/reducer.js
```diff
@@ -288,12 +295,16 @@ export default (state = {}, action) => {
         XP.lensPatch(patchPath), // TODO: can we have a situation where patch does not exist?
         R.compose(
           XP.assocNode(newNode),
+          // TODO: should this be in xod-project? (along with MANAGED_ATTACHMENT_TEMPLATES)
```
nkrkv: The response is dictated by “Should the same behavior take place when I’m using `xodc`?” Well, I can’t add a node with `xodc` :grin: But I think it can be necessary at some point (as a tool to auto-generate something). So, I think, yes it is better suitable for `xod-project`

packages/xod-project/built-in-patches.xodball
```diff
@@ -60,6 +60,10 @@
     "@/output-self": {
       "path": "@/output-self"
     },
+    "@/tabtest": {
+      "path": "@/tabtest",
+      "description": " A marker node which defines a tabular test (tabtest) for the patch node. A tabtest is a kind of unit test expressed as a send-in/check-out table."
```
nkrkv:
```suggestion
      "description": "A marker node which defines a tabular test (tabtest) for the patch it is placed on. A tabtest is a kind of unit test expressed as a send-in/check-out table."
```

packages/xod-project/src/attachment.js
```diff
@@ -11,14 +11,21 @@ export const createAttachment = def(
...
 );
 
-export const createImplAttachment = def(
-  'createImplAttachment :: Source -> Attachment',
-  content => createAttachment(IMPL_FILENAME, 'utf-8', content)
+export const createAttachmentManagedByMarker = def(
+  'createAttachmentManagedByMarker :: PatchPath -> String -> Attachment',
+  (markerName, content) =>
+    createAttachment(
+      // TODO: what about unknown filenames?
```
nkrkv: I think it’s a bug. Pick from `MANAGED_ATTACHMENT_FILENAMES` with Maybe, explode with assertion

packages/xod-project/src/flatten.js
```diff
@@ -795,6 +795,7 @@ export const extractPatches = R.curry(
 const convertPatch = def(
   'convertPatch :: Project -> [Pair PatchPath Patch] -> Patch -> Patch',
   (project, leafPatches, patch) =>
+    // TODO: check for marker instead of hasImpl?
```
nkrkv: See above. There should be just one proper way to check impl.

brusherru:
About UX:
1. It would be awesome if I can add a column or row without manually moving cells.
2. The first row could have special styles

nkrkv:
> Looks nice, but I'm experiencing some strange behavior.
> ### Steps to reproduce
> 
>     1. Drill down to `xod/math/abs` tabtests
> 
>     2. Double click the empty cell to the right of 0 on the second line to invoke the input
> 
> 
> ### Expected behavior
> 
>     * No jumps/resizes are done, the content starts editing inline literally
> 
> 
> ### Actual behavior
> 
>     * The cell becomes ~50px wider
> 
>     * The cell becomes ~4px taller

The behavior is still here on Linux. Interesting thing: I can shut it up if add:

```css
.data-grid-container .data-grid .cell > input {
  width: 140px;
}
```

140px is a measured width of a `.cell` in the normal state.

nkrkv:
The table looks more organized if the very first row is rendered center-aligned and in bold:

```css
.tabtest-editor tr:first-child .cell {
    font-weight: bold;
    text-align: center;
}
```

knopki:
IMHO, "tabt..." looks not very nice. There is no easy way to make node two-cells-wide or just name it "tests"?

nkrkv:
May I suggest alternate coloring?

screenshot from 2018-12-18 11-50-07

It’s 2px borders of the color matching background `$chrome-bg` and foreground of `$dark-light`. Selection borders should be adjusted as well.

nkrkv:
> IMHO, "tabt..." looks not very nice. There is no easy way to make node two-cells-wide or just name it "tests"?

It’s a common pain for marker nodes in XOD. Eventually, we’ll implement auto-growth (#937) and the problem goes away. As a temporary solution we might make all markers two-three cells in width and/or manually resizable. That’s a question for a separate PR though.

packages/xod-client/src/core/styles/components/ReactDatasheet.scss
```diff
@@ -0,0 +1,89 @@
...
+    background-color: rgba(0, 145, 253, 0.16);
+    transition : background-color 0ms ease ;
+}
+.data-grid-container .data-grid .cell {
+  box-sizing: border-box;
+  user-select: none;
+  cursor: cell;
+  background-color: $input-bg;
+  color: $input-color-text;
+  transition : background-color 500ms ease;
```
nkrkv:
```suggestion
   transition : background-color 100ms ease;
```

500ms is more than human reaction time, so looks boring. Animations look snappier when done in 50-200ms

packages/xod-client/src/editor/components/TabtestEditor.jsx
```diff
@@ -0,0 +1,108 @@
...
+          </li>
+        </ul>
+      </div>
+      <div className="tabtest-editor">
+        <ReactDataSheet
+          data={cells}
+          valueRenderer={R.prop('value')}
+          overflow={'nowrap'}
+          onCellsChanged={(changes, additions = []) => {
+            R.compose(
```
nkrkv: Can you please extract it to a function? Now we have a complex nested function right inside deeply nested React component.

nkrkv:
Also, what’s for the past comments about a single way to check whether a C++ implementation is present?

nkrkv:
…and I suggest making cell editor inputs left-aligned. Otherwise experience is strange when I’m entering a long comment which overflows single cell boundaries

brusherru:
🐛 **Can't copy-paste markers `tabtest` and `not-implemented-in-xod`**

brusherru:
**Bad UX**
When I write a wide tabtest and horizontal scrollbar has appeared, after any change in right column horizontal scrollbar position is dropped and I have to scroll again...

Also if I might change column width, it can fit the window :)

evgenykochetkov:
Fixed copy-pasting bug, thanks!

Navigating with arrow keys should be available in the very near future: https://github.com/nadbm/react-datasheet/pull/92

Not sure what you mean by
> Also if I might change column width, it can fit the window :)

Do you want resizable columns?

brusherru:
@evgenykochetkov yep, I'm about resizable columns :) But, actually, it's out of scope 🙈
And it's not so trivial... where to store column width? The easiest and very questionable solution that I know is to store it right in local storage like a `Map PatchPath [WidthInPx]`.

packages/xod-client/src/core/styles/components/ReactDatasheet.scss
```diff
@@ -0,0 +1,97 @@
...
+}
+
+.data-grid-container .data-grid .cell {
+  box-sizing: border-box;
+  user-select: none;
+  cursor: cell;
+  background-color: $dark-light;
+  color: $input-color-text;
+  transition: background-color 100ms ease;
+  vertical-align: middle;
```
brusherru:
```suggestion
  vertical-align: middle;
  line-height: 1.6;
```
It provides better vertical aligning. What do you think?

packages/xod-client/src/core/styles/components/ReactDatasheet.scss
```diff
@@ -0,0 +1,97 @@
...
+  overflow-x: hidden;
+}
+
+.data-grid-container .data-grid .cell .value-viewer, .data-grid-container .data-grid .cell .data-editor {
+  display: block;
+}
+
+.data-grid-container tr:first-child .cell {
+  font-weight: bold;
+  text-align: center;
```
brusherru:
```suggestion
   text-align: right;
```
All values are right aligned. So I think it will looks better if we'll do a right aligned headers too.

packages/xod-client/src/core/styles/components/ReactDatasheet.scss
```diff
@@ -0,0 +1,97 @@
...
+}
```
brusherru:
```suggestion
  background-color: $grey;
}
```
It's a little lighter than other cells.

PR #1594, opened 2018-12-14
Improve tabtest of `xod/core/modulo`

brusherru:
I'm writing an article about writing tabular tests and use `modulo` as an example.


nkrkv:
@brusherru Merge?

PR #1598, opened 2018-12-24
Fix `xod/core/select` description and add test patch

brusherru:


workspace/test/core/project.xod
```diff
@@ -0,0 +1,6 @@
+{
+  "description": "The collection of test patches over some complicated Nodes of `xod/core` library, which cannot be tested at the same patch.",
+  "license": "AGPL-3.0",
+  "name": "test-core",
```
nkrkv: This one “conflicts” with the dir name

PR #1599, opened 2018-12-25
Tweak double click on constant nodes

### LINKED ISSUE #1586: As a xoder I want double click on constant node to change its value so that I’m not confused with drilling down to nowhere
### Rationale

[Mentioned on the forum](https://forum.xod.io/t/change-constant-value-in-patch/1539)

### User story

When I double-click on a `constant-*` node instead of being drilled down, Inspector is shown for that node, the `VAL` input get focus with all the current contents (i.e., the current value) selected.

In other words:

- I double click a number constant
- Press 4, press 2, press Enter
- The constant gets the value of 42

### Acceptance criteria

* [x] Double-click works as described
* [x] If Inspector was hidden at the moment of the double-click, it shows up

brusherru:
It closes #1586


packages/xod-client/src/core/domSideeffectsMiddleware.js
```diff
@@ -0,0 +1,31 @@
...
+          '.PinWidget input, .PinWidget select'
+        );
+        if (!ctrl) return;
+        if (ctrl.setSelectionRange)
+          ctrl.setSelectionRange(0, ctrl.value.length);
+        ctrl.focus();
+      })
+      .catch(err => {
+        // TODO: Add User-friendly error message?
+        throw err;
```
nkrkv: This should never happen as I understand. So if it happens it is a bug. Just throwing the error here looks fine. Suggest removing TODO

Cesar-S:
Hi, the topic of the forum is closed, and in order not to open another issues, could the same thing be done with the pins? without deepening

br

PR #1602, opened 2018-12-25
Add ability to compile tabtest in the cloud and run it locally

### LINKED ISSUE #1543: As a xoder I want to run tabtests from IDE so that they are accessible for casual use
### Rationale

Having tabular tests runnable from CLI is good but not good enough for an average xoder. UX smoothly embedded into IDE is desired.

### User story

I enter a tabtest editor and create/edit the data. I see a “Run” button on the breadcrumbs header along with quota remaining next to it. I click the button and Deployment panel pops up with a new “Tester” logging tab.

There I see:

```
Generating C++ code...
Building...
Running tests...

./is-nan.run
===============================================================================
All tests passed (7 assertions in 1 test case)
 ```

At the end of the process, a snackbar message (success or fail) is shown:

```markdown
**Tests passed**
All tests passed (7 assertions in 1 test case)
```

or

```markdown
**Tests failed**
Assertions: 5 | 4 passed | 1 failed
See the log for details.
```

On compilation error, the following gets added to the log:

```markdown
C++ compiler exited with an error. See the log for details.
The generated C++ code contains errors. It can be due to a bad node implementation
or if your board is not compatible with XOD runtime code. The original compiler error
message is above. Fix C++ errors to continue. If you believe it is a bug, report the
problem to XOD developers.
```

### Acceptance criteria

- [x] I can click Run button to get the UX described
- [x] The button gets disabled while the test is in progress
- [ ] The quota gets updated on the process completion

### Out of scope

- Applying any extra styles to the grid depending on success/fail

evgenykochetkov:
Closes #1543


packages/xod-cloud-tabtest/test/endToEnd.spec.js
```diff
@@ -0,0 +1,38 @@
...
+    loadProject(
+      [path.resolve(__dirname, '../../../workspace')],
+      path.resolve(__dirname, '../../../workspace/__lib__/xod/bits')
+    )
+      .then(
+        R.compose(eitherToPromise, project =>
+          generatePatchSuite(project, '@/bcd-to-dec')
+        )
+      )
+      .then(compileSuite('xod.show')) // TODO
```
knopki: :suspect:

knopki:
Can you pull runtime.js from service again? I have a suspicion that we both made edits to it in parallel. Just to be sure.

knopki:
Not sure, do you send `main.cpp` and `catch.hpp` in payload? Then can be dropped (for speed up) because backend service use own bundled versions.

packages/xod-cloud-tabtest/src/compile.js
```diff
@@ -0,0 +1,59 @@
...
+import * as EC from './errorCodes';
+
+// :: String -> StrMap Source -> Promise
+const compileSuite = R.curry((hostname, suite) => {
+  const reqUrl = `https://api.${hostname}/compile/enqueue`;
+  const reqBody = R.compose(
+    JSON.stringify.bind(JSON),
+    filesMap => ({
+      fqbn: 'wasm:tabtest',
+      options: {
```
knopki: Maybe put this to `vendor` dir with `runtime.js`? Because this is very static options.

knopki:
@nkrkv Same error. I think we need `electron` >= 1.7.0 or some magic startup argument (as old nodejs and chromium had).

nkrkv:
@knopki would you help @evgenykochetkov with the research?

knopki:
@evgenykochetkov @nkrkv  `--expose-wasm` (from nodejs) and `--enable-webassembly` (from chromium) and `--js-flags="--expose-wasm"` doesn't help. Unfortunately, `electron` update is must. WebAssembly support added and enabled by default in 1.7.0, but we are on 1.6 branch.

PR #1603, opened 2018-12-26
Do not save terminal patches for custom types

### LINKED ISSUE #1593: Do NOT save terminals for users custom types
### Environment
* Scope: desktop IDE
* Version: 0.25.3

### Steps to reproduce
1.  Create new project
2.  Create patch `a` and place `output-self` and `not-implemented-in-xod`
3.  Save project as a multi-file project

### Expected behavior
On FS in the project directory, we will see two directories: `main`, `a` and `project.xod` file.

### Actual behavior
It also saves `input-a` and `output-a`.

**Interesting thing**
If you delete it manually and then reload project and save again — it will not be saved again. So it reproduces only on the first saving of the project.

brusherru:
It fixes #1593

### DIFF
--- a/packages/xod-fs/src/unpack.js
+++ b/packages/xod-fs/src/unpack.js
@@ -55,6 +55,7 @@ export const arrangeByFiles = def(
     ];
     const patchFiles = R.compose(
       R.chain(arrangePatchByFiles),
+      R.filter(XP.isGenuinePatch),
       XP.listLocalPatches
     )(project);
 


packages/xod-fs/src/unpack.js
```diff
@@ -55,6 +55,7 @@ export const arrangeByFiles = def(
     ];
     const patchFiles = R.compose(
       R.chain(arrangePatchByFiles),
+      R.reject(XP.isCustomTypeTerminalPatch),
```
evgenykochetkov: Why not perform the check using `isGenuinePatch`?

PR #1606, opened 2018-12-27
Fix readme generation in xod-cli

knopki:


.prettierignore
```diff
@@ -0,0 +1 @@
+packages/xod-cli/README.md
```
nkrkv: Missing EOL

PR #1648, opened 2019-01-25
Run tabtests in web worker

brusherru:
There is no issue, but this is a prerequisite for issues about simulation.

Also, I've removed `runSuite` and vendored `runtime.js` from `xod-cloud-tabtest`, because now we get an URL from service response and then import it into the worker. So I have to remove end to end test, but I'll bring something like this back in the next pull requests.


packages/xod-client/src/debugger/messages.js
```diff
@@ -21,3 +21,5 @@ or if your board is not compatible with XOD runtime code. The original compiler
 message is above. Fix C++ errors to continue. If you believe it is a bug, report the
 problem to XOD developers.`;
 export const TABTEST_RUNNING = 'Running tests...';
+export const TABTEST_ABORTED = 'Tests aborted by user';
```
knopki: Don't like plural forms here. 
`Running tests...` => `Running test…`
`Tests aborted by user'` => `The test was interrupted by the user`
`Can't get URL to WASM runtime script` => `Unable to get URL of WASM runtime script`

yarn.lock
```diff
@@ -698,6 +698,16 @@ ajv@^5.1.0:
     json-schema-traverse "^0.3.0"
     json-stable-stringify "^1.0.1"
 
+ajv@^6.1.0:
```
knopki: `ajv` and `loader-utils` can be resolved to upper versions. `yarn-deduplicate` can do that.

packages/xod-client/src/workers/reducer.js
```diff
@@ -0,0 +1,23 @@
...
+  switch (action.type) {
+    case EAT.TABTEST_LAUNCHED: {
+      return R.assoc(action.payload.workerId, action.payload.worker, state);
+    }
+    case EAT.TABTEST_RUN_FINISHED:
+    case EAT.TABTEST_ERROR: {
+      const workerId = action.meta.workerId;
+      const worker = R.prop(workerId, state);
+      if (worker) {
+        worker.terminate();
```
nkrkv: Reducers must not have side effects. Try to make a pure test for this reducer? Termination should belong to a dispatcher or middleware.

packages/xod-client/src/workers/run.js
```diff
@@ -0,0 +1,27 @@
+import * as R from 'ramda';
+import { noop } from 'xod-func-tools';
+import { prepareModuleOptions } from 'xod-cloud-tabtest';
+
+import WasmWorker from './wasm.worker';
```
nkrkv: Why `new.module.naming`? Should it be `wasmWorker` instead? I can’t understand why does it even work? You don’t export anything `wasm.worker.js` :confused: 

What am I missing?

packages/xod-client/src/workers/run.js
```diff
@@ -0,0 +1,27 @@
...
+
+const runWasmWorker = ({ suite, runtimeUrl, handlers = {} }) => {
+  const worker = new WasmWorker();
+  const workerId = Date.now(); // TODO: Replace it!
+  worker.id = workerId;
```
nkrkv: If `id` isn’t a native property of a worker, should we inject it here? It’s better to return a record with `id` and `worker` as fields. Also, should the `Date.now` be replaced with a Posix timestamp or something?

.eslintrc.js
```diff
@@ -40,6 +40,8 @@ module.exports = {
     after: true,
     afterEach: true,
     URLSearchParams: true,
+    Worker: true,
+    self: true,
```
nkrkv: Too broad setting. Placing a single comment in the beginning of `wasm.worker.js` should be enough

packages/xod-client/webpack.config.js
```diff
@@ -130,6 +134,11 @@ module.exports = pkgDir => ({
     new webpack.NoEmitOnErrorsPlugin(),
     new CopyWebpackPlugin([
       { from: findup('node_modules/xod-client/src/core/assets/index.html') },
+      // {
+      //   test: /\.js$/,
+      //   from: findup('node_modules/xod-client/src/workers/runners'),
+      //   to: './workers',
+      // },
```
nkrkv: Trash?

packages/xod-client/src/editor/actions.js
```diff
@@ -717,71 +725,89 @@ export const runTabtest = patchPath => (dispatch, getState) => {
     ProjectSelectors.getProject
   )(getState());
 
+  const abortSymbol = Symbol('ABORT');
+  const shouldContinue = () => Selectors.isTabtestRunning(getState());
+  const abortOrPass = fn => x => {
+    if (!shouldContinue()) {
+      return Promise.reject(abortSymbol);
```
nkrkv: A new pattern detected. I’d stick to the current system: a custom `code` of a new `Error` object

packages/xod-client/src/workers/reducer.js
```diff
@@ -7,12 +7,14 @@ const workersReducer = (state = {}, action) => {
...
     case EAT.TABTEST_ERROR: {
-      const workerId = action.meta.workerId;
+      const workerId = R.path(['meta', 'workerId'], action);
+      if (!workerId) return state;
+
       const worker = R.prop(workerId, state);
-      if (worker) {
-        worker.terminate();
-      }
+      if (worker) worker.terminate();
```
nkrkv: No side effects in reducer

package.json
```diff
@@ -80,9 +80,10 @@
     "uglifyjs-webpack-plugin": "^1.1.4",
     "webpack": "^3.10.0",
     "webpack-dev-server": "^2.9.6",
-    "webpack-merge": "^4.1.1"
+    "webpack-merge": "^4.1.1",
+    "worker-loader": "^2.0.0"
```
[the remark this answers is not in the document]
brusherru: "o" > "e" 🤔
Also, I've added it using command `yarn add worker-loader` :)

packages/xod-client/src/editor/actions.js
```diff
@@ -710,19 +717,53 @@ export const runTabtest = patchPath => (dispatch, getState) => {
...
-    .then(runSuite)
+    .then(
+      suite =>
+        new Promise((resolve, reject) => {
+          const stdout = [];
+          const stderr = [];
+
+          const worker = runWasmWorker({
+            suite,
+            runtimeUrl: 'http://localhost:8000/vendor/runtime.inl.js',
```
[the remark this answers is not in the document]
brusherru: Looks like you commented outdated code 🙈

package.json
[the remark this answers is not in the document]
nkrkv: Oh, my bad. All’s fine

packages/xod-client/src/workers/run.js
brusherru: New dependency `worker-loader` takes care about it

brusherru: About naming (`wasm.worker.js`, if you mean that), it points that file is not a js module like others, this is js for the web worker. And it will be loaded using `worker-loader`.

nkrkv: Ok, it’s clear now. Would you pass the comment from the top of `wasm.worker.js` through Grammarly? It contains too many mistakes to communicate the idea.

packages/xod-client/src/editor/actions.js
```diff
@@ -710,39 +726,69 @@ export const runTabtest = patchPath => (dispatch, getState) => {
...
+        runWasmWorker(suite, worker =>
+          dispatch({
+            type: ActionType.TABTEST_LAUNCHED,
+            payload: { worker },
+          })
+        )
+      )
+    )
+    .then(
+      abortOrPass(({ stdout, worker }) => {
```
nkrkv: These repetitive `.then(abortOrPass(...))` is yet another reason to upgrade to latest Ramda. Its `R.composeWith` and `R.pipeWith` make this chain more elegant and shallowly nested.

Just a note, not a change request.

packages/xod-client/src/workers/wasm.worker.js
```diff
@@ -0,0 +1,46 @@
+// Worker imported as usual js file (using `worker-loader`)
+// so tests fails with ReferenceError "self is not defined"
+// to avoid this we define _self as self for worker
+// or empty object for tests and etc.
```
nkrkv: :scroll: This file is a WASM web worker implementation.
The worker is imported as a usual JS module by `worker-loader`.
In this case tests fail with ReferenceError "self is not defined".
To avoid this we define _self as self for worker, or an empty object
for tests and etc.

packages/xod-client/src/debugger/containers/Debugger.jsx
```diff
@@ -129,7 +133,20 @@ class Debugger extends React.Component {
...
             <span className="title">Deployment</span>
 
-            <div className="progress">{uploadProgress}</div>
+            <div className="progress">
+              {uploadProgress}
+              {isTabtestRunning ? (
+                <Icon
+                  Component="button"
+                  className="abort-tabtest-button"
+                  name="ban"
```
nkrkv: A title is required.

Suggest: “Abort Test”

PR #1649, opened 2019-01-31
Simulate program and see values in watches

### LINKED ISSUE #32: As a xoder I want to simulate my program so that I can see what output will be without touching board
### Rationale

Having a feature to run a program on PC without any Arduino opens XOD to people who don’t yet have a board, for very quick experiments, and quicker development.

### User story

- I put `system-time` linked to `watch`
- I hit Deploy → Simulate
- A progress bar in the Deployment pane header appears
- When done, the Debug tab appears in the work area exactly as if I upload with the debugger
- The watch is live and shows time past the start of Simulation

### Acceptance criteria

- [x] The basic scenario works as described above
- [x] The behavior of the Deployment pane is the same as while uploading with the debugger (collapsed by default, logs are populated, tabs are auto-switched)

### How to implement

- In C++ make a `WasmSerial` object which mimics the standard `Serial` print/write/flush operations
- When compiling use macros to define `XOD_DEBUG_SERIAL` as `WasmSerial` in case of simulation and basic `DEBUG_SERIAL` otherwise. Use the `XOD_DEBUG_SERIAL` in the `watch` implementation

brusherru:
It closes #32


packages/xod-client/src/core/containers/App.jsx
```diff
@@ -74,12 +79,48 @@ export default class App extends React.Component {
...
+      R.map(
+        R.applySpec({
+          code: transpile,
+          nodeIdsMap: getNodeIdsMap,
+        })
+      )
+    )(eitherTProject);
+  }
+
+  transformProjectForTranspiler(debug = false, simulation = false) {
```
nkrkv: Please, don’t use positional args defaults. They complicate currying and code reasoning.

packages/xod-client/src/debugger/messages.js
```diff
@@ -14,12 +14,19 @@ export const UPDATING_ARDUINO_DEPENDECIES = 'Updating Arduino packages...';
 export const UPDATING_ARDUINO_DEPENDECIES_SUCCESS =
   'All Arduino packages are updated successfully';
 
-export const TABTEST_GENERATING_CODE = 'Generating C++ code...';
-export const TABTEST_BUILDING = 'Building...';
-export const TABTEST_BUILDING_ERROR = `The generated C++ code contains errors. It can be due to a bad node implementation
+export const WASM_NO_RUNTIME_FOUND = 'Unable to get URL of WASM runtime script';
+export const WASM_BUILDING_ERROR = `The generated C++ code contains errors. It can be due to a bad node implementation
```
nkrkv: Didn’t ESLint caught you? This line should be hard-wrapped

packages/xod-client/src/debugger/messages.js
```diff
@@ -14,12 +14,19 @@ export const UPDATING_ARDUINO_DEPENDECIES = 'Updating Arduino packages...';
...
+
+export const TABTEST_GENERATING_CODE = 'Generating C++ code...';
+export const TABTEST_BUILDING = 'Building...';
 export const TABTEST_RUNNING = 'Running tabular test...';
 export const TABTEST_ABORTED = 'The test was interrupted by the user';
-export const WASM_NO_RUNTIME_FOUND = 'Unable to get URL of WASM runtime script';
+
+export const SIMULATION_GENERATING_CODE = 'Generating C++ code for simulation…';
+export const SIMULATION_BUILDING = 'Building…';
+export const SIMULATION_RUNNING = 'Running simulation…';
```
nkrkv: Inconsistency between `...` and `…` in the messages. Use either of variant everywhere. I suggest sticking to `...` because we’re using a monospaced font.

packages/xod-client/src/workers/run.js
```diff
@@ -4,36 +4,49 @@ import { prepareModuleOptions, ERROR_CODES } from 'xod-cloud-tabtest';
...
 
     const worker = new WasmWorker();
-
     worker.postMessage({
       type: 'init',
       payload: {
         suite: prepareModuleOptions(suite),
         runtimeUrl,
+        // HACK:
+        wasmUrl: R.path(['artifactUrls', 'main.wasm'], suite),
```
nkrkv: Hack? Some elaboration would be cool

packages/xod-client/src/workers/run.js
```diff
@@ -4,36 +4,49 @@ import { prepareModuleOptions, ERROR_CODES } from 'xod-cloud-tabtest';
...
       },
     });
 
+    worker.onReceive = noop;
+
     const handlers = {
+      'serial:receive': data => worker.onReceive(data),
+      // 'serial:sendOk': len => console.log('Data sent. Buffer length: ', len),
```
nkrkv: Forgotten comment?

workspace/__lib__/xod/core/watch/patch.cpp
```diff
@@ -10,14 +10,14 @@ void evaluate(Context ctx) {
...
   TimeMs tNow = transactionTime();
-  DEBUG_SERIAL.print(F("+XOD:"));
-  DEBUG_SERIAL.print(tNow);
-  DEBUG_SERIAL.print(':');
-  DEBUG_SERIAL.print(getNodeId(ctx));
-  DEBUG_SERIAL.print(':');
+  XOD_DEBUG_SERIAL.print(F("+XOD:"));
+  XOD_DEBUG_SERIAL.print(tNow);
+  XOD_DEBUG_SERIAL.print(':');
+  XOD_DEBUG_SERIAL.print(getNodeId(ctx));
```
nkrkv: Should the similar be applied to `xod/core/console-log`?

packages/xod-client/src/core/containers/App.jsx
```diff
@@ -74,12 +79,48 @@ export default class App extends React.Component {
     )();
   }
 
-  transformProjectForTranspiler(debug = false) {
+  onRunSimulation() {
+    if (this.props.isSimulationRunning) {
+      this.props.actions.addError(SIMULATION_ALREADY_RUNNING);
```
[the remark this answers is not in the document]
nkrkv: If this is not trivial, let it be

nkrkv:
A few UX observations:

- The gamepad icon might be made larger to align in visual weight with neighbors. `font-size: 16px` looks fine
- The button tooltip is “Run simulation” whereas the menu item is “Simulate”. They should be consistent. Just “Simulate” on both is good
- When starting a simulation the log shows “Generating C++ code for simulation…” / “Building...” I think the messages belong to the “Compiler” tab, because the “regular” debug session works this way.
- The message “Running simulation...” is for “Uploader” tab?!
- “Simulation is launched” and “The simulation was interrupted by the user” should be consistent with those used by the debugger, the wording and the `===` affixes
- If I change the program and hover the (?) icon I see the message about a board/upload/debug which is wrong in the context of the simulation.

nkrkv:
A very annoying behavior is number auto-roundings. So, a value of `3.00` gets rendered as `3` while simulation. Take a look at a watch linked to `system-time`, its content jumps. I expect it to be the same as on a board: `3.00`, with the zeros preserved.

nkrkv:
@evgenykochetkov is going to rename the “Debugger” tab to “Serial”. When it will be done, the current output of the simulation starts to look misplaced.

packages/xod-client/src/debugger/messages.js
brusherru: Nope. But it does if I adds a new line before the string :)

packages/xod-client/src/core/styles/components/Debugger.scss
```diff
@@ -76,15 +76,15 @@
...
 
       &:hover {
-        color: #ccc;
+        color: #aaa;
         background: #444;
       }
     }
 
-    .abort-tabtest-button {
+    .abort-process-button {
```
evgenykochetkov: 👍🏻

PR #1661, opened 2019-02-04
Stop simulation on closing debugger tab

brusherru:
There is no issue 😬


brusherru:
@nkrkv added two more middlewares and decoupled simulation from middleware about stop debug session. Check it out, please

PR #1662, opened 2019-02-04
C++ implementation for tweak nodes

evgenykochetkov:
Would appreciate feedback about messages format (do we need to begin them with some unique symbol to avoid consuming _all_ serial input?)

### DIFF
--- a/packages/xod-arduino/platform/program.tpl.cpp
+++ b/packages/xod-arduino/platform/program.tpl.cpp
@@ -39,11 +39,57 @@ namespace xod {
 {{/unless}}
 {{/each}}
 
+#if defined(XOD_DEBUG) // can't do that in XOD_SIMULATION yet
+namespace detail {
+void handleTweaks() {
+    if (XOD_DEBUG_SERIAL.available() > 0 && XOD_DEBUG_SERIAL.find("+XOD:", 5)) {
+        int tweakedNodeId = XOD_DEBUG_SERIAL.parseInt();
+
+        switch (tweakedNodeId) {
+          {{#eachTweakNode nodes}}
+            case {{ id }}:
+            {{#switch patch.patchPath}}
+              {{#case "xod/debug/tweak-number"}}
+                node_{{ id }}.output_OUT = XOD_DEBUG_SERIAL.parseFloat();
+              {{/case}}
+              {{#case "xod/debug/tweak-byte"}}
+                node_{{ id }}.output_OUT = XOD_DEBUG_SERIAL.parseInt();
+              {{/case}}
+              {{#case "xod/debug/tweak-pulse"}}
+                node_{{ id }}.output_OUT = 1;
+              {{/case}}
+              {{#case "xod/debug/tweak-boolean"}}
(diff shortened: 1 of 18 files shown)


packages/xod-arduino/platform/program.tpl.cpp
```diff
@@ -39,12 +39,63 @@ namespace xod {
...
+char debuggerProtocolPrefix[] = "+XOD:";
+uint8_t prefixLength = strlen(debuggerProtocolPrefix);
+
 void runTransaction() {
     g_transactionTime = millis();
 
     XOD_TRACE_F("Transaction started, t=");
     XOD_TRACE_LN(g_transactionTime);
 
+    #if defined(XOD_DEBUG)
```
brusherru: I suggest using `#ifdef XOD_DEBUG_SERIAL` and `XOD_DEBUG_SERIAL` instead of `DEBUG_SERIAL`, so this code will automatically work fine in the Simulation mode. I hope :)

workspace/__lib__/xod/debug/tweak-string-16/patch.cpp
```diff
@@ -0,0 +1,20 @@
...
+
+void evaluate(Context ctx) {
+    auto state = getState(ctx);
+
+    if (isSettingUp()) {
+        state->view = CStringView(state->buff);
+    }
+
+    // Additional code that sets value of `buff`
+    // if automagically injected elsewhere
```
nkrkv:
```suggestion
    // is automagically injected elsewhere
```

workspace/__lib__/xod/debug/tweak-byte/patch.cpp
```diff
@@ -0,0 +1,10 @@
+
+struct State {
+};
+
+{{ GENERATED_CODE }}
+
+void evaluate(Context ctx) {
+    // The actual code that makes tweak-nodes work
+    // is automagically injected elsewhere
```
nkrkv:
```suggestion
    // is injected in detail::handleTweaks
```

...and it should be really there :)

packages/xod-arduino/platform/program.tpl.cpp
```diff
@@ -45,6 +45,41 @@ void runTransaction() {
     XOD_TRACE_F("Transaction started, t=");
     XOD_TRACE_LN(g_transactionTime);
 
+    #if defined(XOD_DEBUG) || defined(XOD_SIMULATION)
```
```suggestion
#if defined(XOD_DEBUG) || defined(XOD_SIMULATION)
```

Do not indent preprocessor directives. This is legacy, mainly, but so strong that some editors bark on you if it is indented.

packages/xod-arduino/platform/runtime.cpp
```diff
@@ -153,6 +153,7 @@ void setup() {
     delay(100);
 #ifdef XOD_DEBUG
```
brusherru: I think it's better to use the same `XOD_DEBUG_SERIAL` here and call `begin` and `setTimeout` for the simulation too.

P.S.
Now `begin` in Simulation Serial is just a noop method and there is no `setTimeout` method. But it's a detail of implementation, which we should not think about in this context.

PR #1666, opened 2019-02-06
Tweak values by editing values of tweak nodes in Inspector

### LINKED ISSUE #32: As a xoder I want to simulate my program so that I can see what output will be without touching board

### LINKED ISSUE #1624: As a xoder I want to tweak numeric parameters in real-time so that I debug and develop faster
### Rationale

We have watches, they are cool. We need a counterpart as well. Such way a xoder may not only observe values in real-time but also affect his device very quickly.

### User story

- I put a `tweak-number` node, it has a single unlabeled output
- I link it to a `led`
- I upload with the debugger enabled, the debug session starts
- Now, I can select the `tweak-number` and change its value via Inspector
- On change (commit with Enter), the `led` luminance reacts accordingly

### Acceptance criteria

* [ ] The basic scenario works as described
* [ ] The node is put under `xod/debug`
* [ ] The node’s label behaves the same way as it does for `constant-*` (value replaces name unless overridden)
* [ ] The node color behaves the same way as for `watch` (darker when inactive, lighter when live)
* [ ] The node is resizable

### LINKED ISSUE #1626: Add boolean and byte tweaks
### Prerequisites

#1624 

### Acceptance criteria

* [ ] `xod/debug/tweak-boolean` acts exactly the same way as `tweak-number` but carries boolean values
* [ ] `xod/debug/tweak-byte` acts exactly the same way as `tweak-number` but carries boolean values

evgenykochetkov:
Works for all available tweak nodes except `tweak-pulse`

Closes #1624
Closes #1626 
Closes #32


packages/xod-client/src/editor/containers/Patch/modes/debugging.jsx
```diff
@@ -16,6 +16,9 @@ const debuggingMode = R.merge(selectingMode, {
   onNodeDoubleClick(api, nodeId, patchPath) {
     if (patchPath === XP.NOT_IMPLEMENTED_IN_XOD_PATH) {
       api.props.actions.openImplementationEditor();
+    } else if (XP.isConstantNodeType(patchPath) || XP.isTweakPath(patchPath)) {
+      // TODO: rename this action to something more suitable?
+      api.props.actions.selectConstantNodeValue(nodeId, api.props.patchPath);
```
nkrkv: `focusBoundValue` or `editBoundValue`?

packages/xod-client-electron/src/debugger/sendTweakValuesMiddleware.js
```diff
@@ -0,0 +1,39 @@
...
+import * as XP from 'xod-project';
+import client from 'xod-client';
+import { formatTweakMessage } from 'xod-arduino';
+import { ipcRenderer } from 'electron';
+
+import { DEBUG_SERIAL_SEND } from '../shared/events';
+
+export default ({ getState }) => next => action => {
+  const state = getState();
+  const result = next(action);
```
brusherru: I think it does not make sense in this case, but `next(action)` can change the `state`, so I prefer to get state after calling `next(action)` :)

evgenykochetkov:
- made it work in simulation as well
- fixed `tweak-string-16`

check again, please

PR #1667, opened 2019-02-07
Fix including of bundled Arduino libraries in dev mode

brusherru:
There is no issue 😬

### DIFF
--- a/packages/xod-client-electron/src/app/utils.js
+++ b/packages/xod-client-electron/src/app/utils.js
@@ -55,9 +55,27 @@ export const getFilePathToOpen = app => {
 };
 
 /**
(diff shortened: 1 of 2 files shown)


evgenykochetkov:
Me too 🤔

brusherru:
I don't know why, but:
1. when we call `getPathToBundledWorkspace()` in the renderer process in dev mode it returns `/workspace` instead of `/some/path/to/xod-repo/packages/xod-client-electron/src-babel/workspace`
2. we get just `/workspace` instead of a full path because `__dirname` is equal to `/`. And it's strange because if I open the console and write `__dirname` it will return the correct path 🤔 

So I call this function in the main process, where it gets the correct path :)

PR #1681, opened 2019-02-12
Refactor inspector widgets

### LINKED ISSUE #755: Tab ordering in the Inspector is broken
### Steps to reproduce

- Place a `text-lcd-16x2` node
- Click `RS` value input in the Inspector to focus it
- Enter a value (e.g. `8`) and press `TAB`

### Expected behavior

- The cursor moves to the next pin input (`EN`), its text is selected
- I can quickly enter a value → `TAB` → enter value → `TAB` → … → enter value → `ENTER` to bind all pin values.

### Actual behavior

- The focus is lost
- Pressing `TAB` more times brings the focus back to the `RS` input
- Pressing `TAB` again looses focus
- Pressing `TAB` one more time focuses the `EN`’s input
- I have to press `TAB` twice to move to the next input

### LINKED ISSUE #766: Simplify Inspector Widgets, split general input handling utils and code specific to Inspector

evgenykochetkov:
Closes #766
Closes #755 (`tab` now works properly)

and also:
* after committing values with Enter key, the whole contents of the input is selected
* incrementing with up/down arrows works again on Number widgets (default step is 0.1, with shift key it's 1)


nkrkv:
May we have the Enter key to commit the value (bind it) and then select the whole contents of the widget (preserving the focus) so that I can start typing a new value without extra movements?

brusherru:
@evgenykochetkov comment https://github.com/xodio/xod/pull/1681#issuecomment-462698275 still actual :)

brusherru:
Floats modified with arrow keys looks ugly 🙈

PR #1682, opened 2019-02-12
Fix normalizing pin labels for C++ and add QuickHelp for C++ and Tabtest editors

### LINKED ISSUE #1664: tabtests transpile pin names with - (dash) instead of _ (underscore)
### Environment

* Scope: desktop IDE
* Version: 0.26.1
* OS: Ubuntu 16.04
* Machine:x86_64, 32 GB RAM

### Steps to reproduce

1. Create a patch with an input named "a-a" (anything with a dash)
1. Add a tabtest
1. Add a row to tabtest, use the name "a-a"
1. Try to run the tab test
1.You'll get a c++ compile error something like this:
```
/wait-all.catch.inl:6:7: error: declaration of variable 'probe_bit' with deduced type 'auto &' requires an initializer
auto& probe_bit-count = xod::node_15;
      ^
./wait-all.catch.inl:6:16: error: expected ';' after top level declarator
auto& probe_bit-count = xod::node_15;
               ^
```

### Expected behavior

The tabtest should use the pin-name.
If you can set the pin name to something, then it should work everywhere.

### Actual behavior

c++ compile error as above.

### Console output 

above

### How to fix 

I think the pin names should be made c++ safe by changing - (dash) to _ (underscore). You have some other issues (cf. #491) that are moving toward the prohibition of dash and underscore.

Dashes make names easier to read. I would hate to see dashes go away.

it feels like pin-names (and patch-names) are used in the c++ code inconsistently. Assuming there is an object for the pin, there should be a cpp_name() method used when generating c++ code. In js it should be just
```
pin_name.replace( /-/, '_' )
```

brusherru:
It fixes #1664 
Also, it adds a new feature: QuickHelp now contains a list of inputs/outputs available in the C++ editor and in the tabtest editor. Also, it contains a link to a full reference in our docs.


packages/xod-tabtest/src/Tabtest.re
```diff
@@ -111,7 +111,7 @@ module Bench = {
       probeMap: Map.String.empty,
     };
     Patch.listPins(patchUnderTest)
-    |. Pin.normalizeLabels
+    |. Pin.normalizeLabelsForCpp
```
nkrkv: Are we forcing a xoder to enter escaped pin labels in the header row? I don’t think it’s a good idea

packages/xod-project/src/pin.js
```diff
@@ -272,6 +273,31 @@ export const normalizeEmptyPinLabels = def(
...
+  R.compose(unicodeCharsToUnicodeSequence, R.replace(/(\s|-)+/g, '_'), R.trim)
+);
+
+/**
+ * Gives pins C++ friendly labels:
+ * - Normalized empty pin labels
+ * - Normalized non-ASCII and forbidden chars
+ */
+export const normalizePinLabelsForCpp = def(
+  'normalizePinLabelsForCpp :: [Pin] -> [Pin]',
```
nkrkv: If everything other is right, this may be safely made a private detail of `xod-arduino`

packages/xod-client/src/editor/components/CppPatchDocs.jsx
```diff
@@ -0,0 +1,70 @@
+import React from 'react';
+import PT from 'prop-types';
+import * as R from 'ramda';
+import * as XP from 'xod-project';
+import Icon from 'react-fa';
+
+import { getUtmSiteUrl } from '../../utils/urls';
+
+const CppPatchDocs = ({ patch }) => {
+  const pins = R.compose(
```
evgenykochetkov:
```suggestion
  const [ inputPins, outputPins ] = R.compose(
```

packages/xod-client/src/editor/constants.js
```diff
@@ -70,6 +70,11 @@ export const TAB_TYPES = {
   DEBUGGER: 'DEBUGGER',
 };
 
+export const ATTACHMENT_TYPES = {
```
evgenykochetkov: Can `MANAGED_ATTACHMENT_FILENAMES` from `xod-project` be reused here instead?

PR #1684, opened 2019-02-13
New look of patch board with bound values

### LINKED ISSUE #1678: Improve patch board design to show bound values
### Rationale

Hiding nodes’ bound values behind Inspector is like writing a text program with parameter names `openBrowser(url, width, height)` instead of argument values `openBrowser('https://xod.io', 1024, 768)`. Usable, but suboptimal: cannot be easy transferred as an image, cannot be perceived as the whole.

### Mockup

new-patch-board

### Acceptance criteria

- [x] Nodes render as in mockup
- [x] Pins (and generic pins) render as in mockup
- [x] Bound values and pin labels too
- [x] Long values and pin labels are truncated
- [x] Constants do not render their value twice

### Out of scope

- Old projects grid migration
- Links/pins color tweaking

brusherru:
It closes #1678 

Vois-la

### DIFF
--- a/packages/xod-client-browser/test-func/creatingBlinkPatch.spec.js
+++ b/packages/xod-client-browser/test-func/creatingBlinkPatch.spec.js
@@ -91,7 +91,7 @@ describe('creating blink patch', () => {
     await clockNode.drag(150, 10);
 
     const { x, y } = await clockNode.getBoundingClientRect();
-    assert.deepEqual({ x, y }, { x: 389, y: 114 });
+    assert.deepEqual({ x, y }, { x: 399.5, y: 128.5 });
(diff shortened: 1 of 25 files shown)


evgenykochetkov:
Also, bound values are not shown in latest the Firefox and Chromium on macOS :(

evgenykochetkov:
values bound to generic pins are not colored:
screen shot 2019-02-13 at 18 44 34

brusherru:
All fixed and tweaked :)

Vois-la-2

nkrkv:
I think it’s a bit better to put node labels 2px higher: 10px top margin instead of 12px

Issue #1427, opened 2018-09-03
Add description to nodes in xod/patch-nodes


nkrkv:
### Rationale

Because nodes in `xod/patch-nodes` used to be auto-generated they are lacking any description about themselves. However, providing the description was always possible and is possible now, so it has to be done.

The absence of any note [confuses xoders](https://forum.xod.io/t/abstract-and-utility/1048/3). Let’s fix it.

### Acceptance criteria

- [ ] Every node in `xod/patch-nodes` has a description for helpbox

### How to implement

- Add patch descriptions to `built-in-patches.xodball`
- Add `input-*` and `output-*` to `built-in-patches.xodball` where `*` is a built-in type.

nkrkv:
Closed by #1685

PR #1685, opened 2019-02-13
Add meta- and generated nodes descriptions

### LINKED ISSUE #1647: missing description for the jumper node
* Scope: desktop IDE
* Version: 0.26.1

The xod/patch-nodes/jumper node has no description. In the Quick info there is no sentence what this node does.
For other nodes there is a helpful sentence like e.g. "Projects a number from one range to another preserving ratio" so it would be nice if the jumper gets such a sentence too.

nkrkv:
Closes #1647, #1427

Also, generates descriptions for custom type terminals.

From now on, all built-in type terminals must be listed in `built-in-patches.xodball`. Otherwise, a build-time error will occur.

### DIFF
--- a/packages/xod-project/built-in-patches.xodball
+++ b/packages/xod-project/built-in-patches.xodball
@@ -2,12 +2,15 @@
   "name": "built-in-patches",
   "patches": {
     "@/abstract": {
+      "description": "Makes a patch containing this node abstract. That is, a patch node which only defines a generic interface in terms of generic-type inputs and outputs without an actual implementation. To be useful there should be one or more sibling specialization patches which implement the abstraction for particular input types.",
       "path": "@/abstract"
     },
     "@/deprecated": {
+      "description": "Deprecates a patch which contains this node. Use it to archive outdated patch nodes. The projects which already uses the deprecated node will not break. Only a warning message will be shown. Use the marker node description to provide a deprecation reason and migration hint, it will be shown along the warning to end-users. For new projects, deprecated nodes are unlisted by default, so usage of such nodes by library users is discouraged.",
       "path": "@/deprecated"
     },
     "@/from-bus": {
+      "description": "Defines an attachment point to a patch-level data bus with the name defined by this node label.",
       "nodes": {
         "__out__": {
           "id": "__out__",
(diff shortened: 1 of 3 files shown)


packages/xod-project/built-in-patches.xodball
```diff
@@ -2,12 +2,15 @@
   "name": "built-in-patches",
   "patches": {
     "@/abstract": {
+      "description": "Makes a patch containing this node abstract. That is, a patch node which only defines a generic interface in terms of generic-type inputs and outputs without an actual implementation. To be useful there should be one or more sibling specialization patches which implement the abstraction for particular input types.",
       "path": "@/abstract"
     },
     "@/deprecated": {
+      "description": "Deprecates a patch which contains this node. Use it to archive outdated patch nodes. The projects which already uses the deprecated node will not break. Only a warning message will be shown. Use the marker node description to provide a deprecation reason and migration hint, it will be shown along the warning to end-users. For new projects, deprecated nodes unlisted by default, so usage of such nodes by library users is discouraged.",
```
evgenykochetkov:
```suggestion
      "description": "Deprecates a patch which contains this node. Use it to archive outdated patch nodes. The projects which already uses the deprecated node will not break. Only a warning message will be shown. Use the marker node description to provide a deprecation reason and migration hint, it will be shown along the warning to end-users. For new projects, deprecated nodes are unlisted by default, so usage of such nodes by library users is discouraged.",
```

packages/xod-project/built-in-patches.xodball
```diff
@@ -20,7 +23,48 @@
...
+    "@/input-boolean": {
+      "description": "Input terminal node. Adds a new boolean input pin to the patch node which contains this node. Terminal label and description are propagated to the pin. Horizontal position relative to other terminals defines the pin order. A value bound to the terminal node pin becomes the default for the created input pin.",
+      "path": "@/input-boolean"
+    },
+    "@/input-byte": {
+      "description": "Input terminal node. Adds a new byte input pin to the patch node which contains this node. Terminal label and description are propagated to the pin. Horizontal position relative to other terminals defines the pin order. A value bound to the terminal node pin becomes the default for the created input pin.",
+      "path": "@/input-byte"
+    },
+    "@/input-dead": {
+      "description": "This terminal should not be visible to an end-user.",
```
evgenykochetkov: 😄

PR #1688, opened 2019-02-14
Remove debug nodes from nested patches

evgenykochetkov:
Before debug-nodes were removed only from the top level.
It could cause compilation error when uploading without debug.


nkrkv:
A unit test that will check it forever would be nice

PR #1698, opened 2019-02-25
make `clock` behave like `continuously` when IVAL is set to 0

### LINKED ISSUE #1676: The changes have no effect on the simulation
### Environment

* Scope: desktop IDE, browser IDE, CLI
* Version: 0.27.0
* OS: Windows 10 

### Steps to reproduce

When making the changes, they are not reflected in the simulation.

1.  New patch
2.  Add tweak-number + clock + flip-flop  + watch
3.  Run and change value

### Expected behavior

The initial simulation in 0, then it is changed to 1, increases the speed of the output

### Actual behavior

The output remains the same as when it was in 0. (false)

P.E.: if the initial value is greater than 0, it works correctly

evgenykochetkov:
Closes #1676

### DIFF
--- a/workspace/__lib__/xod/core/clock/patch.test.tsv
+++ b/workspace/__lib__/xod/core/clock/patch.test.tsv
@@ -11,3 +11,13 @@ __time(ms)	IVAL	EN	RST	TICK
 500	1	true	pulse	no-pulse
 1001	1	true	no-pulse	no-pulse
 1501	1	true	no-pulse	pulse
+// edge case: when IVAL is 0, `clock` should constantly emit pulses
+0	0	true	no-pulse	pulse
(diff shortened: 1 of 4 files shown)


workspace/__lib__/xod/core/clock/patch.test.tsv
```diff
@@ -11,3 +11,11 @@ __time(ms)	IVAL	EN	RST	TICK
 500	1	true	pulse	no-pulse
 1001	1	true	no-pulse	no-pulse
 1501	1	true	no-pulse	pulse
+0	0	true	no-pulse	pulse
```
nkrkv: A comment above the block telling what we’re testing (edge-case intervals) would be helpful

PR #1700, opened 2019-02-26
Add xod/math/min and xod/math/max

evgenykochetkov:
also tweak `xod/math/map-clip` to output correct values when target range bounds are reversed

### DIFF
--- a/workspace/__lib__/xod/math/max/patch.test.tsv
+++ b/workspace/__lib__/xod/math/max/patch.test.tsv
@@ -1,0 +1,16 @@
+IN1	IN2	OUT
+0.1	0.01	0.1
+-0.5	0.5	0.5
+500	2300	2300
+-20	-1	-1
+
+Inf	100	Inf
+-Inf	100	100
+
+// when any of inputs is NaN — return NaN
+NaN	1	NaN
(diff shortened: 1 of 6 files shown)


workspace/__lib__/xod/math/max/patch.test.tsv
```diff
@@ -0,0 +1,5 @@
+IN1	IN2	OUT
+0.1	0.01	0.1
+-0.5	0.5	0.5
+500	2300	2300
+-20	-1	-1
```
nkrkv: Tests with infinities and NaNs are desired much.

workspace/__lib__/xod/math/min/patch.test.tsv
```diff
@@ -0,0 +1,5 @@
+IN1	IN2	OUT
+0.1	0.01	0.01
+-0.5	0.5	-0.5
+500	2300	500
+-20	-1	-20
```
nkrkv: The same

workspace/__lib__/xod/math/max/patch.test.tsv
evgenykochetkov: Done!

PR #1709, opened 2019-03-01
Add possibility to set a custom width of the screenshot (docs screenshotter)

brusherru:
There is no issue.


brusherru:
We decided that this is a dangerous solution because it is tempting in which case to take screenshots wider, instead of fixing patches.

PR #1717, opened 2019-03-11
Make node and pin labels a bit brighter

evgenykochetkov:
This should deal with readability issues that were reported on our forum: https://forum.xod.io/t/xod-0-28-0-more-detailed-patch-board/1961/6?u=evgenykochetkov


nkrkv:
Agree. However, a sane deadline is required. Let’s wait for a day.

PR #1732, opened 2019-03-26
Fix default board index value

brusherru:
There is no issue. But this bug was reported on the forum: https://forum.xod.io/t/cannot-properly-install/2050

### DIFF
--- a/packages/xod-client-electron/src/upload/components/PopupUploadConfig.jsx
+++ b/packages/xod-client-electron/src/upload/components/PopupUploadConfig.jsx
@@ -43,7 +43,7 @@ class PopupUploadConfig extends React.Component {
   }
 
   onBoardChanged(event) {
-    this.changeBoard(event.target.value);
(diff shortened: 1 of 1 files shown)


packages/xod-client-electron/src/upload/components/PopupUploadConfig.jsx
```diff
@@ -94,8 +94,15 @@ class PopupUploadConfig extends React.Component {
           isBoardSelected && boards[selectedBoard.index];
 
         const defaultBoardIndex = R.compose(
-          R.defaultTo(0),
-          R.findIndex(R.propEq('fqbn', 'arduino:avr:uno'))
+          R.when(R.equals(-1), R.always(0)),
+          R.findIndex(
+            R.either(
+              // If arduino:avr is not installed yet:
+              R.propEq('name', 'Arduino/Genuino Uno'),
```
nkrkv: :+1:

PR #1741, opened 2019-04-08
Docs screenshotter auto-detects patch height if it's not specified by User

brusherru:
There is no issue. But this is a very handy tool for auto-generating new tutorial docs from a xodball.
Also, we can remove manually specified height from most of the `update-screenshots.sh` to make the same paddings on all screenshots.

### DIFF
--- a/tools/screenshot-xodball
+++ b/tools/screenshot-xodball
@@ -23,6 +23,38 @@ const {
   default: EditorTab,
 } = require('xod-client-browser/test-func/pageObjects/EditorTab');
 
+// For bounding box calculation:
+const path = require('path');
+const { loadProject } = require('xod-fs/dist/load');
+const XP = require('xod-project/dist');
+const {
+  SLOT_SIZE,
+  slotPositionToPixels,
+} = require('xod-client/dist/project/nodeLayout');
+const getBBox = require('xod-client/dist/editor/utils').getBoundingBoxSize;
+
+const getPatchHeight = (pathToProject, patchName) =>
+  loadProject([path.resolve(__dirname, '../workspace')], pathToProject).then(
+    proj => {
+      const patch = XP.getPatchByPathUnsafe(`@/${patchName}`, proj);
+      const entities = {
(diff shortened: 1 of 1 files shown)


tools/screenshot-xodball
```diff
@@ -23,6 +23,40 @@ const {
   default: EditorTab,
 } = require('xod-client-browser/test-func/pageObjects/EditorTab');
 
+// For bounding box calculation:
+const path = require('path');
+const loadProject = require('xod-fs/dist/load').loadProject;
```
evgenykochetkov:
```suggestion
const { loadProject } = require('xod-fs/dist/load');
```

tools/screenshot-xodball
```diff
@@ -23,6 +23,40 @@ const {
...
+        links: XP.listLinks(patch),
+      };
+      return (
+        slotPositionToPixels(getBBox(patch, proj, entities)).y + SLOT_SIZE.GAP
+      );
+    }
+  );
+
+const calculateHeight = (pathToProject, patchName, desiredHeight) =>
+  new Promise((resolve, reject) => {
```
evgenykochetkov: Why `new Promise`?
I suggest something like
```
const h = parseInt(desiredHeight, 10);
return h ? Promise.resolve(h) : getPatchBbox(pathToProject, patchName)
```

tools/screenshot-xodball
```diff
@@ -23,6 +23,40 @@ const {
...
+const path = require('path');
+const loadProject = require('xod-fs/dist/load').loadProject;
+const XP = require('xod-project/dist');
+const {
+  SLOT_SIZE,
+  slotPositionToPixels,
+} = require('xod-client/dist/project/nodeLayout');
+const getBBox = require('xod-client/dist/editor/utils').getBoundingBoxSize;
+
+const getPatchBbox = (pathToProject, patchName) =>
```
evgenykochetkov: Shouldn't it be called `getPatchHeight` instead?

PR #1749, opened 2019-04-11
Fix screenshot tools

brusherru:
There is no issue.

1. `generate-tutorial-docs` outputs wrong paths in the `update-screenshots.sh`. Fixed.
2. `screenshot-xodball` hangs when tried to make a screenshot for the empty patch. Fixed.

### DIFF
--- a/tools/screenshot-xodball
+++ b/tools/screenshot-xodball
@@ -42,9 +42,13 @@ const getPatchHeight = (pathToProject, patchName) =>
         comments: XP.listComments(patch),
         links: XP.listLinks(patch),
       };
-      return (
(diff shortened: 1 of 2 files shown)


tools/screenshot-xodball
```diff
@@ -79,7 +83,10 @@ const patchboardTopLeft = {
     pathToXodball,
     patchBaseName,
     desiredScreenshotHeight
-  );
+  ).catch(err => {
+    process.stdout.write(`${err}\n`);
+    process.exit(0);
```
nkrkv:
```suggestion
    process.exit(1);
```

On error, we should exit with a non-zero code. Isn’t it?

PR #1777, opened 2019-05-13
Introduce nodes and C++ API for error handling

### LINKED ISSUE #1769: Implement errors
### Rationale

Currently, errors are signalized with `ERR` pulses. They require much boilerplate to pass an error further (big `any`’s with all `ERR` united) and do not carry any information about the error reason. Also, using pulses for errors makes it hard to express that a _value_ is malformed. We can improve the system by introducing a dedicated concept for errors.

### Errors concept

From time to time things go wrong. A sensor can refuse to respond, a server can become unavailable, a list index can fall out of range. Nodes should have a way to signalize that they fail so that other nodes which use their results can recover or show the error to a device user. XOD makes it possible with an error concept.

A node can raise an error to demonstrate that it has failed to evaluate. When the node raises an error all downstream nodes become “infected” and enter the error condition too. They won’t be evaluated as long as the original error is in effect. Their downstream nodes—in their turn—inherit the same error as well. The cascade goes on until the whole downstream subgraph of the original node is considered erroneous.

Errors are neither values nor pulses. They spread via links regardless of the link types and node evaluation policies. Errors are an overlay mechanism.

Any error risen has a code attached. The code loosely describes the error reason. It is a choice between one of the values defined by XOD. With the codes, someone who faces an error can better understand how to deal with it. Examples of the error codes are:

- `E16` (Bad Use) — the node input values are invalid
- `E32` (Hardware Fail) — general hardware communication problem
- `E48` (Network Fail) — network or internet communication failure

In contrast to errors, the error codes _are_ values of a type `Errcode` which exist specifically to handle errors.

> The error mechanism in XOD is something between exceptions in C++ or Python, and `Either` monads in Haskell or other functional programming languages. Unlike the “elder brothers” XOD does not carries custom error objects around: to keep resource usage at minimum, only short error codes are used.

During a transaction, if an erroneous node should be evaluated following the execution model rules, it _will be_ evaluated if it raises an error but did not inherit any. This way it can recover and “cure” the whole downstream subgraph bringing back the normal operation.

Also, a node may declare itself as an error catcher. In this case, it will be evaluated even if it inherits an error. Such nodes are used to recover from far upstream errors and take a specific action when something went wrong. In particular, the standard library provides the following error handlers:

- `if-error`
- `pulse-on-error`
- `error-code`

With these nodes, you can extract the code out to display it, retry an attempt, or freeze the device until a reboot.

At another end, to produce an error use the `error` node or the `raiseError` function in the C++ code.

### Acceptance criteria

- [x] The new `Errcode` built-in type is introduced; accepts literals of the `Exx` form
- [x] C++ API has functions `raiseError`, `getError<input_IN>(ctx)`
- [x] C++ supports `#pragma XOD catch enable`
- [x] The `if-error`, `pulse-on-error`, `error-code`, and `error` nodes are implemented

### How to implement

- Add a `uint8_t` error field to the `Node` structs
- Use the MSB to denote whether an error is own (0) or inherited (1)
- In `runTransaction`, for each node, update the error value by picking and merging upstream errors
- In `runTransaction`, expand the logic of dirtiness markup to mark downstream nodes dirty if the current node just got or lost the error state
- If a node inherits an error and does not declares itself as a catcher, skip `evaluate` and go on

### LINKED ISSUE #1771: Make watch node catch errors
### Rationale

The `watch` node is for observing what happens at different points of the program. In its current state, it will refuse to update the shown value if an upstream node raised an error. This can be addressed by turning the `watch` node into an error catcher which shows an error (e.g. `E16`) in case of an error.

### Acceptance criteria

- [x] In case of an error, `watch` sends debug value containing the error code prefixed with `E`
- [ ] In case of an error, `watch` re-raises the error for itself to keep the error-ish look in the IDE

evgenykochetkov:
Closes #1769
Closes #1771


packages/xod-arduino/src/Directives.re
```diff
@@ -122,6 +122,8 @@ let isNodeIdEnabled = code =>
   |. Code.lastPragmaEndis("nodeid")
   |. Endis.toBoolean(Code.doesReferSymbol("getNodeId", code));
 
+let doesRaiseErrors = Code.doesReferSymbol("raiseError");
```
nkrkv: For consistency should take `#pragma XOD error_raise` as the first source of truth with fallback to `doesReferSymbol`

See `isNodeIdEnabled` and `areTimeoutsEnabled`

packages/xod-arduino/src/Directives.re
```diff
@@ -138,6 +140,11 @@ let isDirtienessEnabled = (code, identifier) =>
        }
      );
 
+let isErrorCatchingEnabled = (code) =>
```
nkrkv: `doesCatchErrors` to match `doesRaiseErrors`?

packages/xod-project/src/utils.js
```diff
@@ -194,6 +194,14 @@ export const isValidCharLiteral = def(
   R.both(isLikeCharLiteral, R.complement(isAmong(unescapedCharLiterals)))
 );
 
+export const isValidErrcodeLiteral = def(
+  'isValidErrcodeLiteral :: String -> Boolean',
+  R.both(
+    R.test(/^E\d{1,3}$/g),
+    R.pipe(R.tail, n => parseInt(n, 10), n => n >= 0 && n <= 127)
```
nkrkv: Is `E010` a valid error literal? If so, we have a problem during conversion to C++. It will strip leading E and the value will be `010` which is octal `9` in C++ (leading zero denotes octal) which can be a big source of confusion.

Maybe it’s a good idea to forbid zero after `E` except the `E0` case

workspace/__lib__/xod/core/cast-to-string(errcode)/patch.cpp
```diff
@@ -0,0 +1,28 @@
...
+    auto errcode = getValue<input_IN>(ctx);
+
+    char* str = state->str;
+
+    do
+        *str++ = (char)(48 + (errcode % 10));
+    while (errcode /= 10);
+
+    *str++ = 'E';
+    *str = '\0';
```
nkrkv: Perhaps the standard [`itoa`](http://www.cplusplus.com/reference/cstdlib/itoa/) will be more straightforward here

packages/xod-arduino/platform/patchContext.tpl.cpp
```diff
@@ -3,6 +3,7 @@
 
 struct Node {
     State state;
+    uint8_t error;
```
nkrkv: Wastes a byte even on non-raisers

workspace/__lib__/xod/core/cast-to-string(errcode)/patch.cpp
evgenykochetkov: I don't think it's available on every platform

packages/xod-client/test/normalizeErrcode.spec.js
```diff
@@ -0,0 +1,33 @@
...
+import normalizeErrcode from '../src/utils/normalizeErrcode';
+
+describe('normalizeErrcode', () => {
+  const test = (from, to) => {
+    it(`${from} -> ${to}`, () =>
+      assert.strictEqual(normalizeErrcode(from), to));
+  };
+
+  test('complete gibberish', 'E0');
+  test('z42', 'E0');
```
nkrkv: May we check `E009` becomes `E9` here?

workspace/__lib__/xod/core/cast-to-string(errcode)/patch.cpp
nkrkv: OK, let it be

packages/xod-arduino/platform/runtime.cpp
```diff
@@ -143,6 +156,26 @@ bool isValidAnalogPort(uint8_t port) {
...
 
+/**
+  If node error's MSB is 1, it was propagated from the upstream.
+*/
+inline bool isPropagatedError(uint8_t err) {
+    return err & 0x80;
+}
+
+uint8_t getErrorCode(uint8_t err) {
+    return err & 0x7F;
```
nkrkv:
```suggestion
    return err & ~PROPAGATED_ERROR_MASK;
```

packages/xod-client/src/utils/normalizeErrcode.js
```diff
@@ -0,0 +1,11 @@
+import * as R from 'ramda';
+
+// :: DataValue -> DataValue
+export default R.pipe(
+  R.when(R.test(/^[eE].*/g), R.tail),
+  x => parseInt(x, 10),
+  R.defaultTo(0),
+  R.min(127),
+  R.max(0),
```
brusherru: It makes possible to bind "E0" to the "error" node. So when it receives a True into "ERR" pin — it does nothing... 🤔

PR #1784, opened 2019-05-30
Add ability to raise and capture errors in tabtests

### LINKED ISSUE #1772: As a tabtest user I want errors support so that my suite can be made comprehensive
Errors are the core concept of XOD which works above the regular values level. The tabtest framework should take them into account. Otherwise, my test can pass even if it actually failed because of an error. And if I want to test for an error, I have to create a dedicated patch which would extract an error with the `error-code` node. This is inconvenient boilerplate.

### User story

When I fill down value into a cell I should be allowed to enter an errcode literal (e.g. `E16`).

- If this is an output pin cell the test will pass only if the pin carries the error specified
- If this is an input pin cell, XOD will simulate receiving error specified from an upstream node
- If the pin type is indeed Errcode, the injection/capture will work as usual, against the real value

A point to state explicitly:

- If an output pin cell specifies some expected value of the type matching the pin type but receives an error, the whole tabtest line should fail: “Expected value 1234, got error E16”

### Acceptance criteria

- [x] Errors are handled as described above
- [x] Tabtests for `error`, `if-error`, `error-code`, and `pulse-on-error` are created

evgenykochetkov:
Closes #1772


packages/xod-tabtest/src/TabData.re
```diff
@@ -55,6 +57,11 @@ module Value = {
     | "-Inf" => Number(neg_infinity)
     | errcodeString when Re.test_(errcodeRegex, errcodeString) =>
       Errcode(tail(errcodeString) |> int_of_string)
+    | raisedErrorString when Re.test_(raisedErrorRegex, raisedErrorString) =>
+      Js.log(raisedErrorString);
+      let trimmedErrcode = String.sub(raisedErrorString, 2, String.length(raisedErrorString) - 3);
+      Js.log(trimmedErrcode);
```
nkrkv: `Js.log` forgotten

packages/xod-tabtest/src/TabData.re
```diff
@@ -13,12 +13,14 @@ module Value = {
...
     | Errcode(int)
     | Pulse(bool)
+    | RaisedError(int)
     | Invalid(string);
   let numberRegex = [%re {|/^[+-]?(?=.)*\d*(?:\.\d+)?$/|}];
   let approxNumberRegex = [%re {|/^[+-]?(?=.)*\d*(?:\.\d+)?~$/|}];
   let stringRegex = [%re {|/^".*"$/|}];
   let byteRegex = [%re {|/^[0-9a-f]{2}h|[0,1]{8}b|\d{1,3}d$/i|}];
   let errcodeRegex = [%re {|/^E\d{1,3}$/|}];
+  let raisedErrorRegex = [%re {|/^e\(\d{1,3}\)$/|}];
```
nkrkv: Not sure about the syntax. I feel it differs too much from the regular error code. Perhaps a better design would be trailing exclamation: `E42!` What do you think?

PR #1787, opened 2019-05-31
Visualize errors in the IDE

### LINKED ISSUE #1773: As a xoder I want errors to be visualized while debugging so that I can understand why everything stopped working
### Rationale

With the introduction of errors, we get another layer of data distribution across the graph which is invisible to a xoder who uses the debugger. [Pete](https://github.com/xodio/xod/wiki/Pete) would like to see when errors appear and which nodes they affect.

### User story

- I start a debug session, it works as usual
- A node in the middle of the graph raises an error
- The node immediately gets a redish look
- Its downstream nodes affected by the error recursively get the same look
- All links causing the error cascade gets redish look too
- When the error disappears, the look of nodes and links animates back to normal in 100 ms

The animation on error disappear is necessary to spot errors which occur in short time spans (e.g, a single transaction)

### Acceptance criteria

- [x] The errors visualization works as described

### How to implement

- When a node raises a _new_ error, send a debug protocol line with the node ID and error code
- When a node clears own error, send the same line with error code `E00` (no error)
- Do not send information about inherited errors to save serial bandwidth; the inheritance should be recomputed in IDE

brusherru:
It closes #1773 

Here is my example project to test all possible cases (I hope so):
[visualize-errors.xodball.zip](https://github.com/xodio/xod/files/3242103/visualize-errors.xodball.zip)
- `0-test` contains identity nodes, jumpers, buses, nested error raisers and etc
- `1-defers` contains a loop with `defer` node and `if-error` catcher

### DIFF
--- a/packages/xod-client/src/project/selectors.js
+++ b/packages/xod-client/src/project/selectors.js
@@ -32,6 +32,11 @@ import {
 } from '../project/utils';
 import { setPxPosition, setPxSize } from './pxDimensions';
 
+import {
+  getInteractiveNodeErrorCodesForCurrentChunk,
+  getPinsAffectedByErrorRaisersForCurrentChunk,
+} from '../debugger/selectors';
+
 import { createMemoizedSelector } from '../utils/selectorTools';
 
 export const getProject = R.prop('project');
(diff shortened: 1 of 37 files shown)


brusherru:
@nkrkv rebased

packages/xod-project/src/Traversing.rei
```diff
@@ -0,0 +1 @@
+let listUpstreamPinsToNiix: (list((Pin.t, Node.t)), Patch.t, Project.t) => list((Pin.t, Node.t));
```
nkrkv: Functions in `*.rei` must have a documentation comment

packages/xod-project/src/Traversing.re
```diff
@@ -0,0 +1,145 @@
...
+  Utility functions
+==============================================================================================================
+*/
+
+let isCompositionPatch = patch => (
+  !Patch.isNotImplementedInXod(patch) &&
+  !Patch.isTerminal(patch)
+);
+
+let getUpstreamPinPairFromInputPin: (Pin.t, Node.t, Patch.t, Project.t) => option(pinWithOwnerNode) =
```
evgenykochetkov: I suggest to get rid of "PinPair" term, since it's called "PinWithOwnerNode" now

packages/xod-client/src/project/selectors.js
```diff
@@ -304,15 +436,28 @@ export const getRenderableNodes = createMemoizedSelector(
...
+    R.equals,
+    R.equals,
+    R.equals,
+    R.equals,
+    R.equals,
+    R.equals,
+    R.equals,
+    R.equals,
   ],
-  [R.equals, R.equals, R.equals, R.equals, R.equals, R.equals],
```
evgenykochetkov: 😄

brusherru: Yep 🙈

Issue #1805, opened 2019-06-11
I propose to build AppImage and drop support of rpm and debs


knopki:
Currently, we build `rpm`s and `deb`s for Linux. And this is broken in so many way:
* it's not normal when rpm and deb packaging on the developer side, this should be done in distributions by maintainers but it's almost impossible to include `electron` application into the popular linux distributions
* updates can be shipped only by creating repositories with all that huge pain with building, signing, forcing user to add repositories configuration to their systems
* `rpm` and `deb` targets are not first class citizens of `electron-builder`
* we can't create file/url associations (see previous paragraph)
* desktop entry for program are broken and can't be fixed because of `electron-builder`:
  * spaces in the path - breaks some utilities
  * path in the quotes - breaks more utilities (hello i3 users)
  * %U as argument just make program unusable on Mate/KDE
  * %U as argument force file paths in url style (`file://...`) - we doesn't support this

All these problems are likely to never be fixed.

Better alternatives: `flatpak` and `snap`. This should seize all major distributions. Also, build and host on flathub.org and snapcraft.io marketplaces -> more users. But, `electron-builder` can't do flatpak (required to use another build procedure). `snap` is also not easy peasy. So, `flatpak` and `snap` look more like a bright future.

Not perfect at all but realistic and practical solution is `AppImage`. It's like 'portable' one-file univarsal version with some integration with the system.
Pros:
* should be more understandable by users (it's just click-and-run file!)
* we can enable updates for linux that are consistent with win and mac
* user rights are sufficient for installing and upgrading
* all problems with desktop entry file are gone (created on first run with user want integration)
* mime types created for all xod file types
* file associations created and just works
Cons:
* needs much testing
* require some transition period
* (may be) require changes in CI/CD
* (may be) require some changes of `releases` endpoint
* require changes in site's download page

For testing purpose I changed root `package.json` with this build section:
```js
  "build": {
    "appId": "io.xod.ide",
    "productName": "XOD IDE",
    "buildDependenciesFromSource": true,
    "extraResources": [
      {
        "from": "arduino-cli",
        "to": "arduino-cli"
      },
      {
        "from": "arduino-cli.exe",
        "to": "arduino-cli.exe"
      },
      {
        "from": "src-babel/workspace",
        "to": "workspace"
      }
    ],
    "linux": {
      "executableName": "xod-ide",
      "target": "AppImage",
      "category": "Development"
    },
    "appImage": {
[...]

PR #1811, opened 2019-06-24
Upgrade electron from 1.8.8 to 4.2.4

evgenykochetkov:
There are not much [breaking changes from 1.x to 4.x](https://github.com/electron/electron/blob/master/docs/api/breaking-changes.md#planned-breaking-api-changes-40), so let's return from the stone age to (relatively) modern times ;)

Also, this will make dealing with #1805 a bit easier.


Cesar-S:
This may be what caused the increase of the generated code?  efficiency has been lost

 [Big code ](https://forum.xod.io/t/problem-with-code-generated-too-big-in-0-30-1-vs-0-29-1/2456?u=cesars)

nkrkv:
No, it’s definitely not an Electron issue but is related to the new errors mechanism. Will take a closer look.

PR #1812, opened 2019-06-24
Fix workspace creation and show a bundled project on opening not a XOD file

### LINKED ISSUE #1804: Workspace not created correctly when you open not a xod file
### Environment

* Scope: desktop IDE
* Version: 0.29.1
* OS: Fedora 29

### Steps to reproduce

1.  Remove current workspace
2.  Try to open some broken/nonexistent file as the first argument of `xod-client-electron`
3.  XOD IDE starts
4. Normal error `Tried to open not a xod file` happens

### Expected behavior

A workspace created correctly.

### Actual behavior

A workspace created as a directory with `.xodworkspace` in it, but without `__packages__`. You can't upload to a device because `arduino-cli` stuff not initialized (empty list of board models, can't update with button `Update`). And there is no direction for a user how to fix this situation.

brusherru:
It fixes #1804

### DIFF
--- a/packages/xod-client-electron/src/app/workspaceActions.js
+++ b/packages/xod-client-electron/src/app/workspaceActions.js
@@ -1,7 +1,6 @@
 import * as R from 'ramda';
 import EventEmitter from 'events';
 import path from 'path';
-import { Maybe } from 'ramda-fantasy';
 
 import * as XP from 'xod-project';
 import {
@@ -21,7 +20,7 @@ import {
   rejectWithCode,
   allPromises,
   notNil,
-  explodeMaybe,
+  foldMaybeWith,
 } from 'xod-func-tools';
 
 import * as settings from './settings';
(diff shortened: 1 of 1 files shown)


packages/xod-client-electron/src/app/workspaceActions.js
```diff
@@ -261,30 +272,22 @@ export const onIDELaunch = R.curry(
...
-            projectName
-          ).then(() => {
-            // TODO: better pass it along the project with REQUEST_SHOW_PROJECT event
-            if (!isFirstLaunch) {
-              send(EVENTS.PAN_TO_CENTER);
-            }
-          });
+        const maybeProjectPathToOpen = getFileToOpen();
+
+        if (Maybe.isNothing(maybeProjectPathToOpen)) {
```
evgenykochetkov: I think it's better to use `foldMaybeWith` from `xod-fun-tools` instead of manually checking for nothingness and then exploding

brusherru: Good idea! 👍 
Done!

PR #1816, opened 2019-06-28
Respect serial.disableRTS option from boards.txt when opening serialport for debugging

evgenykochetkov:
This fixes problems with debugging Arduino Leonardo and some other boards.


nkrkv:
Merge?

PR #1818, opened 2019-07-09
Add C++ API to raise errors separately for each output

evgenykochetkov:
In addition to existing `raiseError(ctx)`, there is now a `raiseError<output_OUT>(ctx)`, which raises an error only for a specific output.

Also,
- uses dirtiness to spread errors
  - this allows to stop storing `prevCaughtError` byte for catchers
  - makes nodes like `pulse-on-error` react to _each_ error even when they were raised consequently without "normal" values in between
- makes tabtests strictly check that no error was raised when a "normal" value was specified
 - errors from pulse outputs are cleaned after each transaction to be consistent with how "normal" pulses behave

### DIFF
--- a/packages/xod-arduino/platform/patchContext.tpl.cpp
+++ b/packages/xod-arduino/platform/patchContext.tpl.cpp
@@ -4,7 +4,15 @@
 struct Node {
     State state;
   {{#if raisesErrors}}
-    bool hasOwnError;
+    union {
+        struct {
+          {{#each outputs}}
+            bool outputHasError_{{ pinKey }} : 1;
+          {{/each}}
+        };
+
+        ErrorFlags errorFlags;
+    };
   {{/if}}
   {{#if usesTimeouts}}
     TimeMs timeoutAt;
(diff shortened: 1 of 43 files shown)


workspace/blink/__fixtures__/arduino.cpp
```diff
@@ -1615,6 +1614,15 @@ void runTransaction() {
     node_4.dirtyFlags = 0;
     node_5.dirtyFlags = 0;
     node_6.dirtyFlags = 0;
+
+    // Сlean errors from pulse outputs
+    if (node_6.outputHasError_DONE) {
+      node_6.outputHasError_DONE = false;
+#if defined(XOD_DEBUG) || defined(XOD_SIMULATION)
+      detail::printErrorToDebugSerial(6, node_6.errorFlags);
+#endif
```
nkrkv: These `#if ... printError #endif` are all over the source. They ugly a bit. What if you’ll guard the *body* of  `detail::printErrorToDebugSerial` and omit the checks at all use points?

If we’re not in simulation, it would be an empty function call which, I’m pretty sure, will be thrown away by C++ optimizer.

A bulletproof solution is creating a macro like `XOD_TRACE`, but for error purposes. For example, `XOD_DEBUG_ERROR(6, node_6.errorFlags)`. It can be expanded to no-op if we’re not simulating.

packages/xod-arduino/platform/patchContext.tpl.cpp
```diff
@@ -147,6 +147,20 @@ template<> void raiseError<output_{{ pinKey }}>(Context ctx) {
...
+void raiseError(Context ctx) {
+  {{#each outputs}}
+    ctx->_node->outputHasError_{{ pinKey }} = true;
+    {{#if isDirtyable}}
+    ctx->_node->isOutputDirty_{{ pinKey }} = true;
+    {{/if}}
+  {{/each}}
+
+#if defined(XOD_DEBUG) || defined(XOD_SIMULATION)
+    detail::printErrorToDebugSerial(ctx->_nodeId, ctx->_node->errorFlags);
```
nkrkv: Not necessary?

workspace/test/test-core/test-error-propagation-through-defer-nodes/patch.test.tsv
```diff
@@ -3,4 +3,5 @@ __time(ms)	ERR	OUT1	OUT2	OUT3	OUT4	OUT5
 5	true	false	pulse	""	0	0
 6	true	error	error	error	error	error
 15	false	error	error	error	error	error
-16	false	false	no-pulse	""	0	0
+16	false	false	error	""	0	0
+			// TODO
```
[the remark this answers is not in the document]
evgenykochetkov: That's before "make errors for pulse outputs behave like pulses" decision. It's outdated now.

workspace/blink/__fixtures__/arduino.cpp
evgenykochetkov: Moved `#if ... printError #endif` into the body of `detail::printErrorToDebugSerial`.

PR #1825, opened 2019-07-18
Fix and add nodes that works with errored pulses

brusherru:
There is no issue.

In short, not all nodes that handle errors (`if-error`, `has-error`) worked correctly with a pulse type. Even the `error(pulse)` node did not work correctly.

Here is an example project:  [test-pulse-errors.xodball.zip](https://github.com/xodio/xod/files/3406640/test-pulse-errors.xodball.zip)

### DIFF
--- a/workspace/__lib__/xod/core/error(pulse)/patch.cpp
+++ b/workspace/__lib__/xod/core/error(pulse)/patch.cpp
@@ -6,10 +6,11 @@ struct State {
 {{ GENERATED_CODE }}
 
 void evaluate(Context ctx) {
-    if (getValue<input_ERR>(ctx)) {
+    if (!isInputDirty<input_IN>(ctx))
+        return;
+
+    if (getValue<input_ERR>(ctx))
         raiseError<output_OUT>(ctx);
-    } else {
-        if (isInputDirty<input_IN>(ctx))
-            emitValue<output_OUT>(ctx, 1);
-    }
+    else
+        emitValue<output_OUT>(ctx, 1);
 }
--- a/workspace/__lib__/xod/core/has-error(pulse)/patch.cpp
+++ b/workspace/__lib__/xod/core/has-error(pulse)/patch.cpp
@@ -1,0 +1,22 @@
+
+#pragma XOD error_catch enable
+
+struct State {
+};
+
+{{ GENERATED_CODE }}
+
+void evaluate(Context ctx) {
+    if (isInputDirty<input_IN>(ctx) && getError<input_IN>(ctx)) {
+        emitValue<output_OUT>(ctx, true);
+        // Manually trigger a reevaluation by setting timeout.
+        // The reevaluation is needed to set the output to False
+        // even when no pulse is received.
+        setTimeout(ctx, 0);
+        return;
+    }
+
+    if (isTimedOut(ctx)) {
+        emitValue<output_OUT>(ctx, false);
+    }
+}
--- a/workspace/__lib__/xod/core/has-error(pulse)/patch.xodp
+++ b/workspace/__lib__/xod/core/has-error(pulse)/patch.xodp
@@ -1,0 +1,32 @@
+{
+  "description": "Outputs `True` while the upstream node is in the error state.",
+  "nodes": [
+    {
+      "id": "BJDwgk0-r",
+      "position": {
+        "units": "slots",
+        "x": 0,
+        "y": 0
+      },
+      "type": "xod/patch-nodes/input-pulse"
+    },
+    {
+      "id": "S1WPi66aWH",
+      "position": {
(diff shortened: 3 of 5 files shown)


workspace/__lib__/xod/core/has-error(pulse)/patch.cpp
```diff
@@ -0,0 +1,20 @@
...
+
+struct State {
+};
+
+{{ GENERATED_CODE }}
+
+void evaluate(Context ctx) {
+    if (isInputDirty<input_IN>(ctx) && getError<input_IN>(ctx)) {
+        emitValue<output_OUT>(ctx, true);
+        setTimeout(ctx, 1);
```
nkrkv:
```suggestion
        setTimeout(ctx, 0);
```

Otherwise, 1000 us magic duration is introduced

workspace/__lib__/xod/core/has-error(pulse)/patch.cpp
```diff
@@ -0,0 +1,20 @@
...
+        return;
+    }
+
+    if (isTimedOut(ctx)) {
+        emitValue<output_OUT>(ctx, false);
+        clearTimeout(ctx);
```
nkrkv: No need to `clearTimeout` if we currently in `isTimedOut`

workspace/__lib__/xod/core/error(pulse)/patch.cpp
```diff
@@ -6,10 +6,11 @@ struct State {
...
-        raiseError<output_OUT>(ctx);
-    } else {
-        if (isInputDirty<input_IN>(ctx))
-            emitValue<output_OUT>(ctx, 1);
+    if (isInputDirty<input_IN>(ctx)) {
+        if (getValue<input_ERR>(ctx)) {
+            raiseError<output_OUT>(ctx);
+            return;
+        }
+        emitValue<output_OUT>(ctx, 1);
```
nkrkv: IMO can be less nested, more readable

```cpp
if (!isInputDirty<input_IN>(ctx))
    return;

if (getValue<input_ERR>(ctx))
    raiseError<output_OUT>(ctx);
else
    emitValue<output_OUT>(ctx, 1);
```

workspace/__lib__/xod/core/has-error(pulse)/patch.xodp
```diff
@@ -0,0 +1,31 @@
+{
+  "nodes": [
```
nkrkv: A description is missing

workspace/__lib__/xod/core/if-error(pulse)/patch.cpp
```diff
@@ -0,0 +1,29 @@
...
+        raiseError<output_OUT>(ctx);
+        return;
+    }
+
+    if (isInputDirty<input_IN>(ctx)) {
+        if (getError<input_IN>(ctx)) {
+            if (defDirty) {
+                emitValue<output_OUT>(ctx, 1);
+            }
+            return;
```
nkrkv: No need to `return` here

workspace/__lib__/xod/core/if-error(pulse)/patch.cpp
```diff
@@ -0,0 +1,29 @@
...
+    auto defDirty = isInputDirty<input_DEF>(ctx);
+    auto defError = getError<input_DEF>(ctx);
+
+    if (defDirty && defError) {
+        // "DEF" input should not contain an error — reraise it
+        raiseError<output_OUT>(ctx);
+        return;
+    }
+
+    if (isInputDirty<input_IN>(ctx)) {
```
nkrkv: The condition could be inversed to make 1-level less nesting

workspace/__lib__/xod/core/if-error(pulse)/patch.xodp
```diff
@@ -0,0 +1,41 @@
+{
+  "nodes": [
```
nkrkv: A description is missing

workspace/__lib__/xod/core/has-error(pulse)/patch.cpp
```diff
@@ -0,0 +1,19 @@
...
+
+struct State {
+};
+
+{{ GENERATED_CODE }}
+
+void evaluate(Context ctx) {
+    if (isInputDirty<input_IN>(ctx) && getError<input_IN>(ctx)) {
+        emitValue<output_OUT>(ctx, true);
+        setTimeout(ctx, 0);
```
evgenykochetkov: A comment explaining that we need to manually trigger a reevaluation because pulse errors are cleared at the end of transaction would be nice. There is no way I (and probably anyone else) could explain the reason this line is here in a couple of months.

PR #1838, opened 2019-08-22
Add `xod/json` library

evgenykochetkov:


workspace/__lib__/xod/json/parse-json/patch.xodp
```diff
@@ -0,0 +1,61 @@
+{
+  "description": "Pasres a stream of characters that represent a JSON document.",
```
knopki: Can't pasre this :)

workspace/__lib__/xod/json/string-value/patch.cpp
```diff
@@ -0,0 +1,47 @@
...
+};
+
+{{ GENERATED_CODE }}
+
+void evaluate(Context ctx) {
+    auto state = getState(ctx);
+    auto parsed = getValue<input_IN1>(ctx);
+
+    if (isSettingUp()) {
+        // save initial cap to ignore possiple input changes during program execution
```
knopki: inpossiplu

workspace/__lib__/xod/json/parse-json/patch.cpp
```diff
@@ -0,0 +1,446 @@
...
+#define STACK_OBJECT             0
+#define STACK_ARRAY              1
+#define STACK_KEY                2
+#define STACK_STRING             3
+
+bool isFirstCharInNumber(char c) {
+    return (c >= '0' && c <= '9') || c == '-';
+}
+
+// We don't use `strchr` because buffer we search in is not \0 - terminated
```
nkrkv: It’s [`memchr`](http://www.cplusplus.com/reference/cstring/memchr/)

workspace/__lib__/xod/json/parse-json/patch.cpp
```diff
@@ -0,0 +1,446 @@
...
+        return true;
+      }
+    }
+    return false;
+}
+
+struct ContextObject;
+
+// we only use buffer to validate numbers, booleans and `null`
+#define BUFFER_MAX_LENGTH 64
```
nkrkv: In modern C++ prefer constexpr

```suggestion
constexpr size_t BUFFER_MAX_LENGTH = 64;
```

And are we really need to keep 64 bytes? The number we can precisely keep is ~7 digits. A 15-digit number is huge. I think we could limit ourselves to 16 chars.

workspace/__lib__/xod/json/parse-json/patch.cpp
```diff
@@ -0,0 +1,446 @@
+
+using xod::xod__json__parsed_json_stream::ParserState;
+
+#define STACK_OBJECT             0
+#define STACK_ARRAY              1
+#define STACK_KEY                2
+#define STACK_STRING             3
```
nkrkv: Looks like an enum. Let it be enum

workspace/__lib__/xod/json/parse-json/patch.cpp
```diff
@@ -0,0 +1,446 @@
...
+
+struct ContextObject;
+
+// we only use buffer to validate numbers, booleans and `null`
+#define BUFFER_MAX_LENGTH 64
+
+class JsonParser {
+  private:
+    ParserState state;
+    int stack[20];
```
nkrkv: Memory waste? Might be `uint8_t` array.

workspace/__lib__/xod/json/parse-json/patch.cpp
```diff
@@ -0,0 +1,438 @@
+
+using xod::json_parser::ParserState;
+
+enum class Stack : uint8_t {
```
nkrkv: `StackItem` would be a more correct name

PR #1839, opened 2019-08-27
Fix tabtests runner exit code

evgenykochetkov:
Failing tests were ignored because of the loop.


evgenykochetkov:
(Yes, now `test-cpp` is failing because we had an undetected broken test in master 🙈)

PR #1851, opened 2019-09-12
Extract hc-sr04 into separate library

### LINKED ISSUE #1792: Extract HC-SR04 nodes to new library
The HC-SR04 sensors are not vendor-neutral enough to stay in `xod/common-hardware`. Let’s extract their nodes to a new `xod-dev/hc-sr04` library.

It’s a good moment to start following the [Hardware library guidelines](https://github.com/xodio/xod/wiki/Hardware-Library-Guidelines). That is, make a device node, action node (`ping`), and a throttled quickstart node (`hc-sr04-range-meter`).

We currently have `hc-sr04-ultrasonic-range` and `hc-sr04-ultrasonic-time`. I doubt the latter has wide usage. Let’s leave only the first one and make it the quickstart node. The conversion between `ping`’s time domain and the distance domain might be encapsulated inside the quickstart node. A good idea would be placing `standard-sound-speed` constant node to `xod/units` which is the sound speed in dry air at 20 °C.

### Prerequisites

- #1778 

### Acceptance criteria

Provide some steps that an end-user would be able to perform when the feature could be considered as implemented. E.g.:

- [x] The `hc-sr04-*` nodes are extracted to `xod-dev/hc-sr04`
- [x] The library follows the hardware library guidelines
- [x] The current nodes are deprecated
- [x] The links in https://xod.io/docs/reference/supported-hardware/ point to the new nodes
- [x] The quickstart node is throttled
- [x] The quickstart node is called `hc-sr04-range-meter` and outputs the distance in meters
- [x] `xod/units` got `standard-sound-speed` constant

brusherru:
With this PR https://github.com/xodio/xod-docs/pull/141 it will close #1792

### DIFF
--- a/workspace/__lib__/xod-dev/hc-sr04/hc-sr04-range-meter/patch.xodp
+++ b/workspace/__lib__/xod-dev/hc-sr04/hc-sr04-range-meter/patch.xodp
@@ -1,0 +1,237 @@
+{
+  "links": [
+    {
+      "id": "ByTGWCvIr",
+      "input": {
+        "nodeId": "BkeSs3988B",
+        "pinKey": "__in__"
+      },
+      "output": {
+        "nodeId": "rkuJ-0P8B",
+        "pinKey": "Bkimz58LS"
+      }
+    },
+    {
+      "id": "HJSJZ0PLB",
+      "input": {
+        "nodeId": "By7JWCPUS",
+        "pinKey": "SJeMLBsIUH"
+      },
+      "output": {
+        "nodeId": "H1gBkhc88H",
+        "pinKey": "__out__"
+      }
+    },
+    {
+      "id": "HkY7-0vLS",
+      "input": {
+        "nodeId": "HJboW-CPUB",
+        "pinKey": "B1GfLR_SPk-"
+      },
+      "output": {
+        "nodeId": "SkxiZZAvLH",
+        "pinKey": "BkqLCOSw1W"
+      }
+    },
+    {
+      "id": "HySpgCw8B",
+      "input": {
+        "nodeId": "HkMaeCw8r",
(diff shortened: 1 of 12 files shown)


workspace/__lib__/xod-dev/hc-sr04/hc-sr04-range-meter/patch.xodp
```diff
@@ -0,0 +1,237 @@
...
+      "id": "HJboW-CPUB",
+      "position": {
+        "units": "slots",
+        "x": -1,
+        "y": 3
+      },
+      "type": "xod/core/multiply"
+    },
+    {
+      "description": "Last measured distance in meters.",
```
nkrkv:
```suggestion
      "description": "Last measured distance in meters. Will be set to +Inf if the sensor gets no echo back",
```

What about the no-echo case?

workspace/__lib__/xod-dev/hc-sr04/hc-sr04-range-meter/patch.xodp
```diff
@@ -0,0 +1,237 @@
...
+        "x": -2,
+        "y": 2
+      },
+      "type": "xod/core/divide"
+    },
+    {
+      "boundLiterals": {
+        "__out__": "Continuously"
+      },
+      "description": "Triggers new ultrasonic ping which would result in the output update once sound echo will be captured. Pulses coming within 60 ms window after last one are ignored to fight sensor’s PCB resonance.",
```
nkrkv:
```suggestion
      "description": "Triggers new ultrasonic ping which will result in the output update once sound echo is captured. Pulses coming within 60 ms window after the last one are ignored to fight sensor’s PCB resonance.",
```

PR #1858, opened 2019-09-19
Change error clear conditions 

### LINKED ISSUE #1834: Change error clear conditions
### Rationale

Currently, we have error reset behavior which is impractical in real tasks. Two issues exist:

1. Errors on scalar values are cleared automatically before node evaluation when it is given a chance to recover. This contradicts to the documentation and intent.
2. Errors on pulse links get cleared at the end of the transaction. This introduces counter-intuitive effects when the sub-graph is recovered almost immediately after an error and works “fine” even if we can’t trust the data because there was the error

Let’s get rid of both issues.

### Acceptance criteria

- [x] Value-type outputs keep errors until a node’s `evaluate` emits a new non-error value explicitly
- [x] Pulse-type outputs _do_ lose errors right before recovering
- [x] Pulse-type outputs _do not_ lose errors at the end of a transaction
- [x] Docs are updated to describe the updated pulse errors behavior

evgenykochetkov:
Value-type outputs keep errors until a node’s `evaluate` emits a new non-error value explicitly, not before each evaluate

Closes part of #1834


workspace/test/test-core/test-defer-pulse/patch.test.tsv
```diff
@@ -0,0 +1,13 @@
+IN	OUT
+no-pulse	no-pulse
+
+pulse	no-pulse
+no-pulse	pulse
+
+error	no-pulse
+no-pulse	error
+
+pulse	error	// is this okay?
```
nkrkv: If we’d talk about pulses as of type with values now (pulse) / past (no-pulse) / error (error), the line above the current (no-pulse) sets it to “past”. So, I’d expect `no-pulse` instead of the `error` here.

Does it conflict with any other reasoning?

evgenykochetkov:
- ~~pulse outputs do lose errors right before `evaluate`, not at the end of transaction~~
- ~~when pulse input loses error, downstream nodes reevaluate~~

Had to revert this part because of the discovered drawbacks.

Now this PR is only about value types.

packages/xod-arduino/platform/program.tpl.cpp
```diff
@@ -183,6 +183,16 @@ void runTransaction() {
             g_transaction.node_{{id}}_isNodeDirty = false;
             detail::clearTimeout(&node_{{ id }});
         }
+
+        if (node_{{ id }}.errorFlags) {
```
```suggestion
        // propagate the error hold by the defer node
        if (node_{{ id }}.errorFlags) {
```

packages/xod-arduino/platform/program.tpl.cpp
```diff
@@ -326,6 +326,18 @@ void runTransaction() {
             {{/if}}
           {{/each}}
         }
+
+      {{#if patch.raisesErrors}}
+        if (node_{{ id }}.errorFlags) {
```
```suggestion
        // propagate errors hold by the node outputs
        if (node_{{ id }}.errorFlags) {
```

packages/xod-arduino/platform/program.tpl.cpp
```diff
@@ -286,9 +285,15 @@ void runTransaction() {
           {{/eachDirtyablePin}}
 
           {{#if patch.raisesErrors}}
-            ErrorFlags previousErrorFlags = node_{{ id }}.errorFlags;
-            // give the node a chance to recover from it's own previous errors
-            node_{{ id }}.errorFlags = 0;
+            // TODO: a copy constructor to make this less ugly?
+            {{ns patch }}::NodeErrors previousErrors = { .flags=node_{{ id }}.errors.flags };
```
nkrkv: Yes, please. But aren’t structs have a default copy-constructor which copies data over? I mean, to me it seems like an equivalent to trivial:

```cpp
{{ns patch }}::NodeErrors previousErrors = node_{{ id }}.errors;
```

workspace/test/test-core/test-defer-pulse/patch.test.tsv
```diff
@@ -7,7 +7,7 @@ no-pulse	pulse
 error	no-pulse
 no-pulse	error
 
-pulse	no-pulse
+pulse	error	// it didn't clear because nothing was evaluated
```
nkrkv: Sounds sane :ok_hand:

packages/xod-arduino/platform/program.tpl.cpp
```diff
@@ -299,8 +304,28 @@ void runTransaction() {
           {{/eachDirtyablePin}}
 
           {{#if patch.raisesErrors}}
-            if (previousErrorFlags != node_{{ id }}.errorFlags) {
-                detail::printErrorToDebugSerial({{ id }}, node_{{ id }}.errorFlags);
+            if (previousErrors.flags != node_{{ id }}.errors.flags) {
+                detail::printErrorToDebugSerial({{ id }}, node_{{ id }}.errors.flags);
+
+              {{#each outputs}}
+                if (node_{{ ../id }}.errors.output_{{ pinKey }} != previousErrors.output_{{ pinKey }}) {
```
nkrkv: Suggest adding a comment here which will go to the generated code

packages/xod-arduino/platform/program.tpl.cpp
```diff
@@ -299,8 +304,28 @@ void runTransaction() {
...
+                    g_transaction.node_{{this}}_isNodeDirty = true;
+                  {{/each}}
+                }
+              {{/each}}
+
+              {{!--
+                // if a pulse output was cleared from error, mark downstream nodes as dirty
+                // (no matter if a pulse was emitted or not)
+              --}}
+              {{#eachPulseOutput outputs}}
```
nkrkv: Same here. Just promote the Handlebars comment to a regular one

PR #1861, opened 2019-09-25
Do not highlight nodes as errored if it recovers an error inside

brusherru:
There is no issue.


brusherru:
Can't do it yet without the bug described by @evgenykochetkov :(
It needs a deep refactoring, so I'll close this PR until the best times.

PR #1870, opened 2019-10-03
Tweak `xod/debug/stream-string` node

brusherru:
Old implementation works only once on the boot.
This one will stream the string on the change. So now this node can be used with `tweak-string` or any other dynamically changed input value.

Example: 
image

❗️ 
Probably, it should be moved to `xod/stream` and should not be a utility node anymore.
What do you think?


brusherru:
👌

PR #1873, opened 2019-10-04
Improve error propagation through defer nodes

evgenykochetkov:
Covers a few edge-cases such as a `defer` node surrounded by closed `gate`s

PR #1878, opened 2019-10-07
Update changelog for upcoming 0.31.1

evgenykochetkov:
### Bug fixes

* [core] Fix a bug that caused compilation error if a defer node was used in a feedback loop without an upstream error raiser (#1871)
* [core] Fix updating of package indexes and bundle the latest one (#1869)
* [nodes] Fix quickstart node for 20x4 I2C text LCD in `xod-dev/text-lcd` library (#1875)
* [nodes] Fix wrong types in `xod-dev/dht` library (#1877)
* [core] Fix errors not propagating through defer nodes in some edge cases (#1873)
* [nodes] Fix debounce time in `xod/common-hardware/button` (#1880)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,17 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+## Not yet released
+
+### Bug fixes
+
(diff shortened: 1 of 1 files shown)


brusherru:
Add this one https://github.com/xodio/xod/pull/1880 after it will be merged, per favore.

evgenykochetkov:
@brusherru, done

PR #1904, opened 2019-11-05
Derive project name from filename

### LINKED ISSUE #1897: Inherit project name from filename
Currently, the project name is totally independent of the filename/dirname where it is saved on FS. This is a good thing because requirements between the FS name and XOD name differ. Moreover, a xoder may intentionally make a fork or a backup copy of a project with requirement to keep the original project name.

From another point of view, in most cases, there’s a direct correlation between the filename and the eventual project name. So, asking to enter similar things twice is suboptimal UX. XOD could assist by mirroring FS name to project name in trivial cases.

### User story

- I start XOD IDE with the untitled project open
- I save it giving the filename `foo.xodball`
- The project name is set to `foo` automatically
- I save the same project as `bar.xodball`
- An info message is shown telling that it is a good idea to update the project name in Preferences

### Name deduction

XOD project name is only allowed to have lower case words separated with hyphens. So, a projection from FS name to XOD name is necessary. Let’s call it `deriveProjectNameFromFilename` It can be done as follows:

- `foo.xodball` → `foo`
- `FooBar.xodball` → `foobar`
- `FooBar Baz.xodball` → `foobar-baz`
- `FooBar Baz 2.xodball` → `foobar-baz-2`
- `_FooBar Baz (2).xodball` → `foobar-baz-2`
- `_FooBar Baz (2).whatever.infix.xodball` → `foobar-baz-2`

### Acceptance criteria

- [x] The basic scenario of saving when the project name is not set works as described
- [x] The project name is derived from the dirname if the project is saved in the multifile mode
- [x] I can Save As… and enter a new filename so that `deriveProjectNameFromFilename(newFilename)` does not match the current project name. A persistent info message¹ will show up.

¹
```markdown
**Project name now differs from filename**
The project name is left to be `foobar`. It might be a good idea to update it to match
the new filename `foobar-copy-1` if you indeed continue to evolve it as a new project.
Hit Edit → Project Preferences to update the settings.
```

brusherru:
It closes #1897

### DIFF
--- a/packages/xod-client-electron/src/view/actions.js
+++ b/packages/xod-client-electron/src/view/actions.js
@@ -1,6 +1,8 @@
 import * as R from 'ramda';
 import { ipcRenderer } from 'electron';
+import { basename } from 'path';
 
+import { getProjectName } from 'xod-project';
 import {
   addProcess,
   progressProcess,
@@ -8,8 +10,13 @@ import {
   deleteProcess,
   failProcess,
   addConfirmation,
+  addNotification,
   addError,
   SAVE_ALL,
+  getProject,
+  getLastSavedProject,
+  updateProjectMeta,
+  deriveProjectName,
 } from 'xod-client';
 import * as EVENTS from '../shared/events';
 import * as MESSAGES from '../shared/messages';
@@ -71,7 +78,7 @@ const createAsyncAction = ({
   silent = false,
   onComplete = R.always(undefined),
   onError = R.always(undefined),
-}) => data => dispatch => {
+}) => data => (dispatch, getState) => {
   let processId = null;
 
   if (!silent) {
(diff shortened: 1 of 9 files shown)


packages/xod-client-electron/src/view/actions.js
```diff
@@ -150,6 +156,37 @@ export const saveAll = createAsyncAction({
     complete: MESSAGES.SAVE_ALL_SUCCEED,
     error: MESSAGES.SAVE_ALL_FAILED,
   },
+  onComplete: (payload, dispatch, getState) => {
+    if (payload.updateProjectPath) {
```
nkrkv: Better use early exit to remove one level of nesting:

```js
if (!payload.updateProjectPath) {
  return;
}
```

It took 5 mins for me to understand how this works because the name is misleading a bit. It would be more intuitive if named `shouldUpdateProjectPath`.

PR #1906, opened 2019-11-12
Fix arduino-cli and spaces in username on Windows

### LINKED ISSUE #1723: Path to arduino-cli config should be enquoted
See: https://forum.xod.io/t/problems-compiling/2020

### Environment

* Scope: desktop IDE
* Version: 0.28.1
* OS: Windows 10

### Steps to reproduce

1.  Start XOD on behalf of a user with spaces in the name (ex: `C:\Users\Foo Bar Qux`)
2.  Hit Deploy → Upload

### Expected behavior

I can choose a board and upload successfully

### Actual behavior

I get an error saying «… unknown command “Bar” for “arduino-cli” … »

### How to fix

Enquote the argument of `--config-file` and other path arguments. Currently, the shell is asked to run (line breaks added for clarity):

```
"C:\Users\Foo Bar Qux\AppData\Local\Programs\XOD IDE\resources\arduino-cli.exe"
--config-file=C:\Users\Foo
Bar
Qux\AppData\Local\Temp\arduino-cli2t6a9o.cli-config.yml
config
dump
```

brusherru:
It fixes #1723

### DIFF
--- a/packages/arduino-cli/src/index.js
+++ b/packages/arduino-cli/src/index.js
@@ -1,4 +1,3 @@
-import os from 'os';
 import * as R from 'ramda';
 import { resolve } from 'path';
 import { promisifyChildProcess } from 'promisify-child-process';
@@ -17,8 +16,6 @@ const spawn = (bin, args, options) =>
   });
 
 const noop = () => {};
-const IS_WIN = os.platform() === 'win32';
-const escapeSpacesNonWin = R.unless(() => IS_WIN, R.replace(/\s/g, '\\ '));
 
 /**
  * Initializes object to work with `arduino-cli`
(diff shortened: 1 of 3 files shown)


packages/arduino-cli/src/index.js
```diff
@@ -30,51 +37,27 @@ const ArduinoCli = (pathToBin, config = null) => {
...
-      R.reject(R.isEmpty),
-      R.split(' ')
-    )(args);
-
-    const promise = spawn(escapeSpacesNonWin(pathToBin), spawnArgs, {
+    const spawnArgs = R.concat([`--config-file=${escapedConfigPath}`], args);
+    const proc = spawn(escapeSpacesNonWin(pathToBin), spawnArgs, {
       stdio: ['inherit', 'pipe', 'pipe'],
-      shell: true,
+      shell: !IS_WIN, // TODO: Can we get rid of `shell: true` for other platforms?
```
nkrkv: This indeed is the question. Shell features are necessary only if they are used (obviously). Do we use either of the things below?

- shell pipes
- filename wildcards
- environment variable expansion
- expansion of `~` to a user’s home directory

At first glance, we don’t. Needs testing though. Also, make sure the long-running process output is output in XOD deployment pane as it goes, that is in real time, with the setting chosen.

knopki: `+1` here. Using shell leads to more bugs.

knopki:
What else needs testing:
- killing child process
- stale process that prevents closing IDE (I remember such bug)

brusherru:
@nkrkv removed `shell: true` for all platforms and tested.
The `xod-cli` and killing child processes on closing IDE tested too.
Check it out, please.

packages/arduino-cli/src/index.js
```diff
@@ -17,8 +16,6 @@ const spawn = (bin, args, options) =>
   });
 
 const noop = () => {};
-const IS_WIN = os.platform() === 'win32';
-const escapeSpacesNonWin = R.unless(() => IS_WIN, R.replace(/\s/g, '\\ '));
```
nkrkv: :heart: for OS neutrality

PR #1908, opened 2019-11-19
tweak: switch to another compilation limits backend

knopki:
Ask `https://api.xod.io/billing` instead of `https://compile.xod.io/limits` about remaining compilations.


knopki:
@nkrkv @brusherru @evgenykochetkov friendly pinger

PR #1920, opened 2019-12-09
Fix mystically failed test

brusherru:
There is no issue and it's very strange... but I have to do it to merge https://github.com/xodio/xod/pull/1919 🙈


brusherru:
Very strange, but when I pushed it, it fixed without this change! 😱

PR #1926, opened 2019-12-13
Support hex color literals

### LINKED ISSUE #1922: Support hex color literals
XOD is visual. So why doesn’t it visualizes a color?

Currently, values bound to pins of type `xod/color/color` are disabled for edit with the “Custom type” gag. Let’s make them first-class citizens allowing literals like `#ffd406` to be used.

### Acceptance criteria

- [ ] I can bind values of the color type in Inspector
- [ ] I can enter `FFD406` as the value and it becomes `#FFD406`
- [ ] I can enter `#ffd406` as the value and it becomes `#FFD406`
- [ ] I can enter `#Fa7` as the value and it becomes `#FFAA77`
- [ ] I can enter an invalid value and it becomes `#000000`

brusherru:
It closes #1922


packages/xod-client/src/utils/normalizeColor.js
```diff
@@ -0,0 +1,11 @@
+import * as R from 'ramda';
+import { DEFAULT_VALUE_OF_TYPE, CUSTOM_TYPE } from 'xod-project';
+
+export default R.ifElse(
```
evgenykochetkov: Shouldn't it also automatically insert the leading `#` if it's missing? (`FFD406 ` -> `#FFD406 ` or even `FFF` -> `#FFFFFF`)?

packages/xod-project/src/constants.js
```diff
@@ -24,6 +24,13 @@ export const PIN_TYPE = {
   T3: 't3',
 };
 
+/**
+ * Enumeration of custom types with literals
+ */
+export const CUSTOM_TYPE = {
```
evgenykochetkov: I'm not sure if this should be called just `CUSTOM_TYPE`. For example, `isBindableCustomType`'s implementation looks a bit weird because of it.

brusherru: +1
I'll replace it with a proper (but probably longer 😞) name.

packages/xod-project/src/custom-types/index.js
```diff
@@ -0,0 +1,45 @@
...
+// =============================================================================
+//
+// Add custom type specifications in BINDABLE_CUSTOM_TYPES_SPECS
+// Key — an alias for easy access by a Developer
+// Value — specification object
+//
+// =============================================================================
+
+const BINDABLE_CUSTOM_TYPES_SPECS = {
+  COLOR: Color,
```
nkrkv: I see no reason to keep `Color` (and subsequent types) in own trivial modules. Suggest simplification: embed the specs right here.

brusherru: 👌

PR #1927, opened 2019-12-17
Add “big patch” fixture

evgenykochetkov:
For more realistic analysis of generated file sizes when developing new features


evgenykochetkov:
Should I also add it to `packages/xod-arduino/tools/test-avr-size.sh`?

nkrkv:
> Should I also add it to packages/xod-arduino/tools/test-avr-size.sh?

That would be great! So that we can continuously put an eye on code size in non-trivial cases.

evgenykochetkov:
Done!

PR #1930, opened 2019-12-24
Introduce read only types

evgenykochetkov:
Read-only types can not be redefined during runtime.

`emitValue` for read-only types is disabled.

This means that their values can be used at compile time (for validation, buffer allocation etc).
If a user defines `evaluateTmpl` function in the patch implementation
```
#pragma XOD evaluate_tmpl enable

// ...

template<uint8_t port>
void evaluateTmpl(Context ctx) {
    static_assert(isValidDigitalPort(port), "must be a valid digital port");

    evaluate(ctx);
}
```
read-only values will be passed to it as template arguments.
```
// inside runTransaction:
//...
xod__gpio__digital_write::evaluateTmpl<node_2_output_VAL>(&ctxObj);
```

Also, this PR introduces `state_constructor_params` pragma.
If used, the values of read-only inputs will be passed to State constructor as arguments.
```
#pragma XOD state_constructor_params enable

struct State {
    uint8_t someStateField;
    State(uint8_t myReadOnlyInput) {
        someStateField = myReadOnlyInput;
    }
};
```


packages/xod-arduino/src/Directives.re
```diff
@@ -127,6 +127,13 @@ let doesRaiseErrors = (code) =>
   |. Code.lastPragmaEndis("error_raise")
   |. Endis.toBoolean(Code.doesReferSymbol("raiseError", code));
 
+let implementsEvaluateTmpl = Code.doesReferSymbol("evaluateTmpl");
```
nkrkv: ...or `#pragma`. `#pragma` should always be checked. Searching for a symbol is an unstable way. Say, I can *refer* to a symbol `evaluateTmpl` and it would be wrongly considered as the definition.

workspace/__packages__/package_index.json
```diff
@@ -1181,6 +1181,64 @@
...
+            {"name": "Arduino Yún"},
+            {"name": "Arduino Uno"},
+            {"name": "Arduino Uno WiFi"},
+            {"name": "Arduino Diecimila"},
+            {"name": "Arduino Nano"},
+            {"name": "Arduino Mega"},
+            {"name": "Arduino MegaADK"},
+            {"name": "Arduino Leonardo"},
+            {"name": "Arduino Leonardo Ethernet"},
+            {"name": "Arduino Micro"},
```
nkrkv: Are these changes relate to the PR?

evgenykochetkov: Nope, this snuck in the changeset by accident. Fixed!

packages/xod-arduino/src/Directives.re
evgenykochetkov: Done, the pragma is named `evaluate_tmpl`

packages/xod-arduino/src/Directives.rei
```diff
@@ -50,6 +50,17 @@ let doesRaiseErrors: code => bool;
...
+/**
+  Returns if `evaluateTmpl` is found in the C++ code
+ */
+let implementsEvaluateTmpl: code => bool;
+
+/**
+  Returns if a node declares that it needs values
+  of read-only inputs in the state constructor
+ */
+let wantsStateConstructorWithParams: code => bool;
```
nkrkv: Can the doc comments describe the corresponding pragma in details. See comments for the functions above for the reference

evgenykochetkov: Done!

PR #1933, opened 2019-12-25
XOD Graphics library basics

gabbapeople:
- Introduce `xod/graphics` and ` __ardulib__ /Graphics` library basics.
- Add the `Line` graphic element.
- Add  `XColor()` struct constructors into the `packages/xod-arduino/platform/types.h`.
- Clang-format the `packages/xod-arduino/platform/types.h`.
- Update Fixtures.

### DIFF
--- a/workspace/__ardulib__/Graphics/Canvas.h
+++ b/workspace/__ardulib__/Graphics/Canvas.h
@@ -1,0 +1,34 @@
+
+#ifndef CANVAS_H
+#define CANVAS_H
+
+#include "XGraphics.h"
+
+/*
+ * Represents a rectangular piece of the screen (i.e. scene) which gets redrawn
+ * completely when at least a single child `XGraphics` element on it is updated.
+ * Holds the backgound color for the scene and the default foreground color for
+ * the child graphics primitives. 
+ */
+class Canvas : public XGraphics {
+private:
+    XColor _bgColor;
+    XColor _fgColor = { 0xFF, 0xFF, 0xFF };
+
+    XVector2<int16_t> _pivot;
+    int16_t _w;
+    int16_t _h;
+
+public:
+    void setPosition(int16_t x, int16_t y, int16_t w, int16_t h);
+    void setStyle(XColor bgColor, XColor fgColor);
+
+    BBox getUpstreamCanvasBBox() const;
+    XColor getFGColor() const;
+
+    void renderScanline(XRenderer* renderer, int16_t scanline, uint16_t* buffer, size_t bufferSize);
+};
+
+#include "Canvas.inl"
+
+#endif // CANVAS_H
--- a/workspace/__ardulib__/Graphics/Canvas.inl
+++ b/workspace/__ardulib__/Graphics/Canvas.inl
@@ -1,0 +1,27 @@
+
+void Canvas::setPosition(int16_t x, int16_t y, int16_t w, int16_t h) {
+    _pivot = XVector2<int16_t>(x, y);
+    _w = w;
+    _h = h;
+}
+
+void Canvas::setStyle(XColor bgColor, XColor fgColor) {
+    _bgColor = bgColor;
+    _fgColor = fgColor;
+}
+
+BBox Canvas::getUpstreamCanvasBBox() const {
+    return { _pivot, _w, _h };
+}
+
+XColor Canvas::getFGColor() const {
+    return _fgColor;
+}
+
+void Canvas::renderScanline(XRenderer* renderer, int16_t scanline, uint16_t* buffer, size_t bufferSize) {
+
+    uint16_t color = xcolorTo565(_bgColor);
+
+    for (int16_t x = 0; x < bufferSize; x++)
+        buffer[x] = color;
+}
--- a/workspace/__ardulib__/Graphics/Line.h
+++ b/workspace/__ardulib__/Graphics/Line.h
@@ -1,0 +1,30 @@
+
+#ifndef LINE_H
+#define LINE_H
+
+#include "XGraphics.h"
+
+/*
+ * Represents a line given by two points A(x0, y0) and B(x1, y1).
+ * The `setStyle` method sets up stroke color of the line otherwise
+ * the default foreground color of the scene is used.
+ */
+class Line : public XGraphics {
+private:
+    XVector2<int16_t> _a;
+    XVector2<int16_t> _b;
+
+    XColor* _strokeColor = nullptr;
+
+public:
+    Line(XGraphics* parent);
+
+    void setPosition(int16_t x0, int16_t y0, int16_t x1, int16_t y1);
+    void setStyle(XColor* strokeColor);
+
+    void renderScanline(XRenderer* renderer, int16_t scanline, uint16_t* buffer, size_t bufferSize);
+};
+
+#include "Line.inl"
+
+#endif // LINE_H
--- a/workspace/__ardulib__/Graphics/Line.inl
+++ b/workspace/__ardulib__/Graphics/Line.inl
@@ -1,0 +1,100 @@
+
+void swapInt16(int16_t& a, int16_t& b) {
+    int16_t t = a;
+    a = b;
+    b = t;
+}
+
+/*
+ * The Bresenham's line algorithm [Wikipedia](https://en.wikipedia.org/wiki/Bresenham%27s_line_algorithm).
+ * This implementation is augmented by the scanline intersection check during rasterisation.
+ * The standard algorithm assumes that the `b.x` > `a.x` and `b.y` > `a.y`. The `steep` value is used to eliminate
+ * this condition and to calculate a line at points with any coordinates.
+ */
+void putLineOnScanline(XVector2<int16_t> a, XVector2<int16_t> b, int16_t scanline, uint16_t* buffer, size_t bufferSize, uint16_t color) {
+
+    int16_t steep = abs(b.y - a.y) > abs(b.x - a.x);
+    if (steep) {
+        swapInt16(a.x, a.y);
+        swapInt16(b.x, b.y);
+    }
+
+    if (a.x > b.x) {
+        swapInt16(a.x, b.x);
+        swapInt16(a.y, b.y);
+    }
+
+    int16_t dx, dy;
+    dx = b.x - a.x;
+    dy = abs(b.y - a.y);
+
+    int16_t err = dx / 2;
+    int16_t ystep;
+
+    if (a.y < b.y) {
+        ystep = 1;
+    } else {
+        ystep = -1;
+    }
+
+    for (; a.x <= b.x; a.x++) {
+        if (steep) {
+            if (scanline == a.x)
+                if (a.y >= 0 && a.y < bufferSize)
+                    buffer[a.y] = color;
+        } else {
+            if (scanline == a.y)
+                if (a.x >= 0 && a.x < bufferSize)
+                    buffer[a.x] = color;
+        }
+        err -= dy;
+        if (err < 0) {
+            a.y += ystep;
+            err += dx;
+        }
+    }
+}
+
+Line::Line(XGraphics* parent)
+    : XGraphics(parent) {}
+
+void Line::setPosition(int16_t x0, int16_t y0, int16_t x1, int16_t y1) {
+    _a = XVector2<int16_t>(x0, y0);
+    _b = XVector2<int16_t>(x1, y1);
+}
+
+void Line::setStyle(XColor* strokeColor) {
+    _strokeColor = strokeColor;
+}
+
+void Line::renderScanline(XRenderer* renderer, int16_t scanline, uint16_t* buffer, size_t bufferSize) {
+
+    int16_t lineMinY = min(_a.y, _b.y);
+    int16_t lineMaxY = max(_a.y, _b.y);
+
+    int16_t lineMinX = min(_a.x, _b.x);
+    int16_t lineMaxX = max(_a.x, _b.x);
+
+    if (scanline > lineMaxY || scanline < lineMinY)
+        return;
+
+    uint16_t color = xcolorTo565(_strokeColor ? *_strokeColor : getFGColor());
+
+    if (lineMinY == lineMaxY) {
+        // Straight horizontal line. Use shortcut instead of slow algorithm.
+
+        for (int16_t x = lineMinX; x <= lineMaxX; x++)
+            if (x >= 0 && x < bufferSize)
+                buffer[x] = color;
+
+    } else if (lineMinX == lineMaxX) {
+        // Straight vertical line. Use shortcut instead of slow algorithm.
+
+        if (_a.x >= 0 && _a.x < bufferSize)
+            buffer[_a.x] = color;
+
+    } else {
+        // Regular diagonal line.
+        putLineOnScanline(_a, _b, scanline, buffer, bufferSize, color);
+    }
+}
--- a/workspace/__ardulib__/Graphics/XGraphics.h
+++ b/workspace/__ardulib__/Graphics/XGraphics.h
@@ -1,0 +1,51 @@
+
+#ifndef X_GRAPHICS_H
+#define X_GRAPHICS_H
+
+#include <Arduino.h>
+
+#include "XRender.h"
+#include "XVector2.h"
+
+using xod::XColor;
+
+/*
+ * A structure to hold a boundary box of a geometric object.
+ */
+struct BBox {
+    XVector2<int16_t> pivot;
+    int16_t width;
+    int16_t height;
+};
+
+/*
+ * A base class for graphics primitives. A primitive is a single node in the scene tree which eventually
+ * gets rendered. The `XGraphics` and derived classes are device-independent. They hold geometry
+ * and style data, and have methods to rasterize them [scan]line-by-line into a plain pixel buffer.
+ */
+class XGraphics {
+protected:
+    XGraphics* _parent;
+
+public:
+    XGraphics();
+    XGraphics(XGraphics* parent);
+
+    virtual XColor getFGColor() const;
+    virtual BBox getUpstreamCanvasBBox() const;
+
+    /*
+     * Rasterizes a specific graphics primitive for the specified `scanline`.
+     */
+    virtual void renderScanline(XRenderer* renderer, int16_t scanline, uint16_t* buffer, size_t bufferSize) {}
+    /*
+     * Rasterizes all graphics primitives in the scene tree for the specified `scanline`.
+     */
+    void renderScanlineRecursively(XRenderer* renderer, int16_t scanline, uint16_t* buffer, size_t bufferSize);
+    /*
+     * Rasterizes all graphics primitives for the specified scene.
+     */
+    void render(XRenderer* renderer);
+};
+
+#include "XGraphics.inl"
--- a/workspace/__ardulib__/Graphics/XGraphics.inl
+++ b/workspace/__ardulib__/Graphics/XGraphics.inl
@@ -1,0 +1,49 @@
+
+XGraphics::XGraphics() {
+    _parent = nullptr;
+}
+
+XGraphics::XGraphics(XGraphics* parent) {
+    _parent = parent;
+}
+
+XColor XGraphics::getFGColor() const {
+    return _parent ? _parent->getFGColor() : XColor();
+}
+
+BBox XGraphics::getUpstreamCanvasBBox() const {
+    return _parent ? _parent->getUpstreamCanvasBBox() : BBox({ { 0, 0 }, 0, 0 });
+}
+
+void XGraphics::render(XRenderer* renderer) {
+
+    BBox canvasBBox = getUpstreamCanvasBBox();
+
+    int16_t screenWidth = renderer->getScreenWidth();
+    int16_t screenHeight = renderer->getScreenHeight();
+
+    int16_t pivotX = max(0, canvasBBox.pivot.x);
+    int16_t pivotY = max(0, canvasBBox.pivot.y);
+
+    size_t bufferSize = (pivotX + canvasBBox.width >= screenWidth) ? screenWidth - pivotX : canvasBBox.width;
+
+    int16_t scanlineCount = (pivotY + canvasBBox.height >= screenHeight) ? screenHeight - pivotY : canvasBBox.height;
+
+    for (int16_t scanline = 0; scanline <= scanlineCount; scanline++) {
+        uint16_t buffer[bufferSize];
+        renderScanlineRecursively(renderer, scanline, buffer, bufferSize);
+        renderer->renderScanlinePart(pivotY + scanline, pivotX, pivotX + bufferSize - 1, buffer);
+    }
+}
+
+void XGraphics::renderScanlineRecursively(XRenderer* renderer, int16_t scanline, uint16_t* buffer, size_t bufferSize) {
+
+    if (_parent != nullptr)
+        _parent->renderScanlineRecursively(renderer, scanline, buffer, bufferSize);
+
+    renderScanline(renderer, scanline, buffer, bufferSize);
+}
+
+inline uint16_t xcolorTo565(XColor color) {
+    return ((color.r & 0xF8) << 8) | ((color.g & 0xFC) << 3) | (color.b >> 3);
+}
--- a/workspace/__ardulib__/Graphics/XRender.h
+++ b/workspace/__ardulib__/Graphics/XRender.h
@@ -1,0 +1,16 @@
+
+#ifndef X_RENDERER_H
+#define X_RENDERER_H
+
+/*
+ * Pure virtual renderer, should be overridden by the renderer of a specific device.
+ */
+class XRenderer {
+public:
+    virtual void renderScanlinePart(int16_t scanline, int16_t xmin, int16_t xmax, uint16_t* lineBuffer) = 0;
+
+    virtual int16_t getScreenWidth() const;
+    virtual int16_t getScreenHeight() const;
+};
+
+#endif // X_RENDERER_H
--- a/workspace/__lib__/xod/graphics/line-colored/patch.cpp
+++ b/workspace/__lib__/xod/graphics/line-colored/patch.cpp
@@ -1,0 +1,47 @@
+
+// clang-format off
+{{#global}}
+#include <Line.h>
+{{/global}}
+// clang-format on
+
+struct State {
+    uint8_t mem[sizeof(Line)];
+    Line* lineColored;
+    int16_t x0, y0, x1, y1;
+    XColor strokeColor;
+};
+
+// clang-format off
+{{ GENERATED_CODE }}
+// clang-format on
+
+void evaluate(Context ctx) {
+    auto state = getState(ctx);
+
+    auto gfx = getValue<input_GFX>(ctx);
+    state->strokeColor = getValue<input_C>(ctx);
+
+    int16_t x0 = (int16_t)getValue<input_X0>(ctx);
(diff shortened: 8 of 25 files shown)


workspace/__ardulib__/Graphics/XGraphics.h
```diff
@@ -0,0 +1,458 @@
...
+    XBitmap() {
+        _parent = nullptr;
+    }
+
+    virtual const XFill* getFill() const;
+    virtual const XColor* getBgColor() const;
+    virtual const XColor* getStrokeColor() const;
+
+    bool renderLine(XRender* renderer, int16_t line, uint16_t* buffer);
+    virtual bool renderBitmapLine(XRender* renderer, int16_t line, uint16_t* buffer) { return false; }
```
brusherru: A bad naming. Only reading of the implementation makes clear the difference.
What do you think about renaming them to `renderLineRecursively` and `renderLine`? 🤔

workspace/__ardulib__/Graphics/XGraphics.h
```diff
@@ -0,0 +1,458 @@
...
+    _old_B = _B;
+    _old_A = _A;
+    _A = XVector2<int16_t>(x0, y0);
+    _B = XVector2<int16_t>(x1, y1);
+}
+
+bool Line::renderBitmapLine(XRender* renderer, int16_t line, uint16_t* buffer) {
+    bool res = false;
+
+    if (getStrokeColor() != nullptr) {
```
brusherru: Invert condition and make an early return.

workspace/__ardulib__/Graphics/XGraphics.h
```diff
@@ -0,0 +1,458 @@
...
+
+        if (_old_A.y == _old_B.y) {
+            if (line == _old_A.y) {
+                res = true;
+            }
+        }
+
+        if (_old_A.y != _old_B.y) {
```
brusherru: else?

workspace/__ardulib__/Graphics/XGraphics.h
```diff
@@ -0,0 +1,458 @@
...
+                }
+
+                if (_A.x == _B.x) {
+                    buffer[_A.x] = color;
+                    res = true;
+                }
+            }
+        }
+
+        if (_A.y != _B.y) {
```
brusherru: else?

workspace/__ardulib__/Graphics/XGraphics.h
```diff
@@ -0,0 +1,458 @@
...
+        if (_old_A.y == _old_B.y) {
+            if (line == _old_A.y) {
+                res = true;
+            }
+        }
+
+        if (_old_A.y != _old_B.y) {
+            int16_t old_minY;
+            int16_t old_maxY;
+            if (_old_A.y > _old_B.y) {
```
brusherru: Looks like such blocks can be easily replaced with the usage of `max` and `min` functions.
Two lines instead of seven 🤓

workspace/__ardulib__/Graphics/XGraphics.h
```diff
@@ -0,0 +1,458 @@
...
+    XFill() {}
+    virtual XColor colorAtPoint(XVector2<int16_t> point) {}
+    virtual void setSolidColor(XColor _color) {}
+};
+
+class XSolidFill : public XFill {
+private:
+    XColor _color;
+public:
+    XSolidFill(XColor color) {
```
nkrkv: This file is way too big. Would you split it into h+cpp pair per concept? I see the following:

- Fill
- Stroke (for now, I’m pretty sure this is an extraneous concept)
- Bitmap
- Canvas
- Point
- Line
- Rectangle

brusherru: But I'm not sure about Stroke, because it's pretty much simpler to set stroke instead of creating another one figure (especially, for curves or even for diagonal lines). :-P

workspace/__ardulib__/Graphics/XGraphics.h
```diff
@@ -0,0 +1,458 @@
...
+class XBitmap {
+protected:
+    XBitmap* _parent;
+public:
+    XBitmap() {
+        _parent = nullptr;
+    }
+
+    virtual const XFill* getFill() const;
+    virtual const XColor* getBgColor() const;
```
brusherru: Does it make sense to use "fill" and "background-color" at the same time?
Maybe better to get rid of the background color and only fill and stroke will remain?

workspace/__ardulib__/Graphics/XGraphics.h
```diff
@@ -0,0 +1,458 @@
...
+    virtual const XColor* getStrokeColor() const;
+
+    bool renderLine(XRender* renderer, int16_t line, uint16_t* buffer);
+    virtual bool renderBitmapLine(XRender* renderer, int16_t line, uint16_t* buffer) { return false; }
+    void renderScreen(XRender* renderer);
+
+    int16_t drawLinePointAtLine(int16_t line, XVector2<int16_t> A, XVector2<int16_t> B);
+    void swapInt16(int16_t& a, int16_t& b);
```
evgenykochetkov: Why is `swapInt16` a member of `XBitmap`?

workspace/__ardulib__/Graphics/XGraphics.h
evgenykochetkov: Also, I would not expect any of those methods from a class named XBitmap. "XBitmap" sounds like a thing with known width and height that you can iterate over and get individual pixels, not something with fill, stroke etc

workspace/__ardulib__/Graphics/XGraphics.h
```diff
@@ -0,0 +1,32 @@
...
+using xod::XColor;
+
+class XGraphics {
+protected:
+    XGraphics* _parent;
+public:
+    XGraphics();
+
+    virtual const XFill* getFill() const;
+    virtual const XColor* getBgColor() const;
```
nkrkv: Can we get rid of `getBgColor`?

workspace/__ardulib__/Graphics/XGraphics.h
```diff
@@ -0,0 +1,32 @@
...
+    virtual const XColor* getStrokeColor() const;
+
+    bool renderScanlineRecursively(XRenderer* renderer, int16_t line, uint16_t* buffer);
+    virtual bool renderScanline(XRenderer* renderer, int16_t line, uint16_t* buffer) { return false; }
+    void renderScreen(XRenderer* renderer);
```
nkrkv: Comments are necessary: for what, what are return values

workspace/__ardulib__/Graphics/XGraphics.h
```diff
@@ -0,0 +1,32 @@
...
+
+    int16_t drawPointAtLine(int16_t scanline, XVector2<int16_t> A, XVector2<int16_t> B);
```
nkrkv: Move out of the class. Think about a better name.

workspace/__ardulib__/Graphics/Canvas.h
```diff
@@ -0,0 +1,22 @@
+
+#ifndef CANVAS_H
+#define CANVAS_H
+
+#include "XGraphics.h"
+
+class Canvas : public XGraphics {
```
nkrkv: Rename to `Background`?

workspace/__ardulib__/Graphics/Canvas.h
```diff
@@ -0,0 +1,22 @@
...
+private:
+    XColor _color = { 0, 0, 0 };
```
nkrkv: Move to `XColor` ctor? Follow RAII

workspace/__ardulib__/Graphics/Canvas.inl
```diff
@@ -0,0 +1,37 @@
...
+    return &_color;
+}
+
+bool Canvas::renderScanline(XRenderer* renderer, int16_t scanline, uint16_t* buffer) {
+    bool res = false;
+
+    if (getBgColor() == nullptr)
+        return res;
+
+    uint16_t color = color888to565(getBgColor()->r, getBgColor()->g, getBgColor()->b);
```
nkrkv: Create a function `xcolorTo565`? So that:

```cpp
uint16_t color = color888to565(*myColor);
```

workspace/__ardulib__/Graphics/FillSolid.h
```diff
@@ -0,0 +1,29 @@
...
+
+#include "XGraphics.h"
+
+class FillSolid : public XGraphics {
+private:
+    XFill* _fill = nullptr;
+
+public:
+    FillSolid(XGraphics* parent, XColor color) {
+        _fill = new XFillSolid(color);
```
nkrkv: Why allocate? You could make `_fill` a regular member `XFillSolid _fill`.

workspace/__ardulib__/Graphics/Line.h
```diff
@@ -0,0 +1,21 @@
+
+#ifndef LINE_H
+#define LINE_H
+
+#include "XGraphics.h"
+
+class Line : public XGraphics {
+private:
+    XVector2<int16_t> _A;
+    XVector2<int16_t> _B;
```
nkrkv: Use the regular naming scheme: `_a`, `_b`

workspace/__ardulib__/Graphics/Line.inl
```diff
@@ -0,0 +1,41 @@
...
+
+void Line::renderScanline(XRenderer* renderer, int16_t scanline, uint16_t* buffer) {
+
+    if (getStroke() == nullptr)
+        return;
+
+    uint16_t color = xcolorTo565(*getStroke());
+
+    int16_t _minY = min(_A.y, _B.y);
+    int16_t _maxY = max(_A.y, _B.y);
```
nkrkv: Why underscore before the name?

workspace/__ardulib__/Graphics/Line.inl
```diff
@@ -0,0 +1,41 @@
...
+
+    if (scanline >= _minY && scanline <= _maxY) {
```
nkrkv: Inverse the condition and use early return instead.

workspace/__ardulib__/Graphics/Rectangle.inl
```diff
@@ -0,0 +1,46 @@
...
+
+void Rectangle::setParams(int16_t x, int16_t y, int16_t w, int16_t h) {
+    _A = XVector2<int16_t>(x, y);
+    _w = w;
+    _h = h;
+}
+
+void Rectangle::renderScanline(XRenderer* renderer, int16_t scanline, uint16_t* buffer) {
+
+    if (scanline >= _A.y && scanline <= _A.y + _h - 1) {
```
nkrkv: Use early return, please

workspace/__lib__/xod/graphics/background/patch.cpp
```diff
@@ -0,0 +1,35 @@
+
+// clang-format off
+{{#global}}
+// clang-format on
+
+#include <Background.h>
+
+// clang-format off
+{{/global}}
+// clang-format on
```
nkrkv: It is enough to replace these strings with:

```cpp
// clang-format off
{{#global}}
#include <Background.h>
{{/global}}
// clang-format on
```

nkrkv:
Hey, great PR. Although a lot of work has been done, some rough edges are still here. They are essential because this PR introduces a fundamental concept, so things have to be polished before going public to allow forward/backward compatibility.

### Fill vs. Stroke: the problem

The concepts of Fill & Stroke are often separated because in powerful graphics packages, Fill settings are inequal to Stroke settings. Besides color information, Stroke also has a thickness, alignment (inset/center/outset), and rounding/ending styles.

Do we need to support this? This is pretty tricky to be correctly implemented. It sounds like a massive effort. So, no, we cannot do it now.

If not, let’s do Fill, and that’s it. A few new primitives to express rectangle/circle outline without body fill will do the job. And we don’t have to handle the ugly cases of filling a polyline, for example.

Furthermore, the current implementation suggests a Fill abstraction with only a solid-fill specialization. And the Stroke is given as a solid color without any abstractions for further customization. I’m sure without specific applications at the current stage, the interface will not sustain the evolution.

 The ideal option is to bring the system to a point where the bare minimum is implemented and polished with a clear way to support things like line thicknesses, gradients, patterns, transformations, etc. in the future.

But how to define the public graphics library API in this case? After all, every `XGraphics` from scratch deals with filling the buffer and does not use any intermediate abstractions for drawing strokes and filled areas.

### Fill vs. Stroke: a possible solution

It is possible that each node, in addition to the required geometric parameters, should have its own style passed to own pin(s): rectangles have RectangleStyle, circles have CircleStyle, lines have LineStyle, and so on. In the current release, these styles can be reduced only to a color or a pair of colors. And then they can evolve with full backward compatibility. Now we can give no promises about universal gradient strokes, pattern fills, and other complicated things.

A definition of style can turn out to be rather cumbersome if placed for each graphics primitive. The style node might have stroke thickness, dash style, border radius, mitter limit, draw direction, arrow settings, and so on, and so on. In most common cases, when a xoder wants to “just draw this thing!” there should be a shortcut.

Putting all together, the current state is:

- We support only 1 pixel solid stroke of a solid color
- We support only solid fill of some primitives (rect, circle, not point and line)
- A user more often needs a consistent color scheme than a rainbow in each primitive
- A demand for rectangles and circles with stroke and fill of different colors is questionable

So, this is how it might be addressed.

Rename `Background` back to `Canvas` (sorry for that) and make it hold two colors: `BG` — the background color and `FG` — the color used as a consistency default if not said otherwise.
Throw away the abstractions related to the Fill and Stroke. Yes, bye-bye, `XSolidFill`, state context, and friends.
Make separate `XGraphics` primitives for filled and outlined shapes (see below)
For each primitive, create an “advanced” node to allow color overriding. Someday the color override might become a complex style override.

So, nodes:

- rect-outline — a rectangle outline drawn with `FG` color
- rect-solid — a rectangle filled with `FG` color
- circle-outline
- circle-solid
- line — a straight line drawn with `FG` color
- point — a point drawn with `FG` color

To change the color, let’s provide additional versions:

- rect-outline-colored — like `rect-outline` but has an extra input which overrides the color
- rect-solid-colored
- circle-outline-colored
- circle-solid-colored
- line-colored
- point-colored

### Partial scanlines

The minimal portion of the update in the current implementation is the whole screen. That’s fine if our display is used to show a single “scene” but will quickly become a problem if it shows multiple widgets that update independently.

Imagine a little ellipsis symbol a the corner of the display, which is shown only when an HTTP request is in progress. Should we redraw the whole weather widget, the device status, the log chart only to show or hide the three little dots?

We clearly have multiple _canvases_ on a single display. So let’s give a way to express it! The topmost `canvas` node could have dimensions: x, y, w, h. Then, the basic `XGraphics` method could evolve from the current

```cpp
void renderScanline(XRenderer* renderer, int16_t scanline, uint16_t* buffer, uint16_t bufferSize);
```

to

```cpp
void renderScanline(XRenderer* renderer, int16_t scanline, int16_t xmin, int16_t xmax, uint16_t* buffer);
```

The `buffer` is guaranted to be at least `xmax - xmin + 1` in size. An `XGraphics` primitive promises to fill the buffer correctly understanding that only a part of scanline [xmin; xmax] is requested.

This way, a xoder can explicitly define isolated areas on the display by defining multiple `GFX` branches and have the updated independently, only when it is necessary.

workspace/__ardulib__/Graphics/Canvas.h
```diff
@@ -0,0 +1,32 @@
...
+#ifndef CANVAS_H
+#define CANVAS_H
+
+#include "XGraphics.h"
+
+class Canvas : public XGraphics {
+private:
+    XColor _bgColor;
+    XColor _fgColor;
+    XVector2<int16_t> _a;
```
nkrkv: `_a` is unclear a bit. Suggest `_pivot`.

workspace/__ardulib__/Graphics/Canvas.inl
```diff
@@ -0,0 +1,36 @@
...
+int16_t Canvas::getCanvasHeight() const {
+    return _h;
+}
+
+void Canvas::renderScanline(XRenderer* renderer, int16_t scanline, int16_t xmin, int16_t xmax, uint16_t* buffer) {
+
+    uint16_t color = xcolorTo565(_bgColor);
+
+    for (int16_t x = 0; x < _w; x++)
+        buffer[x] = color;
```
nkrkv: This should take `xmin` and `xmax` into account for partial updates (for whatever reason).

workspace/__ardulib__/Graphics/Line.h
```diff
@@ -0,0 +1,75 @@
...
+        ystep = 1;
+    } else {
+        ystep = -1;
+    }
+
+    for (; a.x <= b.x; a.x++) {
+        if (steep) {
+            if (scanline == a.x)
+                if (a.y >= xmin && a.y <= xmax)
+                    buffer[a.y - xmin] = color;
```
nkrkv: xs mixed with ys… A bug?

workspace/__ardulib__/Graphics/XGraphics.inl
```diff
@@ -0,0 +1,47 @@
+
+XGraphics::XGraphics() {
+    _parent = nullptr;
+}
+
+const XColor* XGraphics::getFGColor() const {
+    return _parent ? _parent->getFGColor() : nullptr;
+}
```
nkrkv: In many places, it is expected that `XColor*` might never be `nullptr`. Intuitively, it is a sane assumption. To avoid shooting the leg, I’d change the method signature to plain `XColor` instead of `XColor*`. The default may be mono-white or mono-black then (inverse of the background default).

workspace/__ardulib__/Graphics/XGraphics.h
```diff
@@ -0,0 +1,31 @@
...
+
+    virtual XColor getFGColor() const;
+
+    virtual XVector2<int16_t> getCanvasPivot() const;
+    virtual int16_t getCanvasWidth() const;
+    virtual int16_t getCanvasHeight() const;
+
+    virtual void  renderScanline(XRenderer* renderer, int16_t scanline, uint16_t bufferSize, uint16_t* buffer) {}
+    void renderScanlineRecursively(XRenderer* renderer, int16_t scanline, uint16_t bufferSize, uint16_t* buffer);
+    void renderCanvas(XRenderer* renderer);
```
nkrkv: Why should this method belong to the base `XGraphics` and not to the `Canvas` itself

workspace/__ardulib__/Graphics/XRender.h
```diff
@@ -0,0 +1,26 @@
...
+    }
+
+    int16_t getRendererHeight() const {
+        return _height;
+    }
+
+protected:
+
+    int16_t _width = 0;
+    int16_t _height = 0;
```
nkrkv: Width and height are here, but there’s no way to set them. Perhaps, it is better to leave such details to specific implementations.

workspace/__ardulib__/Graphics/XGraphics.h
```diff
@@ -0,0 +1,31 @@
...
+protected:
+    XGraphics* _parent;
+public:
+    XGraphics();
+
+    virtual XColor getFGColor() const;
+
+    virtual XVector2<int16_t> getCanvasPivot() const;
+    virtual int16_t getCanvasWidth() const;
+    virtual int16_t getCanvasHeight() const;
```
nkrkv: Instead of keeping three virtual methods for the polymorphism which is used in a single place, I’d replace all three with a single:

```cpp
virtual const Canvas* getUpstreamCanvas() const;
```

Do not forget the forward declaration in this case:

```cpp
class Canvas;
```

workspace/__ardulib__/Graphics/XRender.h
```diff
@@ -0,0 +1,15 @@
+
+#ifndef X_RENDERER_H
+#define X_RENDERER_H
+
+class XRenderer {
+public:
+    XRenderer() {}
+
+    virtual void setScanlinePart(int16_t scanline, int16_t xmin, int16_t xmax, uint16_t* lineBuffer) = 0;
```
nkrkv:
```suggestion
    virtual void renderScanlinePart(int16_t scanline, int16_t xmin, int16_t xmax, uint16_t* lineBuffer) = 0;
```

?

workspace/__ardulib__/Graphics/Circle.inl
```diff
@@ -0,0 +1,56 @@
...
+        color = xcolorTo565(getFGColor());
+    else
+        color = xcolorTo565(*_strokeColor);
+
+    putCircleOutlineOnScanline(_a, _r, scanline, buffer, bufferSize, color);
+}
+
+CircleSolid::CircleSolid(XGraphics* parent) {
+    _parent = parent;
+}
```
nkrkv: Why override the base constructor if it does nothing new? I suspect you may completely drop it

workspace/__ardulib__/Graphics/Circle.h
```diff
@@ -0,0 +1,128 @@
...
+public:
+    CircleSolid(XGraphics* parent);
+
+    void setPosition(int16_t x, int16_t y, int16_t r);
+    void setStyle(XColor* fillColor);
+
+    void renderScanline(XRenderer* renderer, int16_t scanline, uint16_t* buffer, size_t bufferSize);
+};
+
+void putCircleOutlineOnScanline(XVector2<int16_t> a, int16_t r, int16_t scanline, uint16_t* buffer, size_t bufferSize, uint16_t color) {
```
nkrkv: Shouldn’t this and the following function be in `.inl` file?

workspace/__ardulib__/Graphics/XGraphics.h
```diff
@@ -0,0 +1,40 @@
...
+    virtual XColor getFGColor() const;
+    virtual BBox getUpstreamCanvasBBox() const;
+
+    virtual void renderScanline(XRenderer* renderer, int16_t scanline, uint16_t* buffer, size_t bufferSize) {}
+    void renderScanlineRecursively(XRenderer* renderer, int16_t scanline, uint16_t* buffer, size_t bufferSize);
+    void render(XRenderer* renderer);
+};
+
+inline uint16_t xcolorTo565(XColor color) {
+    return ((color.r & 0xF8) * 256) | ((color.g & 0xFC) * 8) | (color.b / 8);
```
nkrkv: In this case `<<`s make much  more sense than multiplications because you are tweaking bits'n'bytes

```suggestion
    return ((color.r & 0xF8) << 8) | ((color.g & 0xFC) << 3) | (color.b >> 1);
```

workspace/__ardulib__/Graphics/XGraphics.h
```diff
@@ -0,0 +1,36 @@
...
+
+using xod::XColor;
+
+struct BBox {
+    XVector2<int16_t> pivot;
+    int16_t width;
+    int16_t height;
+};
+
+class XGraphics {
```
nkrkv:
```suggestion
/*
 * A base class for graphics primitives. A primitive is a single node in the scene tree which eventually
 * gets rendered. The `XGraphics` and derived classes are device-independent. They hold geometry
 * and style data, and have methods to rasterize them [scan]line-by-line into a plain pixel buffer.
 */
class XGraphics {
```

workspace/__ardulib__/Graphics/XGraphics.h
```diff
@@ -0,0 +1,36 @@
...
+    XGraphics* _parent;
+
+public:
+    XGraphics();
+    XGraphics(XGraphics* parent);
+
+    virtual XColor getFGColor() const;
+    virtual BBox getUpstreamCanvasBBox() const;
+
+    virtual void renderScanline(XRenderer* renderer, int16_t scanline, uint16_t* buffer, size_t bufferSize) {}
```
nkrkv: A doc comment for the major interface method is desired

workspace/__ardulib__/Graphics/XGraphics.inl
```diff
@@ -0,0 +1,55 @@
...
+    canvasBBox.pivot.x = canvasBBox.pivot.x < 0 ? 0 : canvasBBox.pivot.x;
+    canvasBBox.pivot.y = canvasBBox.pivot.y < 0 ? 0 : canvasBBox.pivot.y;
+
+    if (canvasBBox.pivot.x + canvasBBox.width >= screenWidth)
+        bufferSize = screenWidth - canvasBBox.pivot.x;
+    else
+        bufferSize = canvasBBox.width;
+
+    if (canvasBBox.pivot.y + canvasBBox.height >= screenHeight)
+        canvasBBox.height = screenHeight - canvasBBox.pivot.y;
```
nkrkv: Avoid mutations and branches where possible:

```cpp
int16_t pivotX = max(0, canvasBBox.pivot.x);
int16_t pivotY = max(0, canvasBBox.pivot.y);

size_t bufferSize = (pivotX + canvasBBox.width >= screenWidth)
    ? screenWidth - pivotX
    : canvasBBox.width;

int16_t scanlineCount = (pivotY + canvasBBox.height >= screenHeight)
    ? screenHeight - pivotY
    : canvasBBox.height;
```

workspace/__lib__/xod/graphics/line-colored/patch.cpp
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    if (x0 != state->x0 || y0 != state->y0 || x1 != state->x1 || y1 != state->y1) {
+        state->x0 = x0;
+        state->y0 = y0;
+        state->x1 = x1;
+        state->y1 = y1;
+        state->lineColored->setPosition(x0, y0, x1, y1);
+        emitValue<output_GFXU0027>(ctx, state->lineColored);
+    }
```
nkrkv: The code is repetitive. You could do the following:

```cpp
if (isSettingUp()) {
    state->lineColored = new (state->mem) Line(gfx);
}

if (isSettingUp() || x0 != state->x0 || y0 != state->y0 || x1 != state->x1 || y1 != state->y1) {
    /* ... the common code ... */
}
```

workspace/__lib__/xod/graphics/line-colored/patch.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+    if (isInputDirty<input_CLR>(ctx)) {
+        state->lineColored->setStyle(&state->strokeColor);
+        emitValue<output_GFXU0027>(ctx, state->lineColored);
+    }
```
nkrkv: Not sure why this branch is separated from the case above

workspace/__lib__/xod/graphics/line-colored/patch.xodp
```diff
@@ -0,0 +1,85 @@
...
+        "units": "slots",
+        "x": 0,
+        "y": 0
+      },
+      "type": "@/input-graphics"
+    },
+    {
+      "description": "Stroke color.",
+      "id": "r1qVIP3b8",
+      "label": "CLR",
```
nkrkv: A bit unfortunate label. `CLR` is for CLEAR usually. I’d recommend calling it `C`.

workspace/__lib__/xod/graphics/line/patch.cpp
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    if (x0 != state->x0 || y0 != state->y0 || x1 != state->x1 || y1 != state->y1) {
+        state->x0 = x0;
+        state->y0 = y0;
+        state->x1 = x1;
+        state->y1 = y1;
+        state->line->setPosition(x0, y0, x1, y1);
+        emitValue<output_GFXU0027>(ctx, state->line);
+    }
```
nkrkv: Ditto

PR #1934, opened 2019-12-26
Support of ST7735 based TFT displays 

gabbapeople:
Add custom c++ library and `xod-dev/st7735-display` library.

### DIFF
--- a/workspace/__ardulib__/ST7735/XST7735.cpp
+++ b/workspace/__ardulib__/ST7735/XST7735.cpp
@@ -1,0 +1,195 @@
+
+#include "XST7735.h"
+
+inline uint16_t swapColor565(uint16_t color) {
+    return (color << 11) | (color & 0x07E0) | (color >> 11);
+}
+
+ST7735::ST7735(uint8_t cs, uint8_t dc, uint8_t rst) {
+    _cs = cs;
+    _rs = dc;
+    _rst = rst;
+    _width = ST7735_TFTWIDTH;
+    _height = ST7735_TFTHEIGHT;
+}
+
+void ST7735::writeCmd(uint8_t cmd) {
+
+#ifdef SPI_HAS_TRANSACTION
+    SPI.beginTransaction(spisettings);
+#endif
+
+    *rsport &= ~rspinmask;
+
+    *csport &= ~cspinmask;
+    SPI.transfer(cmd);
+    *csport |= cspinmask;
+
+#ifdef SPI_HAS_TRANSACTION
+    SPI.endTransaction();
+#endif
+}
+
+void ST7735::writeData(uint8_t data) {
+
+#ifdef SPI_HAS_TRANSACTION
+    SPI.beginTransaction(spisettings);
+#endif
+
+    *rsport |= rspinmask;
+
+    *csport &= ~cspinmask;
+    SPI.transfer(data);
+    *csport |= cspinmask;
+
+#ifdef SPI_HAS_TRANSACTION
+    SPI.endTransaction();
+#endif
+}
+
+void ST7735::setAddrWindow(uint8_t x0, uint8_t y0, uint8_t x1, uint8_t y1) {
+    writeCmd(ST7735_CASET);
+    writeData(0x00);
+    writeData(x0 + _colstart);
+    writeData(0x00);
+    writeData(x1 + _colstart);
+
+    writeCmd(ST7735_RASET);
+    writeData(0x00);
+    writeData(y0 + _rowstart);
+    writeData(0x00);
+    writeData(y1 + _rowstart);
+
+    writeCmd(ST7735_RAMWR);
+}
+
+void ST7735::commandList(const uint8_t* addr) {
+
+    uint8_t numCommands = pgm_read_byte(addr++); // Number of commands to follow.
+    uint8_t numArgs;
+    uint16_t ms;
+
+    while (numCommands--) { // For each command.
+        writeCmd(pgm_read_byte(addr++)); // Read, issue command.
+        numArgs = pgm_read_byte(addr++); // Number of args to follow.
+        ms = numArgs & DELAY; // If hibit set, delay follows args.
+        numArgs &= ~DELAY; // Mask out delay bit.
+        while (numArgs--) { // For each argument.
+            writeData(pgm_read_byte(addr++)); // Read, issue argument.
+        }
+
+        if (ms) {
+            ms = pgm_read_byte(addr++); // Read post-command delay time (ms).
+            if (ms == 255)
+                ms = 500; // If 255, delay for 500 ms.
+            delay(ms);
+        }
+    }
+}
+
+void ST7735::begin() {
+
+    _colstart = _rowstart = 0; // May be overridden in init func.
+
+    pinMode(_rs, OUTPUT);
+    pinMode(_cs, OUTPUT);
+    csport = portOutputRegister(digitalPinToPort(_cs));
+    cspinmask = digitalPinToBitMask(_cs);
+    rsport = portOutputRegister(digitalPinToPort(_rs));
+    rspinmask = digitalPinToBitMask(_rs);
+
+    SPI.begin();
+
+#ifdef SPI_HAS_TRANSACTION
+    spisettings = SPISettings(4000000L, MSBFIRST, SPI_MODE0);
+#else
+#if defined(ARDUINO_ARCH_SAM)
+    SPI.setClockDivider(24); // 4 MHz (half speed).
+#else
+    SPI.setClockDivider(SPI_CLOCK_DIV4); // 4 MHz (half speed).
+#endif // ARDUINO_ARCH_SAM
+    SPI.setBitOrder(MSBFIRST);
+    SPI.setDataMode(SPI_MODE0);
+#endif // SPI_HAS_TRANSACTION
+
+    // Toggle RST low to reset; CS low so it'll listen to us.
+    *csport &= ~cspinmask;
+    if (_rst == NOT_A_PIN)
+        return;
+
+    pinMode(_rst, OUTPUT);
+    digitalWrite(_rst, HIGH);
+    delay(100);
+    digitalWrite(_rst, LOW);
+    delay(100);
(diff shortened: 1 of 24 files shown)


workspace/__ardulib__/ST7735/XST7735.cpp
```diff
@@ -0,0 +1,200 @@
...
+    // toggle RST low to reset; CS low so it'll listen to us
+    *csport &= ~cspinmask;
+    if (_rst) {
+        pinMode(_rst, OUTPUT);
+        digitalWrite(_rst, HIGH);
+        delay(500);
+        digitalWrite(_rst, LOW);
+        delay(500);
+        digitalWrite(_rst, HIGH);
+        delay(500);
```
brusherru: Ouch. Such long delays which will delay the whole program 🤔 

Of course, it will be delayed once at the beginning of the program (for **each**❗️ connected display), but I think it's not a perfect solution... 🤔 

What if... this function can set some flag/enum, store current time and return `bool`. And then the XOD node which calls `dev->init()` may setTimeout and call it again if it returned `false`. So, this function will be called again, checked for the flag/enum, elapsed time and do the next things...?

But probably it does not matter at all 🤔

workspace/__ardulib__/ST7735/XST7735.cpp
```diff
@@ -0,0 +1,200 @@
...
+inline uint16_t swapColor565(uint16_t color) {
+    return (color << 11) | (color & 0x07E0) | (color >> 11);
+}
+
+ST7735::ST7735(uint8_t cs, uint8_t dc, uint8_t rst, uint8_t type) {
+    _cs = cs;
+    _rs = dc;
+    _rst = rst;
+    _type = type;
+    _width = ST7735_TFTWIDTH;
```
brusherru: Probably the implementation of the current library works fine only with the displays 128x160 (and in this case, I prefer to name it `ST7735_128x160`), I did not dive deep in the commands.
But it looks like `_width` and `_height` are hardcoded, but these hardcoded sizes do not used anywhere in the code. If it's the universal implementation for most of the ST7735 displays, I propose to remove `ST7735_TFTWIDTH` and `...HEIGHT` and get these sizes from the constructor arguments.

P.S.
Google says that there are ST7735 displays of other sizes, such as 128x128.

workspace/__ardulib__/ST7735/XST7735.cpp
```diff
@@ -0,0 +1,200 @@
...
+            commandList(Rcmd3);
+            break;
+        case 3: // RR Type
+            commonInit(Rcmd1);
+            commandList(Rcmd2red);
+            commandList(Rcmd3);
+            break;
+        default:
+            break;
+    }
```
nkrkv: This code refers to _all_ `XcmdX` globals based on a runtime value `_type`. The compiler has no chance to throw out any of them besides that only 1 of 3 is used in any particular moment. It’s a flash memory waste.

Instead of accepting and holding `_type` in the constructor, make a few initialization methods:

- `initG`
- `initB`
- `initRG`
- `initRR`

Call only the necessary `init` directly in corresponding nodes’ implementation.

workspace/__lib__/xod-dev/st7735-display/st7735-128x160/patch.cpp
```diff
@@ -0,0 +1,35 @@
...
+    auto state = getState(ctx);
+
+    uint8_t cs = getValue<input_CS>(ctx);
+    uint8_t dc = getValue<input_DC>(ctx);
+    uint8_t rst = getValue<input_RST>(ctx);
+
+    if (!isValidDigitalPort(cs) || !isValidDigitalPort(dc) || !isValidDigitalPort(rst)) {
+        raiseError(ctx);
+        return;
+    }
```
nkrkv: These checks can now be performed in compile-time. See `digital-write` for the reference.

workspace/__lib__/xod-dev/st7735-display/st7735/patch.cpp
```diff
@@ -0,0 +1,16 @@
...
+
+struct State {};
+
+using Type = ST7735*;
+
+// clang-format off
+{{ GENERATED_CODE }}
+// clang-format on
+
+void evaluate(Context ctx) {}
```
nkrkv: This guy could keep the memory buffer and object. Others can initialize.

nkrkv:
About node/pin naming:

- `render :: ACK` should be `render :: DONE`
- `st7735-wwxhh-x-type` should be `st7735-wwxhh-x-device`
- There should be quickstart nodes which simply accept `GFX` and `ACT`: `st7735-wwxhh-x`

workspace/__ardulib__/ST7735/XST7735.cpp
```diff
@@ -0,0 +1,200 @@
+
+#include "XST7735.h"
+
+inline uint16_t swapColor565(uint16_t color) {
```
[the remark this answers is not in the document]
gabbapeople: By now this function is used exclusively in this library so I suppose it should be left here.

workspace/__ardulib__/ST7735/XST7735.cpp
gabbapeople: It is hard to guarantee the error-free work (e.g the proper byte sequences for initialization) of display of this kind as we don't have real hardware to test on.

I suggest to leave it as it is. Let's take a look. If some user responds about an unsupported 7735 display it will be a great reason to expand the lib with a height/width constructor.

workspace/__ardulib__/ST7735/XST7735.cpp
```diff
@@ -0,0 +1,216 @@
...
+        break;
+    case 2: // RG type.
+        initTypeRG();
+        break;
+    case 3: // RR Type.
+        initTypeRR();
+        break;
+    default:
+        break;
+    }
```
nkrkv: The dependency on a value unknown at compile-time is still here. Just remove this method as the whole here. Call `initTypeXX` directly in nodes’ implementation

gabbapeople: @nkrkv done

workspace/__ardulib__/ST7735/XST7735.cpp
```diff
@@ -0,0 +1,197 @@
...
+#else
+    SPI.setClockDivider(SPI_CLOCK_DIV4); // 4 MHz (half speed).
+#endif // ARDUINO_ARCH_SAM
+    SPI.setBitOrder(MSBFIRST);
+    SPI.setDataMode(SPI_MODE0);
+#endif // SPI_HAS_TRANSACTION
+
+    // Toggle RST low to reset; CS low so it'll listen to us.
+    *csport &= ~cspinmask;
+    if (_rst) {
```
nkrkv: `_rst` should be checked against `NOT_A_PIN`, not `D0`. Do we ever need resetting? In what cases? A reference to official docs would be nice

workspace/__ardulib__/ST7735/XST7735.cpp
```diff
@@ -0,0 +1,197 @@
...
+void ST7735::initTypeG() {
+    commonInit(Gcmd);
+}
+
+void ST7735::initTypeB() {
+    commonInit(Bcmd);
+}
+
+void ST7735::initTypeRG() {
+    _type = 2;
```
nkrkv: I’d call it `_swappedColorBytes`

workspace/__ardulib__/ST7735/XST7735.cpp
```diff
@@ -0,0 +1,197 @@
...
+    commandList(Rcmd3);
+}
+
+void ST7735::initTypeRR() {
+    commonInit(Rcmd1);
+    commandList(Rcmd2red);
+    commandList(Rcmd3);
+}
+
+void ST7735::renderScanlinePart(int16_t scanline, int16_t xmin, int16_t xmax, uint16_t* lineBuffer) {
```
nkrkv:
```suggestion
void ST7735::renderScanlinePart(int16_t scanline, int16_t xmin, int16_t xmax, const uint16_t* lineBuffer) {
```

workspace/__ardulib__/ST7735/XST7735.h
```diff
@@ -0,0 +1,277 @@
...
+class ST7735 : public XRenderer {
+public:
+    ST7735(uint8_t cs, uint8_t dc, uint8_t rst);
+
+    void initTypeG();
+    void initTypeB();
+    void initTypeRG();
+    void initTypeRR();
+
+    void renderScanlinePart(int16_t scanline, int16_t xmin, int16_t xmax, uint16_t* lineBuffer);
```
nkrkv:
```suggestion
    void renderScanlinePart(int16_t scanline, int16_t xmin, int16_t xmax, const uint16_t* lineBuffer) override;
```

workspace/__ardulib__/ST7735/XST7735.cpp
```diff
@@ -0,0 +1,197 @@
...
+#ifdef __ARDUINO_ARC__
+    digitalWrite(_rs, HIGH);
+#else
+    *rsport |= rspinmask;
+#endif // __ARDUINO_ARC__
+
+    *csport &= ~cspinmask;
+    for (uint16_t x = 0; x <= xmax - xmin; x++) {
+        uint8_t hi = lineBuffer[x] >> 8;
+        uint8_t lo = lineBuffer[x];
```
nkrkv: The check against `_swappedColors` may be here.

workspace/__lib__/xod-dev/st7735-display/init-b/patch.cpp
```diff
@@ -0,0 +1,18 @@
...
+
+void evaluate(Context ctx) {
+
+    auto dev = getValue<input_DEV>(ctx);
+
+    if (isSettingUp())
+        emitValue<output_DEVU0027>(ctx, dev);
+
+    if (isInputDirty<input_DEV>(ctx)) {
+        dev->initTypeB();
```
nkrkv: This should be done in the `isSettingUp` branch. Do not check `input_DEV` dirtieness.

PR #1935, opened 2019-12-26
Support of SSD1306 based monochrome I2C displays

gabbapeople:
XOD support of SSD1306 based monochrome I2C displays.

### DIFF
--- a/workspace/__ardulib__/SSD1306/SSD1306.cpp
+++ b/workspace/__ardulib__/SSD1306/SSD1306.cpp
@@ -1,0 +1,92 @@
+
+#include "SSD1306.h"
+
+SSD1306::SSD1306(TwoWire* wire, uint8_t i2cAddress, uint8_t* buffer) {
+    _wire = wire ? wire : &Wire;
+    _i2cAddress = i2cAddress;
+    _buffer = buffer;
+    _width = SSD1306_DISPLAY_WIDTH;
+    _height = SSD1306_DISPLAY_HEIGHT;
+}
+
+void SSD1306::sendCommand(uint8_t command) {
+    _wire->beginTransmission(_i2cAddress);
+    _wire->write(0x80);
+    _wire->write(command);
+    _wire->endTransmission();
+}
+
+void SSD1306::begin() {
+    _wire->begin();
+    sendCommand(SSD1306_DISPLAY_OFF);
+    sendCommand(SSD1306_SET_DISPLAY_CLOCK);
+    sendCommand(0x80);
+    sendCommand(SSD1306_SET_MULTIPLEX_RATIO);
+    sendCommand(0x3F);
+    sendCommand(SSD1306_SET_DISPLAY_OFFSET);
+    sendCommand(0x00);
+    sendCommand(SSD1306_SET_START_LINE | 0);
+    sendCommand(SSD1306_CHARGE_DCDC_PUMP);
+    sendCommand(0x14);
+    sendCommand(SSD1306_ADDR_MODE);
+    sendCommand(0x00);
+    sendCommand(SSD1306_SET_REMAP_L_TO_R);
+    sendCommand(SSD1306_SET_REMAP_T_TO_D);
+    sendCommand(SSD1306_SET_COM_PINS);
+    sendCommand(0x12);
+    sendCommand(SSD1306_SET_CONTRAST);
+    sendCommand(0xFF);
+    sendCommand(SSD1306_SET_PRECHARGE_PERIOD);
+    sendCommand(0xF1);
+    sendCommand(SSD1306_SET_VCOM_DESELECT);
+    sendCommand(0x40);
+    sendCommand(SSD1306_RAM_ON);
+    sendCommand(SSD1306_INVERT_OFF);
+    sendCommand(SSD1306_DISPLAY_ON);
+}
+
+void SSD1306::sendBuffer() {
+    sendCommand(SSD1306_ADDR_PAGE);
+    sendCommand(0);
+    sendCommand(_height / 8 - 1);
+    sendCommand(SSD1306_ADDR_COLUMN);
+    sendCommand(0);
+    sendCommand(_width - 1);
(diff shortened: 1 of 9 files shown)


workspace/__ardulib__/SSD1306/SSD1306.cpp
```diff
@@ -0,0 +1,103 @@
+
+#include "SSD1306.h"
+
+SSD1306::SSD1306(TwoWire* wire, uint8_t i2cAddress, uint8_t width, uint8_t height) {
+    _wire = wire ? wire : &Wire;
+    _i2cAddress = i2cAddress;
+    _width = width;
+    _height = height;
+    _buffer = new uint8_t[_width * _height / 8];
```
brusherru: I have an idea, but I'm not sure that this is a popular case... Here it is:
- let's allocate memory outside this library and just accept the pointer to the buffer of the desired size (probably we need to check it here?)
- then we can allocate the buffer right in one of XOD nodes. E.G. `xod-dev/ssd1306/buffer`
- then this buffer might be reused by multiple displays of the same sizes (but only one should update it === call `render` method)
- and we don't need a deconstructor

nkrkv: I agree with @brusherru for the point that the buffer pointer should be passed from outside, as an argument. But not for the reason mentioned (which is quite synthetic). If you’d define the buffer as a part of `State` in the node it will be stored in the data segment, not the heap. So, it will be clearly reported by the compiler as a thing which consumes RAM whereas in this case the pretty large consumption is masked.

workspace/__ardulib__/SSD1306/SSD1306.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+void SSD1306::renderScanlinePart(int16_t scanline, int16_t xmin, int16_t xmax, const uint16_t* lineBuffer) {
+    if ((scanline >= _height) || (scanline < 0))
+        return;
+    if ((xmin < 0) || (xmax < 0) || (xmin >= _width) || (xmax >= _width) || (xmin > xmax))
+        return;
+
+    for (int16_t x = 0; x < _width; x++) {
+
+        bool _color = lineBuffer[x] < 0x7FFF ? 0 : 1;
```
nkrkv: So, blue is black and red is white. Not very intuitive. It’s better to rely on lightness. Of course, converting to HSL will kill the performance here but given that the color is stored as 565 RGB we can make an approximation. Just strip out the least significant bits of each component and check whether something is left:

```
RRRRRGGGGGGBBBBB &
1100011000011000 =
RR000GG0000BB000
```

So:

```cpp
bool color = lineBuffer[x] & 0xC618;
```

BTW, local vars should not start with an underscore.

workspace/__lib__/xod-dev/ssd1306-display/ssd1306-128x64-i2c-device/patch.cpp
```diff
@@ -0,0 +1,43 @@
...
+    uint8_t addr = getValue<input_ADDR>(ctx);
+
+    if (addr > 127) {
+        raiseError(ctx);
+        return;
+    }
+
+    Type dev = new (state->mem) SSD1306(i2c, addr);
+
+    state->buffer = new uint8_t[dev->getScreenWidth() * dev->getScreenHeight() / 8];
```
nkrkv: Take two :clapper: 

Do not use `new` to allocate the buffer. Keep it in `State`

```cpp
struct State {
    uint8_t mem[sizeof(SSD1306)];
    uint8_t buffer[128 * 64 / 8];
};
```

workspace/__ardulib__/SSD1306/SSD1306.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+void SSD1306::sendCommand(uint8_t command) {
+    _wire->beginTransmission(_i2cAddress);
+    _wire->write(0x80);
+    _wire->write(command);
+    _wire->endTransmission();
+}
+
+void SSD1306::setBuffer(uint8_t* buffer) {
+    _buffer = buffer;
```
nkrkv: Does it make sense to keep a dedicated method? One can forget to call it or change in the middle of the object lifetime, is it expected. I’d pass `buffer` to the constructor.

workspace/__ardulib__/SSD1306/SSD1306.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+void SSD1306::renderScanlinePart(int16_t scanline, int16_t xmin, int16_t xmax, const uint16_t* lineBuffer) {
+    if ((scanline >= _height) || (scanline < 0))
+        return;
+    if ((xmin < 0) || (xmax < 0) || (xmin >= _width) || (xmax >= _width) || (xmin > xmax))
+        return;
+
+    for (int16_t x = 0; x < _width; x++) {
+
+        bool color = lineBuffer[x] & 0xC618;
```
nkrkv: Ooff… a tiny note about what does `0xC618` mean would be nice. As it looks like clear magic.

workspace/__lib__/xod-dev/ssd1306-display/project.xod
```diff
@@ -0,0 +1,6 @@
+{
+  "description": "Nodes to drive SD1306 based monochrome LCD displays with I2C interfaces.",
```
nkrkv:
```suggestion
  "description": "Nodes to drive SD1306-based monochrome LCDs with I2C interface.",
```

- D in LCD is “Display” already

workspace/__lib__/xod-dev/ssd1306-display/ssd1306-128x64-i2c-device/patch.cpp
```diff
@@ -0,0 +1,40 @@
+#pragma XOD error_raise enable
```
nkrkv:
```suggestion
```

In straightforward implementations like this one, the use of `raiseError` is detected automatically. KISS.

PR #1937, opened 2020-01-13
Add patterns and their support for WS2812

brusherru:
There is no issue.


workspace/__lib__/xod/color-pattern/pattern/patch.cpp
```diff
@@ -0,0 +1,20 @@
...
+
+{{ GENERATED_CODE }}
+
+void evaluate(Context ctx) {
+    auto pattern = getValue<input_PAT>(ctx);
+    auto color = getValue<input_C>(ctx);
+    auto state = getState(ctx);
+
+    state->node.color = color;
+    pattern->add(&state->node);
```
nkrkv: Will change the `pattern` multiple times if, say, `input_C` changes

workspace/__ardulib__/XColorPattern/XColorPattern.h
```diff
@@ -0,0 +1,46 @@
+#ifndef __X_COLOR_PATTERN_H__
+#define __X_COLOR_PATTERN_H__
+
+using xod::XColor;
+
+struct PatternNode {
+    XColor color;
+    PatternNode* _next = NULL;
```
nkrkv: Is it a struct (that is, all fields are public and dumb by design)? Or is it a class (members are hidden, the interface exposed through methods)?

It’s an inconsistent mix now. Please, stick to a single camp. Looking at the `Pattern` I recommend to follow the “class” way.

workspace/__lib__/xod/color-pattern/project.xod
```diff
@@ -0,0 +1,6 @@
+{
+  "description": "The color pattern type is a Linked List of colors. This type might be used by WS2812 for lighting up a pattern on LED matrix and etc.",
+  "license": "AGPL-3.0",
+  "name": "color-pattern",
+  "version": "0.1.0"
+}
```
nkrkv: The pattern is soo specific to WS2812 now, that it would be better to merge it to `xod-dev/ws2812`. If this would be usable with other devices (which?), we can extract it back.

workspace/__ardulib__/XColorPattern/XColorPattern.h
```diff
@@ -0,0 +1,46 @@
+#ifndef __X_COLOR_PATTERN_H__
+#define __X_COLOR_PATTERN_H__
```
nkrkv: Wouldn’t it be much simpler to embed the types to the constructor node implementation instead of a dedicated “Third party” library?

brusherru: Until the `WS2812` knows something about `Pattern` type (even if I implement `fill-pattern` node inside NIIX instead of calling `fillPattern` method) — NO.

However, if I combine XColorPattern with the `WS2812` (like you said in [https://github.com/xodio/xod/pull/1937#discussion_r366311550](previous comment)) I can combine the library too, so `XColorPattern.h` will be a part of `WS2812_NoBuffer` c++ library.

nkrkv: > so XColorPattern.h will be a part of WS2812_NoBuffer c++ library.

Sounds good

brusherru:
@nkrkv done! Check it out, please

workspace/__ardulib__/WS2812_NoBufffer/WS2812.h
```diff
@@ -29,6 +32,9 @@ class WS2812 {
 public:
     WS2812(uint8_t pin, uint32_t length);
     void fill(uint8_t r, uint8_t g, uint8_t b);
+    void fill(XColor color);
+    void fill(XColor color, uint32_t pixelsCount, bool fromTail = false);
```
nkrkv:
```suggestion
    void fill(XColor color, uint32_t pixelCount, bool fromTail = false);
```

Not critical though

Issue #1938, opened 2020-01-17
Pulse-on-change type byte


Cesar-S:
Please add `pulse-on-change` type byte.

I have it created in a personal library, but since it is now generic it would be better to be native in XOD

thanks

nkrkv:
Good idea!

Do you feel the power to make a pull request to the `xod/core`?

Cesar-S:
> Do you feel the power to make a pull request to the xod/core?

I have desktop Github, I could try it although I don't know much about working with github.

nkrkv:
OK, feel free to try to make a PR. If you have problems we’ll implement the issue ourselves.

PR #1939, opened 2020-01-17
added pulse-on-change(byte)

Cesar-S:
added pulse-on-change (byte) to generic type

close [#1938](https://github.com/xodio/xod/issues/1938)

### DIFF
--- a/workspace/__lib__/xod/core/pulse-on-change(byte)/patch.cpp
+++ b/workspace/__lib__/xod/core/pulse-on-change(byte)/patch.cpp
@@ -1,6 +1,6 @@
 
 struct State {
-    Number sample = NAN;
+    uint8_t sample = 0;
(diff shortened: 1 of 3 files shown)


workspace/__lib__/xod/core/pulse-on-change(byte)/patch.cpp
```diff
@@ -0,0 +1,16 @@
+
+struct State {
+    Number sample = NAN;
```
evgenykochetkov:
```suggestion
    uint8_t sample = 0;
```

I'd change the `sample` type to `uint8_t` to save a little bit of memory.

Everything else looks good 👍🏻

nkrkv:
Thank you for contributing! Will be released with 0.33.0

PR #1940, opened 2020-01-22
changes nth-input and switch-case

Cesar-S:
Hi, this is only just a reduction in memory consumption.

Here I have taken as a sample 4 examples with variable x 10, and the differences when compiling.
Perhaps a node formed by others is preferred by way of example instead of one in C ++.
I only make it available, you can discard it if it is not convenient.

Compiled "switch-case" (xod) x 10

1754 bytes of program storage space
85 bytes of dynamic memory

Compiled "switch-case" (personal-library) x 10

1220 bytes of program storage space
63 bytes of dynamic memory

Compiled "nth-input" (xod) x 10

3340 bytes of program storage space
138 bytes of dynamic memory

Compiled "nth-input" (personal-library) x 10

2276 bytes of program storage space
106 bytes of dynamic memory

[changes.zip](https://github.com/xodio/xod/files/4095217/changes.zip)

### DIFF
--- a/workspace/__lib__/xod/core/nth-input/patch.cpp
+++ b/workspace/__lib__/xod/core/nth-input/patch.cpp
@@ -1,0 +1,21 @@
+
+#pragma XOD dirtieness disable
+
+struct State {};
+
+int cond;
+
+{{ GENERATED_CODE }}
+
+void evaluate(Context ctx) {
+    auto x = getValue<input_IDX>(ctx);
+    auto y = 1;
+    if (x < y){
+     cond = 1;
+    }else{cond = 0;}
+
+    auto trueVal = getValue<input_X0>(ctx);
+    auto falseVal = getValue<input_X1>(ctx);
+    emitValue<output_OUT>(ctx, cond ? trueVal : falseVal);
+    emitValue<output__>(ctx, x - 1);
+}
--- a/workspace/__lib__/xod/core/nth-input/patch.xodp
+++ b/workspace/__lib__/xod/core/nth-input/patch.xodp
@@ -1,122 +1,31 @@
 {
   "description": "Selects either `X0` or `X1` based on `IDX` index value.",
-  "links": [
-    {
-      "id": "ByT8JLshM",
-      "input": {
-        "nodeId": "r1ctYrM3G",
-        "pinKey": "__in__"
-      },
-      "output": {
-        "nodeId": "HyXrE1Ls2M",
-        "pinKey": "S13xLCuHvkW"
-      }
-    },
-    {
-      "id": "BysSVkUo2M",
-      "input": {
-        "nodeId": "SyIHNkUinz",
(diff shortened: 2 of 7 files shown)


nkrkv:
The optimization is indeed notable, thank you for PR’ing.

What annoys me a bit is that in such cases we always have to make a hard choice: either keep on the XOD land and be less efficient or surrender and fall back to (not always easy to understand C++). I wonder, can we do something so that the XOD implementation is kept (to define the intent and the “clear” implementation) which _is replaced_ with C++ for the sake of optimization?

As far as I remember, one _may_ leave all nodes as-is, then place the `not-implemented-in-xod`, and the latter will take the precedence. This should be checked out for this PR, I think. That is, try to just roll back the deletion of the existing nodes. If that works, the only thing to think about is how to guarantee that the XOD implementation is always correct. For example, tabtests might somehow test against the C++ implementation _and_ XOD implementation.

@evgenykochetkov @brusherru What do you think?

Cesar-S:
If I understand perfectly the objective, and the idea that is sought with XOD.  

Actually abandon all known programs and sometimes I only use the Arduino IDE for testing.  

But the ease of use that XOD provides as a reusable code and the possibility of "seeing" that makes arduino. 

 I found situations where the lack of resources does not allow me a simulation, I do not remember but a percentage of 15% is needed.  

There are nodes that, thanks to the XOD scheme, manage to pass them to C ++ and learn.  I will try to reverse and see what happens.  

Thanks for the reply.

Cesar-S:
😮 it has worked, even switch-case reduce even more, I will upload the changes.

switch-case x 10 
1212
63

Cesar-S:
no-implemented-in-xod only reduce for switch-case, 
I leave the idea ... I don't understand github's operation well and I'm making awkward mistakes.

Cesar-S:
Before reloading, I would like to solve this, but it is an appropriate place and I should create an issue, please tell me
Something strange happens, I try to find out what it can be but I can't.

**The problem**: 
nth-input vs [dataselect](https://xod.io/libs/cesars/utils/) x 10

I edit the `nth-input` node, add non-implemented ... compile and the amount of resources used does not change  **3340/138**
I remove the nodes and leave the pins and stay the same **3340/138**

I delete all the content, copy it from my [dataselect](https://xod.io/libs/cesars/utils/) node and remain the same 3340/138

I create a new node, copy the [dataselect](https://xod.io/libs/cesars/utils/) content, compile = **2276/106**
I add the `nth-input` nodes and connect to the pins, compile = **3340/138**

Now, what I can't solve.
If I delete everything from the node and copy the  [dataselect](https://xod.io/libs/cesars/utils/) content, compile = **3340/138**

Maybe I should save-reload, but this shouldn't happen

It is as if the content of `nth-input` affects the new node and no longer reduces the consumption of resources

PR #1941, opened 2020-01-22
update switch-case and nth-input

Cesar-S:
by mistake delete the draft files [#1940](https://github.com/xodio/xod/pull/1940)


Cesar-S:
no-implemented-in-xod only reduce for switch-case

PR #1943, opened 2020-01-24
Optimise handling of dirty inputs and outputs in generated C++

evgenykochetkov:
The first commit introduces the `evaluate_on_pin` pragma.

It deals with a common pattern when a node is evaluated only when a special input (usually named `UPD`) is dirty:

```
// ...
void evaluate(Context ctx) {
    if (!isInputDirty<input_UPD>(ctx))
        return;

    // some actual work
}
```

When other inputs receive a new value, the node is marked as dirty and evaluated, only to immediately return from `evaluate` function.

With `evaluate_on_pin` pragma, we can specify on which pin updates will result in node evaluation and avoid generating useless code.
```
// disable evaluation on all pins
#pragma XOD evaluate_on_pin disable
// ... except for the UPD pin
#pragma XOD evaluate_on_pin enable input_UPD

// ...
```

---

The second commit removes `node_NN_isOutputDirty_XXX` fields from transaction state for unlinked outputs, because they are never used for anything meaningful, only copied in and out of context objects.


evgenykochetkov:
Completely agree about pragma naming and convention.

> The more important thing which needs more attention is why runTransaction in the blink fixture has not changed? Intuitively I would expect that it should lose dirtiness propagation from flip-flop MEM to led LUM but I’m not sure.

The dirtiness propagation is gone, just as expected: https://github.com/xodio/xod/commit/dc595ca094abb099240ea7c1e8ef40031f7a1d7b#diff-201e7657cc1fa162e7f4cab98b0ef532L1551

nkrkv:
> The dirtiness propagation is gone, just as expected: dc595ca#diff-201e7657cc1fa162e7f4cab98b0ef532L1551

Oh, my bad, missed it

PR #1950, opened 2020-02-03
Add the Triangle graphic primitive to the Graphic library.

gabbapeople:
Add the Triangle class to the `__ardulib__/Graphics/` and `xod/graphics` library.

### DIFF
--- a/workspace/__ardulib__/Graphics/Triangle.inl
+++ b/workspace/__ardulib__/Graphics/Triangle.inl
@@ -1,0 +1,137 @@
+/*
+ * This is the common algorithm for drawing filled triangles. The original triangle is divided
+ * into two triangles with a flat bottom base and a flat top base. The resulting triangles
+ * are drawn sequentially. A good place to read about the algorithm "http://www.sunshine2k.de/coding/java/TriangleRasterization/TriangleRasterization.html".
+ * This implementation is augmented by the scanline intersection check during rasterization.
+ */
+
+void putTriangleSolidOnScanline(XVector2<int16_t> a, XVector2<int16_t> b, XVector2<int16_t> c, int16_t scanline, uint16_t* buffer, size_t bufferSize, uint16_t color) {
+
+    int16_t xa, xb;
+    int16_t line, lastLine;
+
+    int16_t dx01 = b.x - a.x;
+    int16_t dy01 = b.y - a.y;
+    int16_t dx02 = c.x - a.x;
+    int16_t dy02 = c.y - a.y;
+    int16_t dx12 = c.x - b.x;
+    int16_t dy12 = c.y - b.y;
+
+    int32_t sa = 0;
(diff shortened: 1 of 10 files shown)


workspace/__ardulib__/Graphics/Triangle.inl
```diff
@@ -0,0 +1,84 @@
+
+inline int16_t crossProduct2D(XVector2<int16_t> a, XVector2<int16_t> b) {
+    return a.x * b.y - b.x * a.y;
+}
```
nkrkv: Shouldn’t it be a method (or operator) of `XVector2`?

workspace/__ardulib__/Graphics/Triangle.inl
```diff
@@ -0,0 +1,84 @@
...
+
+/*
+ * The Barycentric Algorithm. This triangle rasterization algorythm decides if a point lies inside or outside a triangle.
+ * A good place to read about http://www.sunshine2k.de/coding/java/TriangleRasterization/TriangleRasterization.html.
+ * This implementation is augmented by the scanline intersection check during rasterisation.
+ */
+void putTriangleSolidOnScanline(XVector2<int16_t> a, XVector2<int16_t> b, XVector2<int16_t> c, int16_t scanline, uint16_t* buffer, size_t bufferSize, uint16_t color) {
```
brusherru: Shouldn’t it be a method (or operator) of `TriangleSolid`?

gabbapeople:
Changed the entire algorithm to the faster one.

workspace/__lib__/xod/graphics/triangle-outline-colored/patch.cpp
```diff
@@ -0,0 +1,51 @@
...
+    int16_t x1 = (int16_t)getValue<input_X1>(ctx);
+    int16_t y1 = (int16_t)getValue<input_Y1>(ctx);
+    int16_t x2 = (int16_t)getValue<input_X2>(ctx);
+    int16_t y2 = (int16_t)getValue<input_Y2>(ctx);
+
+    if (isSettingUp()) {
+        state->triangleOutlineColored = new (state->mem) TriangleOutline(gfx);
+    }
+
+    if (isSettingUp() || x0 != state->x0 || y0 != state->y0 || x1 != state->x1 || y1 != state->y1 || x2 != state->x2 || y2 != state->y2 || isInputDirty<input_C>(ctx)) {
```
brusherru: If anything changed — update state and emit updated triangle... looks like we can get rid of this condition (in each triangle node). Or I missed something?

gabbapeople: That is done to prevent node evaluate on number (float) changes.
Coorinates are integers only. For example, the change  `5.67` -> `5.68` should not trigger node evaluate because it is still `5`.

PR #1953, opened 2020-02-06
Add color picker widget

### LINKED ISSUE #1945: As a xoder I want to set color values with a picker control so that I can see the visual feedback immediatelly
The current color literals look like hex web-colors (`#12е456`). Without a visual hint, it’s hard to imagine what this color is. Compare to: (`#1294e6`).

To be more WYSIWYG we can complement the color inputs with color preview and an edit popup. Here’s our discussion whiteboard shot:

### User story

- I select a node having color inputs
- I see a little square with a color that matches the current bound value
- I focus the input and nothing changes (comparing to the current behavior)
- I click the square and a popup appears with the geometry and behavior similar to the helpbox popup
- In this popup, I see a color picker control which allows me to adjust H, S, L components with sliders or text inputs
- The changes are applied to the input value in real-time, without any OK/Apply buttons
- If I screwed up a value, I can click “Revert” to roll back to the color which was set at the moment when the popup has appeared (the popup itself is kept open)
- I can hide the picker with “(x)” button in the top-left or by giving focus to another element of Inspector

### Acceptance criteria

- [x] The basic scenario works as described
- [ ] The same experience is applied to generic inputs deduced to the color type
- [x] The popup hides automatically if the hosting input goes out of the visible area
- [x] The changes while dragging color sliders are immediately visible on a physical device if adjusted in the debug mode.

brusherru:
It partially closes #1945 
The rest of the issue and additional tweaks is described in the new issue: #1954


nkrkv:
Isn’t #1945 the issue?

nkrkv:
I’d forbid text selection in the picker popup. It’s quite easy to cause the glitch while dragging the wheel:

image

Cesar-S:
this looks great :clap: :clap:

nkrkv:
Is there any chance to apply the picker to generic inputs? If not, please, extract the issue about that into a separate issue.

nkrkv:
The same about “Revert” mentioned in #1945

brusherru:
@nkrkv done! The issue mentioned in the PR description.

packages/xod-client-electron/src/debugger/sendToSerialMiddleware.js
```diff
@@ -20,7 +20,8 @@ export default ({ getState }) => next => action => {
 
   const isTweakActionType =
     action.type === client.NODE_UPDATE_PROPERTY ||
-    action.type === client.TWEAK_PULSE_SENT;
+    action.type === client.TWEAK_PULSE_SENT ||
+    action.type === client.NODE_PROPERTY_TWEAKED;
```
evgenykochetkov: I think that `NODE_PROPERTY_TWEAKED` name will cause some confusion in the future, but I can't think of a better alternative yet. Probably, it should reflect that it is an intermediate value and is not intended to end up in a redux store.

brusherru: What about `SEND_TWEAK_VALUE_TO_INTERACTIVE_SESSION`?

nkrkv: `SEND_TWEAK_VALUE_TO_INTERACTIVE_SESSION` — no, please. Action names are the _facts_. System parts may _subscribe_ to _facts_ they are interested in. `SEND_TWEAK_VALUE_TO_INTERACTIVE_SESSION` is not much better than calling `sendValueToDebugger` right inside the component.

I see what @evgenykochetkov means but it requires a minor clarification. Use present continuous, for example (`NODE_PROPERTY_TWEAKING`)… Or to match the existing name: `NODE_UPDATING_PROPERTY`. Better it would be `NODE_PROPERTY_UPDATING` but we already have a somewhat malformed name for `NODE_UPDATE_PROPERTY`.

brusherru: @nkrkv @evgenykochetkov just renamed these two action types.
- `NODE_UPDATE_PROPERTY` -> `NODE_PROPERTY_UPDATED` (see last commit)
- `NODE_PROPERTY_TWEAKED` -> `NODE_PROPERTY_UPDATING` (fixup)

nkrkv:
Merge?

PR #1956, opened 2020-02-10
Add the text support to the Graphic library.

gabbapeople:
- Add the Text class to the `__ardulib__/Graphics/` and `xod/graphics` library.
- Add the common 5x7 GLCD ASCII font.


workspace/__ardulib__/Graphics/Text.h
```diff
@@ -0,0 +1,37 @@
...
+
+    XColor* _textColor = nullptr;
+    uint8_t _textSize = 1;
+
+public:
+    Text(XGraphics* parent, char* text, size_t textLen);
+
+    void setPosition(int16_t x, int16_t y);
+    void setStyle(XColor* textColor);
+    void setStyle(uint8_t textSize);
```
nkrkv: And what if we’ll want to add letter-spacing or line-height later? Would you add the following?

```cpp
void setStyle(uint8_t letterSpacing);
void setStyle(uint8_t lineHeight);
```

OOops. Ambiguous overload error: `setStyle(4)`. That’s why I’m sure calling everyting `setStyle` instead of `setColor`, `setWeight`, `setTextSize`, `setLineHeight`, etc, etc is a bad idea. A slightly better decision would be:

```cpp
setStyle(XColor* textColor, uint8_t textSize, uint8_t letterSpacing, uint8_t lineHeight /* , ... */);
```

But it does not scale well.

Suggest clear names that reflect what’s going been set.

workspace/__ardulib__/Graphics/Text.inl
```diff
@@ -0,0 +1,46 @@
...
+                    if (scanline >= cursorY && scanline < cursorY + _textSize) {
+
+                        int16_t cursorX = (c * 6 * _textSize) + _pivot.x + i * _textSize;
+                        for (int16_t x = cursorX; x < cursorX + _textSize; x++)
+                            if (x >= 0 && x < bufferSize)
+                                buffer[x] = color;
+                    }
+                }
+        }
+    }
```
nkrkv: For each nesting level, please, add a comment on what’s going on.

To lower the nesting depth replace two big `if`’s with early `continue`:

```cpp
for (...) {
  if (foo) {
    ...
    if (bar) {
       ...
    }
  }
}

// to

for (...) {
  if (!foo) {
    // what happened?
    continue;
  }

  ...

  if (!bar) {
    // why do we continue?
    continue;
  }
  ...
}
```

workspace/__lib__/xod/graphics/text-colored/patch.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+    if (isSettingUp()) {
+        state->cap = getValue<input_L>(ctx);
+        state->str = new char[state->cap + 1];
+        state->textColored = new (state->mem) Text(gfx, state->str, state->cap);
+    }
+
+    auto str = getValue<input_S>(ctx);
+    memset(state->str, '\0', state->cap + 1);
+    dump(str, state->str);
```
nkrkv: :scream: `XString` is a memory-efficient string representation. And you ruin it here, basically, creating a big-buffer copy of everything. Why can’t you just use `XString` directly?

workspace/__ardulib__/Graphics/Text.h
```diff
@@ -0,0 +1,38 @@
...
+ * By now this class uses only the standard 5x7 GLCD ASCII font.
+ * The `setTextColor` method sets up text color otherwise
+ * the default foreground color of the scene is used.
+ * The `setTextScale` method sets up the value that scales the font.
+ */
+class Text : public XGraphics {
+private:
+    XVector2<int16_t> _pivot;
+
+    XString* _str = nullptr;
```
nkrkv: No need to use pointers for `XString`. It’s a thin scalar wrapper that is safe to pass by value. If not set it iterates as an empty string.

workspace/__lib__/xod/graphics/text-colored/patch.cpp
```diff
@@ -0,0 +1,51 @@
...
+{{#global}}
+#include <Text.h>
+{{/global}}
+// clang-format on
+
+struct State {
+    uint8_t mem[sizeof(Text)];
+    Text* textColored;
+
+    XString str;
```
nkrkv: No need to keep in `State`...

workspace/__lib__/xod/graphics/text-colored/patch.cpp
```diff
@@ -0,0 +1,51 @@
...
+        state->textColored = new (state->mem) Text(gfx, &state->str);
+    }
+
+    if (isSettingUp() || state->x != x || state->y != y || state->textScale != textScale || isInputDirty<input_S>(ctx) || isInputDirty<input_C>(ctx)) {
+        state->x = x;
+        state->y = y;
+        state->textScale = textScale;
+        state->textColored->setPosition(x, y);
+        state->textColored->setTextScale(textScale);
+        state->textColored->setTextColor(&state->textColor);
```
nkrkv: ...pass it here with:

```cpp
state->textColored->setText(getValue<input_S>(ctx));
```

workspace/__ardulib__/Graphics/Text.inl
```diff
@@ -0,0 +1,66 @@
...
+
+    uint16_t color = xcolorTo565(_textColor ? *_textColor : getFGColor());
+
+    uint8_t charCount = 0;
+
+    // For each char in the string.
+    for (auto it = _str.iterate(); it; ++it, charCount++) {
+
+        // Render only ASCII symbols from 32 to 127 else render the "?".
+        uint8_t charNum = ((int)*it < 32 || (int)*it > 127) ? 63 : (int)*it;
```
nkrkv: Unreasonable “uint_8t / char / int” type twist in a single expression. All is simpler:

```cpp
char rawChar = *it;
char renderChar = (rawChar < 32 || rawChar > 128) ? '?' : rawChar;
```

workspace/__lib__/xod/graphics/text-colored/patch.xodp
```diff
@@ -0,0 +1,91 @@
+{
```
nkrkv: A description of the node is missing.

workspace/__lib__/xod/graphics/text-colored/patch.xodp
```diff
@@ -0,0 +1,91 @@
...
+      "label": "GFX",
+      "position": {
+        "units": "slots",
+        "x": 0,
+        "y": 0
+      },
+      "type": "@/input-graphics"
+    },
+    {
+      "description": "X coordinate of the pivot point of the text on the canvas.",
```
nkrkv:
```suggestion
      "description": "Left point of the text on the canvas.",
```

workspace/__lib__/xod/graphics/text-colored/patch.xodp
```diff
@@ -0,0 +1,91 @@
...
+      "label": "SCL",
+      "position": {
+        "units": "slots",
+        "x": 4,
+        "y": 0
+      },
+      "type": "xod/patch-nodes/input-number"
+    },
+    {
+      "description": "Y coordinate of the pivot point of the text on the canvas.",
```
nkrkv:
```suggestion
      "description": "Top point of the text on the canvas.",
```

workspace/__lib__/xod/graphics/text-colored/patch.xodp
```diff
@@ -0,0 +1,91 @@
...
+        "x": 0,
+        "y": 2
+      },
+      "type": "@/output-graphics"
+    },
+    {
+      "boundLiterals": {
+        "__out__": "7"
+      },
+      "description": "Text scale size. The scale ratio of the text. Is greater 1.",
```
nkrkv:
```suggestion
      "description": "Text scale. A simple point-based magnification algorithm is used, so the value should be an integer greater or equal to 1.",
```

nkrkv:
Replicate to the `text` node as well, please

PR #1957, opened 2020-02-11
Add the `pulse-on-change` node specialization for graphics.

gabbapeople:
Add the `pulse-on-change(graphics)` node into the `xod/core` library.
`pulse-on-change(graphics)` is necessary for the proper work of the `act` nodes inside quickstart display nodes.


nkrkv:
It’d be better to place it in `xod/graphics`. Isn’t it?

gabbapeople:
@nkrkv Yes, indeed. Done

PR #1963, opened 2020-02-14
Fix throttle node: pass the constant values on setting up

brusherru:
There is no issue, but I've found the bug while testing quick-start nodes for `ws2812`. Without this fix, `throttle` node returns the default value for the data type until the input values do not change.

Tabtests and consumer nodes keep working.

### DIFF
--- a/workspace/__lib__/xod/core/throttle/patch.cpp
+++ b/workspace/__lib__/xod/core/throttle/patch.cpp
@@ -12,8 +12,6 @@ struct State {
 {{ GENERATED_CODE }}
 
 void evaluate(Context ctx) {
-    if (isSettingUp()) return;
-
     auto state = getState(ctx);
     auto inValue = getValue<input_IN>(ctx);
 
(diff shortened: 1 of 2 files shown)


workspace/__lib__/xod/core/throttle/patch.cpp
```diff
@@ -12,10 +12,13 @@ struct State {
 {{ GENERATED_CODE }}
 
 void evaluate(Context ctx) {
-    if (isSettingUp()) return;
+    auto inValue = getValue<input_IN>(ctx);
+    if (isSettingUp() && isInputDirty<input_IN>(ctx)) {
```
evgenykochetkov: I'm not sure there should be a special case for `isSettingUp` at all.

nkrkv: So, what should be the starting output value? Something bound to output (which we are going to make immutable someday) or the initial value of the input? I think the latter has more sense. Anyway, I can’t understand why the `isInputDirty<input_IN>(ctx)` check is necessary in addition to `isSettingUp`?

PR #1966, opened 2020-02-18
Upgrade bs-platform to 7.1.1

evgenykochetkov:
There are a lot of new cool features:
https://bucklescript.github.io/blog/2019/11/18/whats-new-in-7
https://bucklescript.github.io/blog/2020/02/04/release-7-1-0

No reason to stay on older version.


evgenykochetkov:
> Now, Electron to the latest? 😬

Yep. Also, our React version is quite ancient. And then there's `arduino-cli`...

PR #1967, opened 2020-02-18
Add tool for publishing stdlib

evgenykochetkov:
To automate [this part](https://github.com/xodio/xod/wiki/Releasing-a-New-Version#update-xod-libraries) of the release process.

### DIFF
--- a/tools/publish-stdlib.sh
+++ b/tools/publish-stdlib.sh
@@ -1,0 +1,11 @@
+#!/usr/bin/env bash
+
+set -e
+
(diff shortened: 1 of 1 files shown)


tools/publish-stdlib.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
+set -e
+
```
knopki: May be add something like?..
```bash
: "${XOD_USERNAME:?Set XOD_USERNAME, stupid}"
: "${XOD_PASSWORD:?Set XOD_PASSWORD, stupid}"
```

PR #1977, opened 2020-03-03
Add the images support to the Graphic library.

gabbapeople:
- Add the Image class to the __ardulib__/Graphics/.
- Add `image` node to the `xod/graphics` library.
- Add the `icons` library with test bitmaps.


workspace/__ardulib__/Graphics/Image.h
```diff
@@ -0,0 +1,42 @@
...
+#define pgm_read_byte(addr) (*(const unsigned char *)(addr))
+#endif
+
+#include "XGraphics.h"
+
+class Image : public XGraphics {
+private:
+    BBox _imageBBox;
+
+    uint8_t* _bitmapBuffer;
```
brusherru: It looks like pointer to the buffer, width, and height stored twice.
The first time in the `Bmp` struct, the second time in the instances of Image.

I'm not sure, but probably we can replace them with the single pointer to the `Bmp` struct? And rename `setBitmap` method to `linkBitmap` :)

workspace/__lib__/xod/graphics/image/patch.cpp
```diff
@@ -0,0 +1,46 @@
...
+    auto bmp = getValue<input_BMP>(ctx);
+
+    int16_t x = (int16_t)getValue<input_X>(ctx);
+    int16_t y = (int16_t)getValue<input_Y>(ctx);
+    int16_t w = (int16_t)getValue<input_W>(ctx);
+    int16_t h = (int16_t)getValue<input_H>(ctx);
+
+    if (isSettingUp()) {
+        state->image = new (state->mem) Image(gfx);
+        state->image->setBitmap(bmp.bitmapBuffer, bmp.bitmapWidth, bmp.bitmapHeight);
```
brusherru: What if we want to dynamically replace one image with another one, using `select`/`if-else`/etc node?
For example, imagine that you want to show spinner (or other "loading" picture), while the device connects to the API, and then show icon of the sun/rain/cloud/etc?

workspace/__ardulib__/Graphics/Image.h
gabbapeople: It is done for tiling or displaying only a part of image.
For example the size of the source bitmap is fixed 50x50. Image field is 100x100. Image should tile to four bitmaps for both axes.

workspace/__ardulib__/Graphics/Image.h
```diff
@@ -0,0 +1,42 @@
+
+#ifndef IMAGE_H
+#define IMAGE_H
+
+#ifndef PGM_SPACE
```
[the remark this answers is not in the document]
gabbapeople: Yep, thats true.

workspace/__lib__/xod/graphics/image/patch.cpp
gabbapeople: Maybe extend the evaluate condition?
```
if (isSettingUp() || x != state->x || y != state->y || w != state->w || h != state->h || isInputDirty<input_BMP>(ctx)) {
    state->image->setBitmap(bmp.bitmapBuffer, bmp.bitmapWidth, bmp.bitmapHeight);
}
```

workspace/__lib__/xod/bitmaps/16x16-worried-face-apple/patch.cpp
```diff
@@ -0,0 +1,37 @@
...
+{{/global}}
+// clang-format on
+
+struct State {};
+
+{{ GENERATED_CODE }}
+
+void evaluate(Context ctx) {
+    ValueType<output_BMP>::T bmp;
+    bmp.bitmapBuffer = apple_worried_face_16x16;
```
nkrkv: It’s OK to hold the images this way. However, I see no point in making the definition `{{#global}}`. Keep it scoped and use a simple potentially non-unique name.

nkrkv:
Oh! Twitter Emoji are CC-BY 4.0: https://github.com/twitter/twemoji

workspace/__lib__/xod/icons/bmp/patch.cpp
```diff
@@ -0,0 +1,16 @@
+
+struct Bmp {
+    uint8_t* bitmapBuffer;
+    uint16_t bitmapWidth;
+    uint16_t bitmapHeight;
+};
+
+struct State {};
+
+using Type = Bmp;
```
evgenykochetkov: Any reason this type is declared here, in `xod/icons`? It looks generic enough to be moved in `xod/graphics`.

workspace/__ardulib__/Graphics/Image.h
brusherru: Em... I told about `_bitmapWidth` and `_bitmapHeight`, which is always equal to `bitmap->width` and `bitmap-height`. So I think we don't need to copy their values here. The image field is stored in the `_imageBBox`, so it used for tiles / cropping the bitmap. Right?

So we don't need to store `_bitmapWidth` twice, I think, because `_bitmapWidth` always equals to `bitmap->width`.

workspace/__ardulib__/Graphics/Image.inl
```diff
@@ -0,0 +1,71 @@
...
+Image::Image(XGraphics* parent)
+    : XGraphics(parent) {}
+
+void Image::setImagePosition(int16_t x, int16_t y, int16_t w, int16_t h) {
+    _imageBBox.pivot = XVector2<int16_t>(x, y);
+    _imageBBox.width = w;
+    _imageBBox.height = h;
+}
+
+void Image::linkBitmap(Bitmap* bitmap) {
```
brusherru, line 11: Why not just `_bitmap = bitmap;` and use its properties instead of copying?

workspace/__ardulib__/Graphics/Bitmap.h
```diff
@@ -0,0 +1,25 @@
...
+/*
+ * A structure to hold basic bitmap parameters.
+ */
+struct Bitmap {
+    Bitmap(uint8_t* newBuffer = nullptr, uint8_t newColorDepth = 0, uint16_t newWidth = 0, uint16_t newHeight = 0, uint16_t newAlphaColor = 0x0000) {
+        buffer = newBuffer;
+        colorDepth = newColorDepth;
+        width = newWidth;
+        height = newHeight;
+        alphaColor = newAlphaColor;
```
brusherru, line 14: I still thinking that alpha color should be defined in another layer of abstraction 😈
BTW, it will save an extra 2 bytes for each bitmap, which has no transparency at all and could make a more extensible and flexible system 😛

workspace/__ardulib__/Graphics/Bitmap.h
```diff
@@ -0,0 +1,25 @@
...
+struct Bitmap {
+    Bitmap(uint8_t* newBuffer = nullptr, uint8_t newColorDepth = 0, uint16_t newWidth = 0, uint16_t newHeight = 0) {
+        buffer = newBuffer;
+        colorDepth = newColorDepth;
+        width = newWidth;
+        height = newHeight;
+    }
+
+    uint8_t* buffer;
+    uint8_t colorDepth; // 0 = Black and white bitmap, 1 = Colored bitmap, 2 = Colored bitmap with mask color;
```
nkrkv, line 19:
```suggestion
    uint8_t colorDepth; // 0 = Black and white bitmap, 1 = Colored bitmap (565 format), 2 = Colored bitmap (565) with mask color;
```

workspace/__lib__/xod/icons/mask-color/patch.cpp
```diff
@@ -0,0 +1,16 @@
...
+
+{{ GENERATED_CODE }}
+
+void evaluate(Context ctx) {
+    auto bitmap = getValue<input_IN>(ctx);
+
+    if (bitmap->colorDepth > 0) {
+        bitmap->alphaColor = getValue<input_MSK>(ctx);
+        bitmap->colorDepth = 2;
+    }
```
nkrkv, line 13: Upstream object mutation is quite a bad idea here. Imagine two `mask-color` share the same input bitmap: only one will “win”. Worse, the side-effect is hidden. The node has no input pulses to do the destructive action. I see two possible solutions:

1. Return a *copy* of the bitmap which shares the same pix buffer but with `alphaColor` and `colorDepth` fields updated.
2. Do not introduce `mask-color` at all: keep the whole image and format definition in the bitmap constructor nodes

workspace/__lib__/xod/icons/bitmap/patch.cpp
```diff
@@ -0,0 +1,16 @@
+
+// clang-format off
+{{#global}}
+#include <Bitmap.h>
+{{/global}}
+// clang-format on
+
+struct State {};
+
+using Type = Bitmap*;
```
evgenykochetkov, line 10: Not sure that `bitmap` type should live in `xod/icons`.

gabbapeople:
- Moved all content to `/xod/graphics'.

PR #1979, opened 2020-03-19
Introduce tethering internet to provide internet to device/simulation via PC

### LINKED ISSUE #1974: As a xoder I want to share my PC internet connection with board so that I can try IoT applications even on most basic board models
Roughly, all Arduino-compatible boards can be split into two categories: those with Internet-connection hardware built-in and those without it. Nanos and Unos are still widespread as beginner’s choice. They can’t connect to the Internet directly but they could use the connection of the host PC which runs XOD. Given such a feature a user could try IoT applications even on a primitive board like Arduino Nano. If the project works fine, it should be trivial to replace the controller with IoT-compatible one to get a self-consistent device.

### User story

- I place a `xod/debug/tethering-inet` node on the patch board
- It has `INET` output which I can use with any node from `xod/net` which requires Internet
- I upload with the debugger
- While the debug session is active, my board uses XOD host connection to make HTTP requests and keep TCP connections

### Acceptance criteria

- [ ] Using `xod/debug/tethering-inet` I can show the current time from https://api.xod.io/httpbin/now on an LCD screen

### How to implement

Introduce new XOD debug protocol messages to enable Internet tethering. Get inspiration from AT-command reference of the ESP8266 module protocol. Some additional prefixes and markers are likely to be required to avoid protocol clashes with tweaks and watches.

brusherru:
It closes #1974 

You can check out the `xod/debug/example-tethering-internet` or try to create your own patch and make your own requests. The second case for tests are prefer :)


brusherru:
A few little tweaks are coming in the next commits:

- [ ] optimize `handleDebugProtocolMessages` function in `xod-arduino` package
- [ ] restrict to use more than one `tethering-internet` node (error on transpiling)
- [ ] show error/warning message in a browser when the user runs a program simulation with the `tethering-internet` node

brusherru:
Also, there might be a good tweak to remove `DELAY_BETWEEN_CHUNKS` using two-way communication:
1. Data from the TCP connection are split by chunks, which looks like: `+XOD:{nodeId}:{chunkSize}:{chunkData}`
2. WASM/MCU is receiving chunks of data from the TCP connection. And NOW there is a delay between chunks to avoid overflow of Serial buffer on MCU.
3. It can be replaced with the answer of MCU/WASM to the PC when it completes the reading of the chunk. E.G. "READK OK".
4. Then PC will transmit next chunk or "TRANSMIT END" message.

packages/xod-arduino/platform/program.tpl.cpp
```diff
@@ -72,11 +72,28 @@ TransactionState g_transaction;
 
 #if defined(XOD_DEBUG) || defined(XOD_SIMULATION)
 namespace detail {
-void handleTweaks() {
-    if (XOD_DEBUG_SERIAL.available() > 0 && XOD_DEBUG_SERIAL.find("+XOD:", 5)) {
+void handleDebugProtocolMessags() {
+    {{#forLinkedTetheringInetNode nodes}}
+    bool tetheringInetNodeReceivingData = node_{{ id }}.output_INET->isReceiving();
+    {{/forLinkedTetheringInetNode}}
```
nkrkv: Potentially can lead to multiple `tetheringInetNodeReceivingData` definition error. Isn’t it?

packages/xod-arduino/platform/program.tpl.cpp
```diff
@@ -72,11 +72,28 @@ TransactionState g_transaction;
...
+      XOD_DEBUG_SERIAL.find("+XOD:", 5)
+    ) {
         int tweakedNodeId = XOD_DEBUG_SERIAL.parseInt();
 
         switch (tweakedNodeId) {
+          {{#forLinkedTetheringInetNode nodes}}
+            case {{ id }}:
+              node_{{ id }}.output_INET->beginReceiving(XOD_DEBUG_SERIAL.parseInt());
+              XOD_DEBUG_SERIAL.read(); // :
+              // The rest of data should be readen by nodes
```
```suggestion
              // The rest of data should be read by nodes
```

packages/xod-arduino/platform/program.tpl.cpp
```diff
@@ -72,11 +72,28 @@ TransactionState g_transaction;
 
 #if defined(XOD_DEBUG) || defined(XOD_SIMULATION)
 namespace detail {
-void handleTweaks() {
-    if (XOD_DEBUG_SERIAL.available() > 0 && XOD_DEBUG_SERIAL.find("+XOD:", 5)) {
+void handleDebugProtocolMessags() {
```
```suggestion
void handleDebugProtocolMessages() {
```

packages/xod-arduino/platform/program.tpl.cpp
```diff
@@ -115,8 +132,13 @@ void handleTweaks() {
...
           {{/eachLinkedTweakNode}}
         }
-
-        XOD_DEBUG_SERIAL.find('\n');
+        {{#forLinkedTetheringInetNode nodes}}
+        if (!tetheringInetNodeReceivingData) {
+          XOD_DEBUG_SERIAL.find('\n');
+        }
+        {{else}}
+          XOD_DEBUG_SERIAL.find('\n');
```
```suggestion
        XOD_DEBUG_SERIAL.find('\n');
```

Too much indentation. See the generated fixtures

workspace/__lib__/xod/debug/receive-byte/patch.cpp
```diff
@@ -0,0 +1,22 @@
...
+// clang-format off
+{{ GENERATED_CODE }}
+// clang-format on
+
+void evaluate(Context ctx) {
+    if (!isInputDirty<input_READ>(ctx))
+        return;
+
+    auto inet = getValue<input_INET>(ctx);
+    char r = 'Z';
```
nkrkv:
```suggestion
    char r;
```

Why spend extra resources to set something unused?

workspace/__lib__/xod/debug/receive/patch.cpp
```diff
@@ -0,0 +1,40 @@
...
+    XOD_DEBUG_SERIAL.print(t);
+    XOD_DEBUG_SERIAL.println(">");
+    XOD_DEBUG_SERIAL.flush();
+//     char r = '\0';
+//     if (inet->receive(&r)) {
+//         emitValue<output_CHAR>(ctx, (uint8_t)r);
+//         emitValue<output_DONE>(ctx, 1);
+//     } else {
+//         raiseError(ctx);
+//     }
```
nkrkv: Huh?

workspace/__lib__/xod/debug/tethering-inet/patch.cpp
```diff
@@ -0,0 +1,185 @@
...
+const char FAIL[] = "FAIL";
+const char ERROR[] = "ERROR";
+const char IPD[] = "IPD,";
+const char CLOSED[] = "CLOSED";
+const char CIFSR[] = "AT+CIFSR";
+const char STAIP[] = "STAIP,\"";
+const char CIPCLOSE[] = "AT+CIPCLOSE";
+
+
+class TetherinInternet {
```
nkrkv:
```suggestion
class TetheringInternet {
```

nkrkv:
Also, I’d avoid variations of the same term using “TetheringInet” only in symbol/file names, not “Tethering Internet”

packages/xod-client-electron/src/debugger/tetheringInetMiddleware.js
```diff
@@ -0,0 +1,152 @@
...
+  };
+
+  return transmit;
+};
+
+// :: String -> [String]
+// Pack data into chunks, if needed:
+// - Simple AT answers just placed inside array
+// - IPD (Incoming Package Data) is splited on chunks of max 64 bytes length
+//   with prepended prefix `+XOD:{nodeId}:{dataLength}:`
```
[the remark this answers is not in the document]
brusherru: No, it should not.
Other tweaks have different nodeId.
A tethering internet node always waits for the chunks that begin with this header. After parsing this header it will read up any characters until it read the specified in header amount of bytes (even the chunk is: `+XOD:42:14:+XOD:42:4:lala` it will take the second part as a content of the chunk).

packages/xod-arduino/platform/program.tpl.cpp
brusherru: We squash all tethering nodes into a single one.
So it should not. 🤔

workspace/__lib__/xod/debug/receive/patch.cpp
brusherru: Stale cpp 🤔

packages/xod-arduino/platform/program.tpl.cpp
```diff
@@ -107,16 +124,14 @@ void handleTweaks() {
...
                     // to run evaluate and mark all downstream nodes as dirty
                     g_transaction.node_{{ id }}_isNodeDirty = true;
                     g_transaction.node_{{ id }}_isOutputDirty_OUT = true;
                 }
                 break;
-
           {{/eachLinkedTweakNode}}
         }
-
-        XOD_DEBUG_SERIAL.find('\n');
```
nkrkv, line 119:
```cpp
if (rewindToEol)
    XOD_DEBUG_SERIAL.find('\n');
```

workspace/__lib__/xod/debug/close/patch.cpp
```diff
@@ -0,0 +1,21 @@
+#pragma XOD evaluate_on_pin disable
+#pragma XOD evaluate_on_pin enable input_CLS
+#pragma XOD error_raise enable
```
nkrkv, line 3:
```suggestion
```

Detected automatically

packages/xod-client/src/debugger/messages.js
```diff
@@ -40,3 +40,9 @@ export const SIMULATION_BUILDING = 'Building…';
 export const SIMULATION_RUNNING = 'Running simulation…';
 export const SIMULATION_LAUNCHED = 'Simulation is launched';
 export const SIMULATION_ABORTED = 'The simulation was interrupted by the user';
+
+export const DO_NOT_USER_TETHERING_INTERNET_IN_BROWSER = {
+  title: 'Tethering Internet is not available in browser',
```
nkrkv, line 45:
```suggestion
  title: 'Tethering unavailable in browser',
```

packages/xod-client/src/debugger/messages.js
```diff
@@ -40,3 +40,9 @@ export const SIMULATION_BUILDING = 'Building…';
...
+  note:
```
nkrkv, line 46:
```suggestion
  solution:
```

packages/xod-arduino/platform/program.tpl.cpp
```diff
@@ -72,11 +72,30 @@ TransactionState g_transaction;
 
 #if defined(XOD_DEBUG) || defined(XOD_SIMULATION)
 namespace detail {
-void handleTweaks() {
-    if (XOD_DEBUG_SERIAL.available() > 0 && XOD_DEBUG_SERIAL.find("+XOD:", 5)) {
+void handleDebugProtocolMessages() {
+    {{#withTetheringInetNode nodes}}
+    bool tetheringInetNodeReceivingData = node_{{ id }}.output_INET->isReceiving();
```
evgenykochetkov, line 77: During the first transaction, this line executes **before** the `tethering-inet` node's `evaluate` function, where an instance of `TetheringInternet` is created. What's stored  in `output_INET` before that? Could that cause a crash?

packages/xod-client-browser/src/containers/App.jsx
```diff
@@ -61,6 +61,10 @@ class App extends client.App {
     props.actions.openProject(props.tutorialProject);
   }
 
+  _onRunSimulation() {
+    this.onRunSimulation(true);
```
evgenykochetkov, line 65: Explicitly telling "hey, remember, we are in a browser" each time the method is called seems a bit strange. It's not like it would change during program execution.

packages/xod-client-electron/src/debugger/tetheringInetMiddleware.js
```diff
@@ -0,0 +1,153 @@
...
+});
+
+// Mutable list of chunks to send
+// New chunks are placed in the end of list
+//
+// Transmit function (created by `createTransmitter`) is shifting
+// chunks and transmit it using `sender` function, which are different
+// for Simulation and Debug mode.
+//
+// let chunksToSend = [];
```
evgenykochetkov, line 25: 👀

packages/xod-arduino/platform/program.tpl.cpp
nkrkv: Oh, good catch!

Since the node state belongs to the program global state, I’m almost sure `output_INET` is set to `nullptr` before being initialized. @brusherru Please, verify. In any case, calling something through the null pointer is an error. I have no idea how it ever passes through it though.

packages/xod-client/src/debugger/messages.js
```diff
@@ -40,3 +40,11 @@ export const SIMULATION_BUILDING = 'Building…';
 export const SIMULATION_RUNNING = 'Running simulation…';
 export const SIMULATION_LAUNCHED = 'Simulation is launched';
 export const SIMULATION_ABORTED = 'The simulation was interrupted by the user';
+
+export const DO_NOT_USER_TETHERING_INTERNET_IN_BROWSER = {
```
nkrkv, line 44:
```suggestion
export const DO_NOT_USE_TETHERING_INTERNET_IN_BROWSER = {
```

packages/xod-client/src/debugger/messages.js
```diff
@@ -40,3 +40,11 @@ export const SIMULATION_BUILDING = 'Building…';
...
+  title: 'Tethering unavailable in browser',
+  note:
+    'The tethering internet demands on features of OS, which is not supported in the browser.',
```
nkrkv, line 47:
```suggestion
    'The internet tethering requires OS features which are not available to the web-browser.',
```

packages/xod-tethering-inet/src/AtInternet.rei
```diff
@@ -0,0 +1,48 @@
...
+// State
+type t = {
+  mux: ref(bool),
+  sending: ref(sending),
+  connections: ref(openConnections),
+  events: EventEmitter.t,
+};
+
+// Creates an object with a default state
+let _getDefaultState: unit => t;
```
nkrkv, line 20: It starts with an underscore. Is it private? If so, remove from `.rei`, leave in `.re` only. If not, remove the underscore.

packages/xod-tethering-inet/src/Connection.re
```diff
@@ -0,0 +1,99 @@
...
+
+  let fromString = str =>
+    Int.fromString(str)
+    |> (
+      res =>
+        switch (res) {
+        | Some(a) when a > 0 => Result.Ok(a)
+        | _ => Result.Error(Error.INVALID_PORT(str))
+        }
+    );
```
nkrkv, line 29: This might be simplified to:

```reason
let fromString = str =>
  switch(Int.fromString(str)) {
  | Some(a) when a > 0 => Result.Ok(a)
  | _ => Result.Error(Error.INVALID_PORT(str))
  }
```

Not a big deal, though

packages/xod-tethering-inet/src/Connection.re
```diff
@@ -0,0 +1,99 @@
...
+  type t = int;
+
+  let fromString = str =>
+    Int.fromString(str)
+    |> (
+      res =>
+        switch (res) {
+        | Some(a) when a <= 7200 && a >= 0 => Result.Ok(a)
+        | Some(_) => Result.Error(Error.INVALID_KEEPALIVE(str))
+        | None => Result.Error(Error.INVALID_ARGUMENTS)
```
nkrkv, line 41: ditto

packages/xod-arduino/platform/program.tpl.cpp
brusherru: It uses static methods. Replaced with `TetheringInternet::isReceiving()` and `TetheringInternet::beginReceiving(...)` to get rid of pointer to the instance.
Thanks!

packages/xod-client-electron/src/view/containers/App.jsx
```diff
@@ -258,6 +259,10 @@ class App extends client.App {
     );
   }
 
+  _onRunSimulation() {
+    this.onRunSimulation(false);
```
evgenykochetkov, line 263: After introducing `this.isBrowser` this is not required anymore?

Issue #163, opened 2016-08-19
First-class XOD language


nkrkv:
Since we want XOD to be a powerful, extensible system we should provide advanced users a familiar development experience and toolchain. So baking projects in monolith JSON seems not to be maintainable and free (in the sense of freedom from a particular IDE).

What we need is a formalized language to describe projects, recommended file layout, coding standards, module system and CLI-tools. So that a xoder could work with XOD completely from the terminal (Linux and Mac at least), without ever IDE launch.

Since the project is purely declarative, the language could be quite simple. The proposal can be based on an example:

```
import Servo from 'xod/core/servo';

// Full form
nodetype Pot {
  label = 'Pot';

  input sample {
    label = 'SMPL';
    type = Pulse;
    description = 'blabla';
  }

  output value {
    label = 'VAL';
    type = Number;
  }
}

// Short form
nodetype Button label='Mechanical Switch' {
  // Short form
  output state label='ST' type=Boolean;
}

patch main {
  // Full form
  node pot1 {
    type = Pot;
    x = 100;
    y = 100;
  }

  // Full form
  node servo1 {
    type = Servo;
    x = 300;
    y = 300;
  }

  // Short form
  node servo2 type=Servo x=200 y=300;

  // Domain form
  node servo3 nodetype Servo at 400, 300 label='My supaserva!';

  // Full form
  link {
    outputNode = pot1;
    outputPin = 'val';

    inputNode = servo1;
    inputPin = 'val';

    lazy = true;
  }

  // Domain form
  link from pot1.val to servo1.val lazy;
}
```

Having that a project could be a directory that resembles something linke NodeJS package or Bower package:

```
nodetypes/
  button.xod
  gprs.xod
  http.xod
examples/
  foobar.xod
xod_modules/ <-- project dependencies like node_modules or bower_modules
xod.json <-- package description like package.json or bower.json
```

Such projects could then be distributed and evolved via GitHub, NPM, tarballs and in all conventional ways.

The IDE would be a great `*.xod` editor that loads and saves into files of such projects.

What the concept feels like?

PR #1986, opened 2020-05-15
Updating arduino-cli to 0.10

### LINKED ISSUE #1896: Upgrade arduino-cli to >0.6.0
The new version should fix uploading to ESP32 (https://github.com/arduino/arduino-cli/issues/163)

quinkennedy:
fixes #1896 

`yarn test-func` is still reporting some errors in the `xod-cli` package. Some of these are popping up in a simple `master` repository using the older arduino-cli binary. This leads me to believe that some errors are an issue with my environment.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -41,7 +41,7 @@ defs:
     step-install-arduino-cli-on-linux: &step-install-arduino-cli-on-linux
       name: Install arduino-cli
       command: |
-        curl -L -s --create-dirs -o "$HOME/arduino-cli.tar.bz2" "https://github.com/arduino/arduino-cli/releases/download/0.10.0/arduino-cli_0.10.0_Linux_64bit.tar.gz" \
+        curl -L -s --create-dirs -o "$HOME/arduino-cli.tar.gz" "https://github.com/arduino/arduino-cli/releases/download/0.10.0/arduino-cli_0.10.0_Linux_64bit.tar.gz" \
         && tar xvjf "$HOME/arduino-cli.tar.gz" -C "$HOME" \
         && cp "$HOME/arduino-cli" "./packages/xod-client-electron/arduino-cli" \
         && mkdir "/tmp/arduino-cli" \
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,7 +33,7 @@ defs:
       name: Install arduino-cli
       command: |
         curl -L -s --create-dirs -o "$HOME/arduino-cli.tar.gz" "https://github.com/arduino/arduino-cli/releases/download/0.10.0/arduino-cli_0.10.0_macOS_64bit.tar.gz" \
-        && tar xvjf "$HOME/arduino-cli.tar.gz" -C "$HOME" \
+        && tar xvf "$HOME/arduino-cli.tar.gz" -C "$HOME" \
         && cp "$HOME/arduino-cli" "./packages/xod-client-electron/arduino-cli" \
         && mkdir "/tmp/arduino-cli" \
         && cp "$HOME/arduino-cli" "/tmp/arduino-cli/arduino-cli"
@@ -42,7 +42,7 @@ defs:
       name: Install arduino-cli
       command: |
         curl -L -s --create-dirs -o "$HOME/arduino-cli.tar.gz" "https://github.com/arduino/arduino-cli/releases/download/0.10.0/arduino-cli_0.10.0_Linux_64bit.tar.gz" \
-        && tar xvjf "$HOME/arduino-cli.tar.gz" -C "$HOME" \
+        && tar xvf "$HOME/arduino-cli.tar.gz" -C "$HOME" \
         && cp "$HOME/arduino-cli" "./packages/xod-client-electron/arduino-cli" \
         && mkdir "/tmp/arduino-cli" \
         && cp "$HOME/arduino-cli" "/tmp/arduino-cli/arduino-cli"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -32,9 +32,8 @@ defs:
     step-install-arduino-cli-on-mac: &step-install-arduino-cli-on-mac
       name: Install arduino-cli
       command: |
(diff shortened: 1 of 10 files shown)


quinkennedy:
Finally got past all the `arduino-cli` initialization errors. `xod-cli` functional tests seem to be very temperamental, and I can't figure out how to properly debug them.

nkrkv:
Hello and thank you for the contribution! Unfortunately, the update does not work on my machine (Linux). The tests are passing, though. Other devs told they have test failing… they are on macOS. So, very careful testing is required.

Would you share your progress? Are you able to upload programs from within XOD using `arduino-cli`? Are you on macOS, right? What board model do you have?

quinkennedy:
I am unable to select a USB port when I run it on OSX. No USB ports are enumerated in the "upload" dialog. This affects a build on the `master` branch as well as on this feature branch. I'm just testing using `yarn start:electron`. 

Attempting to build a distributable errors out trying to find Mac/Memory.pm which might be fixed in a [newer electron-builder](https://github.com/electron-userland/electron-builder/issues/3990)

brusherru:
I consider that this is not a problem of an old electron builder, because it used only in bundling the production version.
I had the same problems a few times in the past, it might be the same problem, which provided on the edge between NodeJS and Electron versions and `serialport` package.

Read carefully the output of `yarn` and `yarn build` commands, `serialport` package may show warnings. If there is no conflict with NodeJS version rebuilding from the source may help:
`npm rebuild --build-from-source` or try this one: https://serialport.io/docs/guide-installation#electron

quinkennedy:
I was getting a `gyp` error `gyp: No Xcode or CLT version detected!` which didn't make sense since Xcode and CLT _were_ installed. Turns out [reinstalling after upgrading to Catalina](https://medium.com/flawless-app-stories/gyp-no-xcode-or-clt-version-detected-macos-catalina-anansewaa-38b536389e8d) solves that. I will be able to test later.

quinkennedy:
I ran `yarn start:electron` again after fixing my command line tools. I set `XOD_ARDUINO_CLI` to the v0.10 executable. Trying to upload _101-upload_ from the _welcome-to-xod_ tutorials gets stuck at compiling 😞
Screen Shot 2020-05-27 at 11 16 20 AM

nkrkv:
> tutorials gets stuck at compiling

Same here (Linux). At least we are at the same page now.

quinkennedy:
next question: it looks like the compiler is getting stuck when checking for a `.xodworkspace` file. How does this workspace file get created, and why isn't it in my packages/xod-cli/bundle/workspace/ directory? I see it in the root-level workspace/

quinkennedy:
I suspect there is a difference between how the old and new arduino CLI saves sketches. i assume before it would use the path provided as a filename, but now it uses it as a directory, and creates a file inside that directory. i'm still trying to track down the command to test...

quinkennedy:
I'm running into an issue when the code copies arduino libraries into the temp directory. the directories get copied, but no files do.
Screen Shot 2020-06-05 at 11 54 46 PM
The error is 
```
🙈 Error
📓 ENOENT: no such file or directory, open '../../../../../var/folders/7w/hv37xvjx34jfntk0tfp96zwmzk3ccz/T/xod_temp_sketchbooknxhgsE/libraries/README.md'
```
I'll try using [the copy command in `fs-extra`](https://github.com/jprichardson/node-fs-extra/blob/HEAD/docs/copy.md) instead.

quinkennedy:
I forgot to mention, I'm currently trying to track down these bugs by running `packages/xod-cli/bin/run compile --output=test_output --board=esp8266:esp8266:wifi_slot --workspace=test_workspace --debug packages/xod-cli/bundle/workspace/blink @/main`

quinkennedy:
`xodc boards TTY list boards, creates workspace, stderr is empty, zero exit code` seems to throw an error on OSX unless I run the `arduino-cli board listall` command manually first... not totally sure how to remedy that for the unit tests.

I spent a while getting `xodc compile TTY compile patch, stdout is empty, stderr with messages, zero exit code:` to working on my own computer, which seemed partially to do with arduino-cli creating the sketch in the wrong local directory. I figured providing an absolute directory would remove ambiguity and fix the issue, but it has cropped up on both linux and osx CI tests 😡

quinkennedy:
passes the automated tests 🙌

unfortunately, doesn't fix the issue with failing builds for my newer boards. I originally followed [#163](https://github.com/arduino/arduino-cli/issues/163) to #1896 which inspired me to tackle this pull request. I'm running via `yarn start:electron` after setting the `XOD_ARDUINO_CLI` environment variable, is there a way to confirm it is using the v0.10 arduinoCLI ?

patch: 101-upload in welcome-to-xod project
board: Arduino Uno
result: successful compile, upload, & run from within XOD

patch: 101-upload in welcome-to-xod project
board: [Adafruit ItsyBitsy nRF52840 Express](https://www.adafruit.com/product/4481)
bootloader: 0.2.11 SoftDevice s140 6.1.1
result: successful compile, upload, & run _when exporting code via Arduino IDE_
xod error: [full output from compiler tab](https://pastebin.com/zBN8sbdu)
```
Compilation failed
Command maxBuffer size exceeded
The generated C++ code contains errors. It can be due to a bad node implementation or if your board is not compatible with XOD runtime code. The original compiler error message is above. Fix C++ errors to continue. If you believe it is a bug, report the problem to XOD developers.
```

**to be clear** I do not think this is a regression, I would still consider this valuable to merge into master at this point.

evgenykochetkov:
The `Command maxBuffer size exceeded` error happens because the default value for the respective option in `promisify-child-process` is quite small: https://github.com/jcoreio/promisify-child-process/blob/master/src/index.js#L70 But that's a separate issue.

> I do not think this is a regression. I would still consider this valuable to merge into master at this point.

I absolutely agree! Excellent work ❤️

nkrkv:
So, ready to merge. @quinkennedy, thank you a lot. Would you rebase the branch onto the current `master`? See https://github.com/xodio/xod/blob/master/CONTRIBUTING.md#rebasing-on-master

evgenykochetkov:
Rebased and added a couple tweaks to address `maxBuffer size exceeded` error and unnecessarily verbose compiler output.

PR #2000, opened 2020-06-15
Add xod/waves library

evgenykochetkov:
and deprecate old *-wave nodes from xod/core:
`xod/core/saw-wave`
`xod/core/saw-wave-map`
`xod/core/sine-wave`
`xod/core/sine-wave-map`
`xod/core/square-wave`
`xod/core/tri-wave`
`xod/core/tri-wave-map`

### DIFF
--- a/workspace/__lib__/xod/waves/saw-wave/patch.xodp
+++ b/workspace/__lib__/xod/waves/saw-wave/patch.xodp
@@ -1,0 +1,138 @@
+{
+  "description": "Generates sawtooth wave signal",
+  "links": [
+    {
+      "id": "B1FgjpVpI",
+      "input": {
+        "nodeId": "HkEkipVp8",
+        "pinKey": "BJ4ixaQjOf"
+      },
+      "output": {
+        "nodeId": "H1m-sPgGdM",
+        "pinKey": "__out__"
+      }
+    },
+    {
+      "id": "S1PgsT4p8",
+      "input": {
+        "nodeId": "HkEkipVp8",
+        "pinKey": "Syjea7idG"
+      },
+      "output": {
+        "nodeId": "rkSWsDxz_G",
+        "pinKey": "__out__"
+      }
+    },
+    {
+      "id": "SJLls6ETU",
+      "input": {
+        "nodeId": "HkEkipVp8",
+        "pinKey": "HJHig6QidG"
+      },
+      "output": {
+        "nodeId": "B1G-jwgz_M",
+        "pinKey": "__out__"
+      }
+    },
+    {
+      "id": "SJmxopNa8",
+      "input": {
+        "nodeId": "HkEkipVp8",
+        "pinKey": "rk-jxTQsuz"
+      },
+      "output": {
+        "nodeId": "H1gboDlzOM",
+        "pinKey": "__out__"
+      }
+    },
+    {
+      "id": "r16gipVa8",
+      "input": {
+        "nodeId": "SkZowlzOM",
+        "pinKey": "__in__"
+      },
+      "output": {
+        "nodeId": "HkEkipVp8",
+        "pinKey": "r1ljgpXjOM"
+      }
+    }
+  ],
+  "nodes": [
+    {
+      "boundLiterals": {
+        "__out__": "True"
+      },
+      "description": "Enabled or not. If set to `false` pulses on `UPD` do not change the output value. Effectively that means the generator is paused. Set to `true` again to continue generation.\n",
+      "id": "B1G-jwgz_M",
+      "label": "EN",
+      "position": {
+        "units": "slots",
+        "x": 2,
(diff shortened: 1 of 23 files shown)


workspace/__lib__/xod/waves/saw-wave/patch.xodp
```diff
@@ -0,0 +1,160 @@
+{
+  "description": "Generates triangle wave signal",
```
gabbapeople, line 2: "description": "Generates sawtooth wave signal",

workspace/__lib__/xod/waves/saw-wave/patch.xodp
```diff
@@ -0,0 +1,160 @@
...
+        "pinKey": "HJagfP258"
+      }
+    }
+  ],
+  "nodes": [
+    {
+      "boundLiterals": {
+        "__out__": "True"
+      },
+      "description": "Enabled or not. If set to `false` pulses on `UPD` do not change the output value. Effictively that means the generator is paused. Set to `true` again to continue generation.\n",
```
gabbapeople, line 76: Effictively -> Effectively

workspace/__lib__/xod/waves/saw-wave/patch.xodp
```diff
@@ -0,0 +1,160 @@
...
+      "label": "OUT",
+      "position": {
+        "units": "slots",
+        "x": 0,
+        "y": 2
+      },
+      "type": "xod/patch-nodes/output-number"
+    },
+    {
+      "description": "The number of a funcion cycle. Starts from 0.",
```
gabbapeople, line 138: funcion -> function

gabbapeople:
I still suggest that the zero value of the period should not cause an error. It is also confusing to use the future tense in the pin descriptions of some nodes. But, in eneral, everything is ok.

evgenykochetkov:
> I still suggest that the zero value of the period should not cause an error.

How do you think the node should behave in this case?

gabbapeople:
> > I still suggest that the zero value of the period should not cause an error.
> 
> How do you think the node should behave in this case?

Depending on the node. But, theoretically, if the distance between the extreme vertices of the graph tends to 0, the graph turns into a straight line, output = 1, or true.

workspace/__lib__/xod/waves/variable-triangle-wave-map/patch.cpp
```diff
@@ -0,0 +1,17 @@
+
+struct State {
+};
+
+{{ GENERATED_CODE }}
+
+void evaluate(Context ctx) {
+    auto fractionalPart = fmod(getValue<input_X>(ctx), 1);
+    auto peakPosition = constrain(getValue<input_PP>(ctx), 0.0, 1.0);
```
brusherru, line 9: I got an error in simulation:
```
error: use of undeclared identifier 'constrain'
    auto peakPosition = constrain(getValue<input_PP>(ctx), 0.0, 1.0);
                        ^
1 error generated.
```

So you probably have to use conditions or make a patch to our wasm package with the implementation of this function (and pray to God that every Arduino package developers add it into `core.h` 😄 ).

evgenykochetkov: > and pray to God that every Arduino package developers add it into core.h 😄

I'm pretty sure they do. It's not like it's some obscure undocumented function: https://www.arduino.cc/reference/en/#functions

workspace/__lib__/xod/waves/square-wave/patch.cpp
```diff
@@ -0,0 +1,53 @@
...
+
+    if (enabled && !state->wasEnabled) {
+        // just enabled/resumed
+        state->timeToSwitch = (period * duty) * 1000.0;
+        setTimeout(ctx, state->timeToSwitch);
+        state->nextSwitchTime = t + state->timeToSwitch;
+        emitValue<output_OUT>(ctx, true);
+    } else if (!enabled && state->wasEnabled) {
+        // just paused
+        // TODO: we can get rid of storing nextSwitchTime if API would
```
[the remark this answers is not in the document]
evgenykochetkov: This particular implementation is just a copy of `xod/core/square-wave`.
I'm not sure about growing the API for a single use-case. Do you have some examples where it could also be useful?

brusherru:
BTW, what about replacing usage of the deprecated nodes for the new ones in the standard library (`xod/core/flip-n-times` and `xod-dev/ws2812/example/example-pattern-police`)? 🤔

workspace/__lib__/xod/waves/square-wave/patch.cpp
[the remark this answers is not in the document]
nkrkv: Suggest leaving it as is for now to limit the PR scope

PR #2004, opened 2020-07-04
Create abstract nodes for interaction with internet, unify specializations and make it resolves correctly

brusherru:
There is no issue.
But now it provides a possibility to make a faster implementation of the `xod-cloud/feeds/cloud-store` node.

First commit is about unifying interfaces of `open-tcp`, `send`, `receive`, `close` nodes in `xod-dev/esp8266`, `xod-dev/esp8266-mcu`, `xod-dev/w5500`, and `xod/debug` with saving the previous PIN ids.
BUT `xod-dev/esp8266-mcu` has some breaking changes, so if someone used listed above nodes in their libraries or projects (instead of using `http-request` specialization) — it might break the project or library 🤔 

The second one, about resolving abstract nodes, such as `xod/net/open-tcp` to specializations without the type suffix for the types, which is defined in the same library. If it requires more than one type, each of custom types should be defined inside the same library or it still requires the type suffix.
For example,
`some/lib/specialization` with two terminals for `some/lib/custom-type-1` and `some/lib/custom-type-2` does not require a suffix.
`another/lib/specialization` with two terminals for `another/lib/custom-type` and `some/lib/custom-type` requires a suffix.


packages/xod-project/test/patch.spec.js
```diff
@@ -2036,34 +2036,64 @@ describe('Patch', () => {
         );
       });
 
-      it('should check patch basename', () => {
+      it('should check for specialization suffix if specialization has a any of custom types from other library', () => {
```
nkrkv, line 2039:
```suggestion
      it('should check suffix in parens if specialization has custom types from another library', () => {
```

PR #2006, opened 2020-07-08
Fix forward declaration of HardwareSerial for megaAVR architecture

brusherru:
There is no issue.
The problem was reported on the forum: https://forum.xod.io/t/nano-every-upload-issue/4198

### DIFF
--- a/packages/xod-arduino/platform/uart.h
+++ b/packages/xod-arduino/platform/uart.h
@@ -6,7 +6,12 @@
  *
  =============================================================================*/
 
+
+#if ARDUINO_API_VERSION >= 10001
+class arduino::HardwareSerial;
+#else
 class HardwareSerial;
+#endif
 class SoftwareSerial;
 
 namespace xod {
--- a/workspace/big-patch/__fixtures__/arduino.cpp
+++ b/workspace/big-patch/__fixtures__/arduino.cpp
@@ -455,7 +455,12 @@ void* operator new(size_t, T* ptr) noexcept {
  *
  =============================================================================*/
 
+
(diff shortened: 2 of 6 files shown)


nkrkv:
Take a look at the core used by Nano Every (https://github.com/arduino/ArduinoCore-megaavr). Namely, to the file https://github.com/arduino/ArduinoCore-megaavr/blob/master/cores/arduino/UART.h. See:

```cpp
#include "api/HardwareSerial.h"

using namespace arduino;
```

What the `api/`. See README:

> Copy or symlink the api folder from the ArduinoCore-API (https://github.com/arduino/ArduinoCore-API) repo.

Go there. Read:

> This repository hosts the hardware independent layer of Arduino core.

> All Arduino official cores are being ported to the new structure so they take advantage of this single repo.

> Including this repo in your existing Arduino core will allow the language to grow and include new features. For backwards compatibility, every revision of this repo will increase `ARDUINO_API_VERSION` define.

> Some cores have been ported to the new structure, for example: megaAVR, nRF52-mbedos, classic AVR, SAMD

Whoa! Some classic cores *already* use the new API wich has `arduino` namespace. So, merging this PR as is only fixes a little portion of the problem. It will return again and again as users update their packages.

So the proper way of fixing the issue would be `using namespace arduino;` at the global level. However, I afraid it’s too wide and can cause problems which are not apparent immediately. For now, a safer workaround would be looking at `ARDUINO_API_VERSION` macro definition, not `ARDUINO_ARCH_MEGAAVR`.

brusherru:
@nkrkv done!

nkrkv:
The fixtures do not match `uart.h`

PR #2007, opened 2020-07-15
Introduce constant types

evgenykochetkov:
Values of constant types are known at compile time, and can't be changed during the runtime.
This allows for some compile-time checks and optimisations.

The value of a constant inputs are accessible as `constant_input_PINNAME`, as well as through regular `getValue` function.

The value of a constant output has to be declared like `static const TypeOfPINNAME constant_input_PINNAME = value;`.

Types of inputs and outputs are accessible as `TypeOfPINNAME` instead of `ValueType<output_PINNAME>::T`, and are available before declaring `State`, allowing some truly generic implementations.

Custom types that have constant parameters also some receive special treatment. Types of such outputs have to be declared manually. For example, `typedef TypeOfIN TypeOfOUT;` or `typedef some__node<new_constant_value>::Type TypeOfOUT;`. The `typedef TypeOfIN TypeOfOUT;` case is automated if the output pin is named like input pin, but with an `'` at the end, like `DEV` and `DEV'`.


nkrkv:
So, from what I see, almost all the code of a node implementation is implicitly enclosed in a struct:

```cpp
namespace xod {
// ...
struct xod__common_hardware__text_lcd_16x2 {
  // ...
  
  void printLine(LiquidCrystal* lcd, uint8_t lineIndex, XString str) { ... }
  void evaluate(Context ctx) { ... }
};
}
```

This may lead to silly surprises: worse performance, no way to get a pointer to a seemingly zero-argument function, etc. I’m sure, that putting something inside a `struct`/`class` should be visible and explicit to a xoder.

So, how to acheive this? Some experiments are required but here’s what I think in rough detail. First, let’s see what a node implementation looks like now:

```cpp

struct State { int xMemo; };
using Type = uint32_t;
/* Bla bla I’m a top comment */

// Handlebars hack to render generated boilerplate
// *and* to split the source in two parts
{{ GENERATED_CODE }}

/* Bla bla I’m a bottom comment */
void evaluate(Context ctx) {
  ...
}

```

It becomes:

```cpp
namespace xod {
template < /* constant inputs */ >
struct some__fqdn__ns {
    struct State { int xMemo; };
    using Type = uint32_t;
    /* Bla bla I’m a top comment */

    // typeof_xxx defs (refers Type for custom types)
    // getState defs (refers State)
    // emitValue defs (refers Type for custom types)
    // getValue defs
    // setTimeout defs
    // etc etc

    /* Bla bla I’m a bottom comment */
    void evaluate(Context ctx) {
      ...
    }
}
};
```

To recall, besides the comile-time template constants, the idea to wrap some parts of the node implementation with `struct` is to get rid of ugly non-C++ `{{ GENERATED_CODE }}` and to simplify referring the state. That is, get rid of `State` as of enclosing structure for the persistent state.

How the current solution might be adjusted to meet the requirements? The node implementation might look like:

```cpp
// The custom type definition still uses the top level
using Type = uint32_t;
void iWantToStayGlobalFunc() { ... }
/* Bla bla I’m a top comment */

// Look ma, no GENERATED code

/* Bla bla I’m a bottom comment */
defnode {
    int xMemo;
    
    void iWantToBeAMethodWithStateAccess() {
      ++xMemo;
    }
    
    void evaluate(Context ctx) {
      ...
    }
};
```

This might be rendered to:

```cpp
namespace xod {
namespace some__fqdn__ns {
using Type = uint32_t;
void iWantToStayGlobalFunc() { ... }
/* Bla bla I’m a top comment */

template < /* constant inputs */ >
struct Node {
  // typeof_xxx defs (refers Type for custom types)
  // getState defs are not required for the new-style node defs
  // emitValue defs (refers Type for custom types)
  // getValue defs
  // setTimeout defs
  // etc etc
  
  int xMemo;

  void iWantToBeAMethodWithStateAccess() {
    ++xMemo;
  }
    
  void evaluate(Context ctx) {
    ...
  }
}
}
}
```

So, basically, I suggest introducing the `defnode` keyword which is get replaced by the transpiler with stuff that now’s called `GENERATED_CODE`. The substitution might be even performed by C++ compiler if we `#define defnode some long long long long multiline generated code` before each node context and `#undef defnode` after, however it could mess compiler error messages and editor highlight. I think it would be better to ensure that `defnode` is not within comment and perform a direct substitution on the XOD side.

The code outside the `defnode` stays global, only namespaced to `xod` / `some__fqdn__ns`. We can go even further and clear the global scope of any implicit namespaces asking a xoder to use a namespace to avoid conflicts:

```cpp
#include <LiquidCrystal.h>

ISR(...) {
  // Interupt handler must by global
}

nodespace {
  using Type = uint32_t;
  void iWantToStayGlobalFunc() { ... }
}

defnode {
  int xMemo;
    
  void iWantToBeAMethodWithStateAccess() {
    ++xMemo;
  }
    
  void evaluate(Context ctx) {
    ...
  }
};
```

Might become:

```cpp
#include <LiquidCrystal.h>

ISR(...) {
  // Interupt handler must by global
}

namespace some__fqdn_ns {
  using Type = uint32_t;
  void iWantToStayGlobalFunc() { ... }
}

namespace some__fqdn_ns {
template < /* constant inputs */ >
struct Node {
  // typeof_xxx defs (refers Type for custom types)
  // getState defs are not required for the new-style node defs
  // emitValue defs (refers Type for custom types)
  // getValue defs
  // setTimeout defs
  // etc etc
  
  int xMemo;

  void iWantToBeAMethodWithStateAccess() {
    ++xMemo;
  }
    
  void evaluate(Context ctx) {
    ...
  }
};
}
```

What do you think?

workspace/__lib__/xod-dev/dht/dht11-device/patch.cpp
```diff
@@ -1,6 +1,6 @@
 
 struct State {
-    uint8_t port = 0;
+    static const uint8_t port = constant_input_PORT;
 };
```
nkrkv, line 4: Can this canonical implementation be reworked to the new style?

Also, the `port` should be `static constexpr`, not `static const`

workspace/__lib__/xod-dev/text-lcd/print-at(text-lcd-i2c-device)/patch.cpp
```diff
@@ -1,40 +1,37 @@
...
-
-{{ GENERATED_CODE }}
-
-void printAt(LiquidCrystal_I2C* lcd, uint8_t rowIndex, uint8_t posIndex, uint8_t len, XString str) {
-    lcd->setCursor(posIndex, rowIndex);
-    uint8_t whitespace = len;
-    for (auto it = str.iterate(); it && whitespace > 0; ++it, --whitespace)
-        lcd->write(*it);
+node {
+    void printAt(LiquidCrystal_I2C* lcd, uint8_t rowIndex, uint8_t posIndex, uint8_t len, XString str) {
```
nkrkv, line 3:
```suggestion
    static void printAt(LiquidCrystal_I2C* lcd, uint8_t rowIndex, uint8_t posIndex, uint8_t len, XString str) {
```

Because no dependency on `this` here. Saves a few call instructions.

workspace/__lib__/xod-dev/text-lcd/print-at(text-lcd-parallel-device)/patch.cpp
```diff
@@ -1,40 +1,36 @@
+node {
+    void printAt(LiquidCrystal* lcd, uint8_t rowIndex, uint8_t posIndex, uint8_t len, XString str) {
```
nkrkv, line 2:
```suggestion
    static void printAt(LiquidCrystal* lcd, uint8_t rowIndex, uint8_t posIndex, uint8_t len, XString str) {
```

brusherru (approved):
LGTM

nkrkv (approved):
:+1:

PR #2024, opened 2020-08-18
Add missing dependency for xod-dev/pn532-nfc

evgenykochetkov:
Reported by @S-VIN.

`Adafruit-PN532` library depends on `Adafruit_BusIO` and won't compile without it.


brusherru (approved):
LGTM

But probably we have to think about feature how to point on a specific version to avoid such bugs 🤔

nkrkv (approved):

PR #2025, opened 2020-08-25
Fix flatten: get rid of stale output custom type terminals

brusherru:
There is no issue.
You can reproduce the bug by transpiling ("Show Code for Arduino") the `xod-dev/esp8266-mcu/example-get-data-from-dweet` patch.

Or you can create a minimal patch to reproduce the bug. Note, that there is a nested patch with a branch: one link directly to the output, another one — to the `format-color` node.

```
+-----------+
| color-hsl |
|           |
+-----O-----+
      |
      |
+-----O----------------+
|     |                |
|     +-------|        |
|     |     +-O------+ |
|     |     |        | |
|     |     | format | |
|     |     |        | |
|     |     +-O------+ |
|     |       |        |
|     |       |        |
+-----O-------O--------+

    CLR      STR

```

### DIFF
--- a/packages/xod-project/src/flatten.js
+++ b/packages/xod-project/src/flatten.js
@@ -654,8 +654,10 @@ const traverseUpAndCollectTerminalChain = R.curry(
         R.compose(R.equals(nextNodeId), Link.getLinkInputNodeId),
         links
       );
-      // If the uppermost node has no input link — return collected data
-      if (!nextLink) return collected;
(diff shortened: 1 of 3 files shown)


nkrkv (approved):

evgenykochetkov (approved):

evgenykochetkov (approved):
👌🏻

PR #2026, opened 2020-08-28
Use upstream ESP8266 package index instead of locked v2.4.3

brusherru:
There is no issue.
But we used this version by default because the latest release at that time had some critical bug.
Now the latest version works fine.


evgenykochetkov (approved):
Finally! 👍🏻

nkrkv (approved):
:+1:

PR #2027, opened 2020-08-31
Fix follow log in Debugger pane

brusherru:
There is no issue, but there were some small bugs:
- if messages added to the log, but the log is hidden, then it might be opened not on the bottom and "Follow log" will be shown up from the beginning
- after clicking "Follow log" it sometimes does not follow it
- Clearing log did not start following the log


evgenykochetkov (approved):
👍🏻

PR #2028, opened 2020-09-01
Update standard libraries to fit new standards of code

brusherru:


nkrkv (approved):
Net -1300 LOC. Worth doing :grin:

PR #2034, opened 2020-09-13
Fix typo in implementation of xod/common-hardware/ds18b20-thermometer

evgenykochetkov:
The issue was [reported on our forum](https://forum.xod.io/t/xod-35-0-soft-uart-fail/4413/4)


nkrkv (commented):
There will be some similar fixes. Would you target 0.35.x branch instead?

brusherru (approved):

PR #2035, opened 2020-09-14
Make sure implementation keywords are always detected correctly

evgenykochetkov:
Makes identifiers that contain keywords `node`, `nodespace` and `meta` always parse as identifiers, not keywords.


brusherru (approved):
👍

nkrkv (approved):

PR #2038, opened 2020-09-15
Support esp32 boards

brusherru:
This PR includes:
- fix of adding new URLs into `extra.txt`: updated arduino-cli returns "ENOENT" code instead of `6` when JSON file is not found and our `catch` does not work as expected. This bug does not break a loading new file, but it shows an error message, and list of boards appears only after clicking "Update" button or opening of the upload popup next time.
- Bundle https://github.com/madhephaestus/ESP32Servo library into desktop IDE to support PWM and Servos without requiring libraries from the internet
- Tweak `xod-dev/servo/servo-device` and `xod/gpio/pwm-write` to use the bundled library and work fine on ESP32

### DIFF
--- a/packages/xod-deploy-bin/src/arduinoCli.js
+++ b/packages/xod-deploy-bin/src/arduinoCli.js
@@ -433,13 +433,15 @@ export const listBoards = async (wsPath, cli) => {
       available: res[1],
     }))
     .catch(async err => {
-      if (R.propEq('code', 6, err)) {
-        // Catch error produced by arduino-cli, but actually it's not an error:
+      if (R.propEq('code', 'ENOENT', err)) {
         // When User added a new URL into `extra.txt` file it causes that
         // arduino-cli tries to read new JSON but it's not existing yet
         // so it fails with error "no such file or directory"
         // To avoid this and make a good UX, we'll force call `updateIndexes`
-        return updateIndexesInternal(wsPath, cli).then(() => listBoards(cli));
+        // and then run `listBoards` again.
+        return updateIndexesInternal(wsPath, cli).then(() =>
+          listBoards(wsPath, cli)
+        );
       }
 
       throw createError('UPDATE_INDEXES_ERROR_BROKEN_FILE', {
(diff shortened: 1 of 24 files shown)


packages/xod-deploy-bin/src/arduinoCli.js
```diff
@@ -433,13 +433,15 @@ export const listBoards = async (wsPath, cli) => {
...
     .catch(async err => {
-      if (R.propEq('code', 6, err)) {
+      if (R.propEq('code', 'ENOENT', err)) {
         // Catch error produced by arduino-cli, but actually it's not an error:
         // When User added a new URL into `extra.txt` file it causes that
         // arduino-cli tries to read new JSON but it's not existing yet
         // so it fails with error "no such file or directory"
         // To avoid this and make a good UX, we'll force call `updateIndexes`
-        return updateIndexesInternal(wsPath, cli).then(() => listBoards(cli));
+        return updateIndexesInternal(wsPath, cli).then(() =>
```
nkrkv, line 442: The comment above is stale now. Needs update.

nkrkv (changes requested):

packages/xod-deploy-bin/src/arduinoCli.js
brusherru: 🤔 
It still actual, but I'll edit it a bit.

workspace/__ardulib__/ESP32Servo/doxy.doxyfile
```diff
@@ -0,0 +1,2428 @@
+# Doxyfile 1.8.11
```
evgenykochetkov, line 1: Do we really need to include an 2,5k line .doxyfile?

brusherru: 👌 
I'll remove it.

evgenykochetkov (approved):
LGTM 👍🏻

nkrkv (approved):
:+1:

PR #2040, opened 2020-09-17
Fix hotkeys in Inspector Pin widgets

brusherru:
There is no issue.
Some time ago, when we refactor widgets, we'd lost a method `updateValue` of `Widget` class.

Now it's back and these hotkeys will work again:

- Escape button to cancel changes
- Arrow keys (optionally with shift key) to change numbers


evgenykochetkov (approved):
👍🏻

PR #2041, opened 2020-09-17
Fix tweak-bytes: make possible to use char literals in live session

### LINKED ISSUE #1864: Can't bind the char into `tweak-byte`
Binding char literals to the `tweak-byte` node in live session does not work.
It reproduces both in simulation mode and in the debug mode.

### Steps to reproduce
1. Place a `tweak-byte`, `watch` nodes and link them
2. Bind the `'a'` (as a char literal) to the `tweak-byte`
3. Start the simulation
4. You'll see `61h` in the `watch` node
5. Try to bind the `'b'` char to the `tweak-byte`

### Expected behavior
`watch` node value changed to `62h`

### Actual behavior
`watch` node value changed to `00h`

brusherru:
It fixes #1864

### DIFF
--- a/packages/xod-arduino/src/formatTweakMessage.js
+++ b/packages/xod-arduino/src/formatTweakMessage.js
@@ -7,6 +7,24 @@ import { def } from './types';
 
 import { byteLiteralToDecimal } from './templates';
 
+// Convert char literals to decimal byte literals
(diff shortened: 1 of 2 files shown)


evgenykochetkov (approved):
LGTM!

PR #2042, opened 2020-09-17
Copy values bound to `tweak-string`s directly into buffers

evgenykochetkov:
It fixes a problem with `tweak-string-*`s introduced in v0.35.0 and generally optimises memory usage.


nkrkv (approved):
:+1:

packages/xod-arduino/src/templates.js
```diff
@@ -97,6 +97,8 @@ const cppType = def(
   )
 );
 
+const escapeCppString = str => R.replace(/"/g, '\\"', unquote(str));
```
brusherru, line 100: It's time to make escaping safer.
Here is the case that fails in compile-time, but the same tweak value then works fine in live session: `"mwa\\ha"ha\"` (first and last quotes is a part of literal)

brusherru (approved):
Okay. Escaping chars we can fix further in separate PR :)
LGTM

PR #2045, opened 2020-09-21
Add information about constants in C++ Cheatsheet (sidebar)

brusherru:
There is no issue.

Previews:
image
image

### DIFF
--- a/packages/xod-client/src/editor/components/CppPatchDocs.jsx
+++ b/packages/xod-client/src/editor/components/CppPatchDocs.jsx
@@ -7,6 +7,29 @@ import Icon from 'react-fa';
 
 import { getUtmSiteUrl } from '../../utils/urls';
 
+const getPinLabel = R.compose(cppEscape, XP.getPinLabel);
+
+const formatPinLabel = pin =>
+  [
+    XP.isConstantType(pin.type) ? 'constant_' : '',
+    XP.getPinDirection(pin),
+    '_',
+    getPinLabel(pin),
+  ].join('');
+
+const formatConstantOutputNotice = pin =>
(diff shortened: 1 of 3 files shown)


packages/xod-client/src/editor/components/CppPatchDocs.jsx
```diff
@@ -7,6 +7,29 @@ import Icon from 'react-fa';
...
+
+const formatConstantOutputNotice = pin =>
+  XP.isConstantType(pin.type) ? (
+    <span className="constant-notice">
+      <Icon name="warning" /> Must be defined as
+      <code>
+        <span className="keyword">static constexpr </span>
+        <span className="type">typeof_{getPinLabel(pin)} </span>
+        <span className="variable">{formatPinLabel(pin)} </span>
+        =
```
nkrkv, line 28: This explanation looks silly to me. An attempt to teach a particular nuance in very limited space. I think this should be omitted here and explained in other places: in the docs and (at some moment) in the auto-generated C++ template.

brusherru: Idea about the auto-generated C++ template sounds cool and I thought that's an awesome way. But now I think it's not so easy to implement. For example, what if C++ implementation already "touched", but only after that User added `output-port`?

About detailed explanation, in the docs, I totally agree. However, this is a rare case, I think when we create some nodes with the constant output. So I suppose that User will not use such snippet often and it might bring some pain to remember how to do it or go to the documentation page, instead of just copy-pasting the snippet.

So I propose to leave it as is.
At least for some time. Or until we bring some new snippets and "click-to-place" mechanism :)

evgenykochetkov (approved):

PR #2050, opened 2020-10-01
Fix `xod-dev/esp8266-mcu/lan-ip` implementation

evgenykochetkov:
The issue was reported on [our forum](https://forum.xod.io/t/lan-ip-is-not-compiling/4470)


brusherru (approved):
👍

PR #2053, opened 2020-10-05
fix(stdlib): make xod graphics bitmaps compile on non-avr boards

gabbapeople:
This bug was found here: https://forum.xod.io/t/graphics-arduino-due-st7735/4411


nkrkv (approved):
:+1:

evgenykochetkov (approved):
👍🏻

PR #2056, opened 2020-10-20
Introduce setImmediate API

evgenykochetkov:
Before, we used `setTimeout(0, ctx)` to schedule evaluation of a node as soon as possible.
But it would only mark node as dirty on the next millisecond and generate some unnecessary code.

`setImmediate` is a new API for this particular case that works faster, generates less code and occupies less RAM.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,12 +17,23 @@ defs:
       paths:
         - node_modules
 
+    cache-arduino15: &cache-arduino15
+      # update cache only if package_index.json changed
+      key: arduino15-v1-{{ arch }}-{{ .Branch }}-{{ checksum "~/.arduino15/package_index.json" }}
+      paths:
+        - ~/.arduino15/staging # only archives
+
     restore-node_modules: &restore-node_modules
       keys:
         - node_modules-{{ arch }}-{{ checksum ".circleci/cache-version" }}-{{ checksum "yarn.lock" }}
         # if cache for exact version of `yarn.lock` is not present then
         # load any most recent one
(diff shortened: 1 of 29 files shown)


.circleci/config.yml
```diff
@@ -138,10 +138,18 @@ jobs:
...
       XOD_OUTPUT: /tmp/tabtests-cpp
+      XOD_ARDUINO_CLI: /tmp/arduino-cli/arduino-cli
     steps:
       - checkout
       - restore_cache: *restore-node_modules
       - run: *step-install
+      - run: *step-install-arduino-cli-on-linux
+      - run:
+          name: Install arduino:avr platform for arduino-cli
+          command: $XOD_ARDUINO_CLI core install arduino:avr
```
knopki, line 149: Worth caching? In order to create less load on the arduino mirrors.

nkrkv (approved):
:heart:

brusherru (approved):

.circleci/config.yml
knopki: @evgenykochetkov CircleCI cache is immutable, so it can't be updated with the same key.

Proposed solution (not tested):
```yaml
- restore_cache:
    name: Restore arduino-cli cache
    keys:
      - arduino15-v1-{{ arch }}-{{ .Branch }}-
      - arduino15-v1-{{ arch }}-
# Install arduino:avr platform for arduino-cli
# Test AVR Size
- save_cache:
    name: Save arduino-cli cache
    # update cache only if package_index.json changed
    key: arduino15-v1-{{ arch }}-{{ .Branch }}-{{ checksum "~/.arduino15/package_index.json" }}
    paths:
      - ~/.arduino15/staging # only archives
```

PR #2068, opened 2020-11-10
Create nodes with hotkeys while linking

### LINKED ISSUE #1770: As a xoder I want to create essential nodes with hotkeys while linking so that I can patch faster
Follows a forum [thread](https://forum.xod.io/t/easy-access-to-pins-and-constants/2163). The following nodes arguably required more often than others:

- Terminals
- Watches and tweaks
- Constants

We can assign a hotkey which will automatically place a node under the cursor when linking has been started.

### User story

- I start linking, the rubber line is here
- I press the T key and a terminal node of the proper type with the label stollen appears under the cursor
- Likewise, C creates a constant, W creates a t**w**eak or watch

### Acceptance criteria

- [x] When linking, `T` creates a terminal under the cursor of the same type as the link
- [x] The terminal label is copied (non-normalized)
- [x] When linking an input pin, `C` creates a constant node of the proper type (or a custom type **constructor** node)
- [x] The bound value is copied to the constant node
- [x] When linking an input pin of type boolean, byte, number, or pulse, `W` creates a tweak node of the corresponding type
- [x] When linking an input pin of type string, `W` creates a `tweak-string-16` node
- [x] The bound value is copied to the tweak node
- [x] When linking an output pin, `W` creates a `watch` node

brusherru:
It closes #1770 

Each commit refactors a code to reduce code duplication. So probably easier to check out the whole changelog.


packages/xod-project/src/constants.js
```diff
@@ -278,3 +280,10 @@ export const MANAGED_ATTACHMENT_TEMPLATES = {
   [NOT_IMPLEMENTED_IN_XOD_PATH]: IMPL_TEMPLATE,
   [TABTEST_MARKER_PATH]: TABTEST_TEMPLATE,
 };
+
+export const NODE_KIND = {
```
evgenykochetkov, line 284: I don't think this needs to live in `xod-project`

evgenykochetkov (changes requested):
Trying to create a tweak from a generic pin results in an incorrect patch.
Trying to create a constant node from a generic pin crashes the IDE.

packages/xod-project/src/patchPathUtils.js
```diff
@@ -261,6 +262,10 @@ const constantNodeRegExp = new RegExp(
 // :: PatchPath -> Boolean
 export const isConstantNodeType = R.test(constantNodeRegExp);
 
+// :: DataType -> PatchPath
+export const getConstantNodeType = type =>
```
evgenykochetkov, line 266: I'd add checks for generic types here

packages/xod-project/src/patchPathUtils.js
```diff
@@ -327,6 +332,18 @@ const tweakPathRegExp = new RegExp(
   `^xod/debug/tweak-(${bindableTypes.join('|')})`
 );
 
+// :: DataType -> Maybe PatchPath
+export const getTweakPatchPath = R.ifElse(
```
evgenykochetkov, line 336: and here

packages/xod-project/src/patchPathUtils.js
brusherru: 👌

packages/xod-project/src/patchPathUtils.js
brusherru: 👌

nkrkv (approved):
:heart:

evgenykochetkov (approved):

PR #2071, opened 2020-11-17
Paste expanded nodes to the right position

### LINKED ISSUE #1441: Copy-pasting the expanded `watch` nodes.
If a user copies the expanded `watch` by hitting the keyboard `Ctrl + C`, and `Ctrl +V` combinations, the new expanded `watch` node should appear next to the copied one. By now, the obtained in such way `watch` node appears in the next cell of the patch board ignoring the length of the copied node.

brusherru:
It closes #1441 

Previously, the position for pasted nodes calculates only by pins. This PR takes into account a node width, which might be changed on such nodes as constants, watches, tweaks.


evgenykochetkov (approved):
LGTM

Issue #1468, opened 2018-10-15
IDE crashes when changing arity level of numeric-labeled pins


nkrkv:
[Reported](https://forum.xod.io/t/small-bug-with-variadic-patch-nodes-on-0-24-1/1228) on the forum.

### Environment

- Version: 0.24.1

### Steps to reproduce

1. Create a project
2. Create patch `foo`
3. Place 3 `input-number`'s and 1 `output-number`
4. Give the rightmost two inputs labels `1` and `2`
5. Place a `variadic-2` node
6. Switch to patch `main` and add a `foo` node on it
7. Drag the border to increase the arity level

### Expected behavior

- The node grows
- New pin pairs get labels `3`/`4`, `5`/`6`, etc

### Actual behavior

IDE crashes to the gray screen, no chance to recover

### Console output

```
Uncaught TypeError: undefined does not have a method named "concat" or "fantasy-land/concat"
    at concat (concat.js:53)
    at f2 (_curry2.js:25)
    at converge.js:40
    at f1 (_curry1.js:17)
    at _pipe.js:3
    at _arity.js:5
    at unless (unless.js:29)
    at _curry3.js:26
    at f1 (_curry1.js:17)
    at Object.<anonymous> (pin.js:301)
concat @ concat.js:53
f2 @ _curry2.js:25
(anonymous) @ converge.js:40
f1 @ _curry1.js:17
(anonymous) @ _pipe.js:3
(anonymous) @ _arity.js:5
unless @ unless.js:29
(anonymous) @ _curry3.js:26
f1 @ _curry1.js:17
(anonymous) @ pin.js:301
(anonymous) @ index.js:2491
(anonymous) @ index.js:864
(anonymous) @ patch.js:1406
_map @ _map.js:6
map @ map.js:57
(anonymous) @ _dispatchable.js:39
f2 @ _curry2.js:25
(anonymous) @ patch.js:1397
times @ times.js:37
f2 @ _curry2.js:25
(anonymous) @ patch.js:1395
(anonymous) @ index.js:2491
(anonymous) @ index.js:860
(anonymous) @ _pipe.js:3
(anonymous) @ _pipe.js:3
(anonymous) @ _arity.js:5
(anonymous) @ project.js:361
Just.map @ Maybe.js:67
(anonymous) @ _dispatchable.js:30
(anonymous) @ _curry2.js:20
f1 @ _curry1.js:17
(anonymous) @ _pipe.js:3
(anonymous) @ _arity.js:5
[...]

PR #2074, opened 2020-12-08
Recover IDE from runtime exceptions in renderer process

### LINKED ISSUE #1476: Provide user a possibility to recover from runtime exceptions
Right now, we handle most exceptions in a fairly user-friendly way(ones that may occur during transpilation process and in redux thunks).

But some lead to "crash to the gray screen", not giving a user any error messages or even a chance to save currently opened project(like #1468). Those exceptions are caused by a **combination** of current state and some bug in selectors or rendering code. In that case, we can't just try rendering the same thing again, because that would lead to the same exception.

But instead of rendering *nothing*, we can provide some recovery options
- export current changes to a `.xodball`
- undo last action and try rendering again
- simulate opening project in the current state(that could help if crash was caused by some complex UI manipulations)

brusherru:
It closes #1476 

---
It does not catch any exceptions. However, if some error crashes the GUI, the `<Catcher>` component will catch the exception, recover the state of the IDE to the previous stable state and show the report, which might be shared with the developers on GitHub or XOD forum.

The report includes:
- error message with stack trace,
- component stack,
- last three dispatched actions with the message did the last action broke the IDE or not,
- the stable state, where the tokens (user grant and project API token) are replaced with the string `SECRET`

To test this PR you can check it out and then broke something. For example:
- broke some selectors
- throw an exception in the `render()` method of some component
- reverse commit https://github.com/xodio/xod/commit/f96fa6d2732f8eff6df0bd869b93fbf4b348e43d and reproduce one of the old bugs from the closed issue: https://github.com/xodio/xod/issues/1468

---
This PR also could introduce some bugs in the IDE behavior after being recovered (because of reconstructing `<App>` and its children, so their state will be lost, and they should resubscribe on DOM and IPC events and so on), so I'll be glad if you find something and report it here.
Anyway, I'm sure it's better to get a recovered state and possibility to save the project, then just get a crashed IDE. :)


nkrkv (approved):
:+1: LGTM

packages/xod-client/src/core/containers/Catcher.jsx
```diff
@@ -0,0 +1,188 @@
...
+      // Error
+      '# ERROR',
+      this.state.error.stack,
+      '# COMPONENT STACK',
+      this.state.errorInfo.componentStack,
+      '\n\n',
+      '# LAST ACTIONS (max 3 in log)',
+      isActionErrored
+        ? lastOneCausedError
+        : '# BUT NONE OF THEM IS NOT THE REASON',
```
evgenykochetkov, line 92: Maybe 
```suggestion
        : '# BUT NONE OF THEM IS THE REASON',
```
is more correct?

Or I didn't quite get the point of this message.

packages/xod-client/src/core/containers/Catcher.jsx
```diff
@@ -0,0 +1,188 @@
...
+        className={cls('IdeCrashReport', {
+          'is-shaded': this.state.recovering,
+        })}
+      >
+        {recovering}
+        <div className="Message">
+          <CloseButton onClick={this.onClose} />
+          <h1>Something went wrong</h1>
+          <p>
+            First of all, save the project after recovering process complete.
```
evgenykochetkov, line 137:
```suggestion
            First of all, save the project after recovering process completes.
```

or 

```suggestion
            First of all, save the project after recovering process is complete.
```

evgenykochetkov (approved):
The overall user experience is very cool 👍🏻
I have some doubts about a couple of messages, but I might be wrong.

evgenykochetkov (approved):

PR #2077, opened 2020-12-17
tweak(circleci): pin node version on macos

knopki:
Install NodeJS `v10.13.0` on MacOS CircleCI runner via `nvm`. `v10.13.0` is the version, that used in Docker CircleCI runner and the minimal NodeJS version for #2074 build. Adds 10-15s.


nkrkv (approved):
:+1:

evgenykochetkov (approved):
Nice!

PR #2078, opened 2020-12-20
Copy and save logs from deployment pane

### LINKED ISSUE #2076: As a scientific researcher I want to copy and download serial logs so that I can analyze my sensor data in spreadsheets
When collecting data from sensors for some scientific experiment it is often desirable to get a time-series table for further analysis in Excel or another external tool.

I can use `console-log` and then manually:

- Scroll the output to the top
- Hold down the mouse button to start the selection
- Scroll down to select all
- Press Ctrl+C to copy
- Paste elsewhere

But repeating the process again and again is tiresome. We could simplify researches by providing simple accessible buttons to: 

- Copy the whole output to the clipboard
- Save the whole output to a file specified

image

Generalizing the idea, the same buttons will be quite helpful on all other tabs as well. For example, to ask questions on the forum or make bug reports.

### User story

- I make a patch with `console-log` to gather some data
- I upload with debug and see the data coming to the Serial tab
- I press the “Copy” button and the contents of the Serial tab are copied to the clipboard exactly as I see them, applying the current filter
- I press the “Save to File” button and the system file save dialog appears asking for the filename (defaults to `serial-log.txt`)
- I enter the filename and the file gets saved with exactly the same contents as I see them

### Acceptance criteria

- [x] The basic scenario works as described
- [x] If I choose an existing file XOD asks me if I want to overwrite it
- [x] If I copy a log made with `join` with TAB delimiter and then paste in Excel the columns get split correctly (the leading arrow in first is OK)
- [x] If I do the same with Google Spreadsheets, the result is split too
- [x] If I choose to include watched values to the log and have some, their log entries are copied/saved too
- [x] The buttons work the same way for all tabs of the deployment pane
- [x] Whenever I press the “Copy” button it gives some visual feedback acknowledging success

### Out of scope

- Filtering out arrows, headers, footers

brusherru:
It closes #2076.

### DIFF
--- a/packages/xod-client/src/debugger/containers/Debugger.jsx
+++ b/packages/xod-client/src/debugger/containers/Debugger.jsx
@@ -11,7 +11,15 @@ import { Maybe } from 'ramda-fantasy';
 
 import { LOG_TAB_TYPE } from '../constants';
 import * as selectors from '../selectors';
+import { addError, addConfirmation } from '../../messages/actions';
 import * as DA from '../actions';
+import {
+  LOG_COPIED,
+  LOG_COPY_NOT_SUPPORTED,
+  logCopyError,
+  logSaveError,
+} from '../messages';
+
 import Log from './Log';
 import SerialInput from '../components/SerialInput';
 
@@ -44,6 +52,52 @@ const TAB_ORDER = [
 ];
 
 class Debugger extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.onCopyLogClicked = this.onCopyLogClicked.bind(this);
+    this.onSaveLogClicked = this.onSaveLogClicked.bind(this);
+  }
+
+  onCopyLogClicked() {
+    const copy = R.path(['navigator', 'clipboard', 'writeText'], window);
+    if (!copy) {
+      this.props.actions.addError(LOG_COPY_NOT_SUPPORTED);
+      return;
+    }
+
+    window.navigator.clipboard.writeText(this.props.log).then(
+      () => {
+        this.props.actions.addConfirmation(LOG_COPIED);
+      },
+      err => {
+        this.props.actions.addError(logCopyError(err));
+      }
+    );
+  }
+
+  onSaveLogClicked() {
+    const tabLabel = TAB_NAMES[this.props.currentTab];
+    const defaultFilename = R.compose(R.concat(R.__, '-log.txt'), R.toLower)(
+      tabLabel
+    );
+
+    try {
+      const data = new window.Blob([this.props.log], { type: 'text/plain' });
+      const file = window.URL.createObjectURL(data);
+
+      const link = document.createElement('a');
+      link.download = defaultFilename;
+      link.href = file;
+      link.click();
+
+      // We need to manually revoke the object URL to avoid memory leaks.
+      window.URL.revokeObjectURL(file);
+    } catch (err) {
+      this.props.actions.addError(logSaveError(err));
+    }
+  }
+
   renderControlsForExpandedState() {
     const {
       isExpanded,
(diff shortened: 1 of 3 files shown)


packages/xod-client/src/debugger/messages.js
```diff
@@ -48,3 +48,26 @@ export const DO_NOT_USE_TETHERING_INTERNET_IN_BROWSER = {
   solution:
     'Install the desktop version of the XOD IDE to reveal all features of XOD',
 };
+
+export const logCopyDone = tabLabel => ({
+  title: `${tabLabel} log is copied to the clipboard`,
```
nkrkv, line 53:
```suggestion
  title: `Copied`,
```

packages/xod-client/src/debugger/messages.js
```diff
@@ -48,3 +48,26 @@ export const DO_NOT_USE_TETHERING_INTERNET_IN_BROWSER = {
...
+});
+export const logCopyError = (tabLabel, error) => ({
+  title: `Can not copy the ${tabLabel} log`,
```
nkrkv, line 56:
```suggestion
  title: `Cannot copy log`,
```

packages/xod-client/src/debugger/messages.js
```diff
@@ -48,3 +48,26 @@ export const DO_NOT_USE_TETHERING_INTERNET_IN_BROWSER = {
...
+  note: error.message,
+  solution: 'Please, report the issue to the XOD developers ',
+});
+export const logCopyNotSupported = () => ({
+  title: 'Your browser does not support a copy log feature',
+  solution:
+    'Try to install the desktop IDE or use the latest version of Google Chrome.',
```
nkrkv, line 63:
```suggestion
  title: 'Bad browser',
  title: 'Your browser does not support a copy log feature',
  solution:
    'Install the desktop IDE or use the latest version of Google Chrome / Firefox.',
```

packages/xod-client/src/debugger/messages.js
```diff
@@ -48,3 +48,26 @@ export const DO_NOT_USE_TETHERING_INTERNET_IN_BROWSER = {
...
+});
+
+export const logSaveDone = tabLabel => ({
+  title: `${tabLabel} log is saved on your filesystem`,
+});
```
nkrkv, line 68: :bug: In desktop IDE this message appears before the file save dialog pops up, even if I choose to cancel (in the future). If it’s hard to make the message appear in the right moment (after the dialog), it’s OK to not show it at all.

packages/xod-client/src/debugger/messages.js
```diff
@@ -48,3 +48,26 @@ export const DO_NOT_USE_TETHERING_INTERNET_IN_BROWSER = {
...
+export const logSaveError = (tabLabel, error) => ({
+  title: `Can not save the ${tabLabel} log`,
```
nkrkv, line 70:
```suggestion
  title: `Cannot save log`,
```

nkrkv (changes requested):

packages/xod-client/src/debugger/containers/Debugger.jsx
```diff
@@ -79,11 +138,42 @@ class Debugger extends React.Component {
     );
   }
 
+  renderTabActions() {
+    const isDisabled = R.isEmpty(this.props.log);
+    const title = notDisabledTitle =>
```
evgenykochetkov, line 143: I don't think the title should be altered. The reason for disabling the buttons when the log is empty is obvious enough.

evgenykochetkov:
Besides the save message in the desktop IDE, works absolutely fine in both browser and desktop versions 👍🏻

nkrkv (approved):
:+1:

PR #2081, opened 2020-12-28
Resizable expanded deployment pane

### LINKED ISSUE #2080: As a xoder I want to expand deployment pane so that I can see more output at once
### Rationale

Currently, the height of the deployment pane is fixed. If I want to observe a lot of output at once (e.g, serial log or compiler log), I’m in bad luck. Let’s make the deployment pane height adjustable.

The pane contains only textual output only, so breaking the layout of the inner elements is not an issue.

### Acceptance criteria

- [x] I can change the height of the expanded deployment pane by dragging its upper edge
- [x] When I hover the edge, the cursor changes to the traditional North-South double arrow
- [x] The resize range is limited by some minimum so that I can’t make the height less than it is required to contain the “Follow log” button and the “Send via serial” line
- [ ] The resize range is limited by some maximum so that I can’t make the height greater than it is required to see the tabs line and the breadcrumbs pane
- [ ] The height setting is preserved across collapse/expand
- [ ] The height setting is preserved across IDE relaunches to the same extent as for sidebar layout settings
- [ ] If I see a log that is followed, it is still followed after resize

brusherru:
It closes #2080.


packages/xod-client/src/editor/containers/Workarea.jsx
```diff
@@ -0,0 +1,124 @@
...
+      >
+        <ReflexElement minSize={SLOT_SIZE.HEIGHT}>
+          <div className="Workarea-inner">{this.props.children}</div>
+        </ReflexElement>
+        <ReflexSplitter
+          onStartResize={this.onStartResizePanel}
+          onStopResize={this.onStopResizePanel}
+        />
+        <ReflexElement
+          minSize={this.props.isDebugPanelExpanded ? 220 : 24}
```
nkrkv, line 74: I’d relax the 220 limit to something like ~100

nkrkv (approved):
:+1:

nkrkv:
@evgenykochetkov review?

evgenykochetkov (approved):
The issue with "follow log" is fixed, the issue with positioning of collapsed panel after expanding/collapsing and then resizing window is nearly impossible to reproduce. LGTM.

PR #2090, opened 2021-01-25
Move windows builds from AppVeyor to CircleCI

evgenykochetkov:


.circleci/config.yml
```diff
@@ -148,6 +168,19 @@ jobs:
...
+      name: win/default
+    steps:
+      - run: *step-configure-git-on-windows
+      - checkout
+      - run: *step-install-yarn-on-windows
+      - run:
+          name: Install Node modules
+          command: yarn
+      - run: *step-build
+      - run: *step-verify-git-clean
```
knopki, line 182: Executor image is not pinned and can be updated by third party. It's ok for windows image, but I propose to add NVM step to pin Node version. (NVM is already installed).

.circleci/config.yml
```diff
@@ -233,6 +266,34 @@ jobs:
             - "*.dmg"
             - "*.zip"
 
+  dist-windows:
```
knopki, line 269: How to test this?
Also, nvm step here too.

knopki (changes requested):
Awesome!
Windows host works unexpectedly fast.

nkrkv (approved):
Great!

Shouldn’t it be targeted toward `master` instead of `0.37.x`?

brusherru (approved):

.circleci/config.yml
```diff
@@ -72,6 +72,13 @@ defs:
             git config --global core.autocrlf input
             git config --system core.longpaths true
 
+    step-install-node-on-windows: &step-install-node-on-windows
+      name: Install node
+      command: |
+            nvm install 12.12.0
```
knopki, line 78: But I see `.nvmrc` in the root of repository 🤔

knopki: Btw, `12.12.0` is the last pre-LTS version. LTS starts from `12.13.0` :P

knopki (approved):
LGTM

PR #2093, opened 2021-01-27
Pin node versions in CircleCI runners.

evgenykochetkov:
(finishing what was started while discussing #2090)

Now the exact version of node that is shipped with electron is specified in `.nvmrc` and `engines` field of `package.json`.

In windows and macOS runners, that version is installed with nvm.
For linux runners, Docker image is updated to include the required version.


knopki (approved):
But there is nothing in `README.md` about `nvm` or `--ignore-engines` :)

package.json
```diff
@@ -86,5 +86,8 @@
   },
   "resolutions": {
     "**/ramda-fantasy/ramda": "^0.24.1"
+  },
+  "engines": {
+    "node": "12.16.3"
   }
-}
+}
```
nkrkv, line 93: EOL missing

nkrkv (approved):
:+1:

PR #2095, opened 2021-01-27
Change base of CircleCI docker image

knopki:
Use official CircleCI image as base for our CircleCI docker executor image. The `Dockerfile` just got much simpler and the base image is always in CircleCI cache.
Also, `circleci` user used instead of `root`. So, cache reset is required.

Base images for reference:
https://github.com/CircleCI-Public/cimg-base/blob/04b92395b0d7adf8ac30aa7c7c53f737294025dd/18.04/Dockerfile
https://github.com/CircleCI-Public/cimg-node/blob/308ce216a1d6f6979e4eb426b7dccb80fd023001/12.16/Dockerfile


evgenykochetkov (approved):
Awesome!

brusherru (approved):
LGTM

PR #2098, opened 2021-02-05
Table logs feature: collect, view and save tabular data

brusherru:
There is no issue 🙈 

- Adds new node `xod/debug/table-log`
- Adds new "Table Log" tab with table-view of the collected data

## How to test:
1. Place the new `xod/debug/table-log` node
2. Place `join` node
3. Set `\t` as a delimiter
4. Link some data sources (sensors / `system-time` / `sine-wave` / etc)
5. Link join with `table-log` node
6. Link pulse source to the `PUSH` pin of `table-log`
6. Run the program: simulate or upload with debugging
7. Double click on the `table-log` node to drill-down in the special tab "Table Log"

Also, you can
- place any amount of `table-log` nodes,
- send pulses to the `NEW` to create new sheets,
- `table-log` nodes might be placed inside other nodes,
- change node labels to make source names more readable
- run few different programs from one project and see logs of all runs


brusherru:
Also, I have an idea of how to improve source labels and make it looks like:
- `@/rootPatch`
- `@/rootPatch > unlabeled (nodeId) #1`
- `@/rootPatch > unlabeled (nodeId) #2`
- `@/rootPatch > my-label #2`
- `@/rootPatch > DELETED (nodeId)`

workspace/__lib__/xod/debug/table-log/patch.xodp
```diff
@@ -0,0 +1,46 @@
+{
+  "description": "Shows incoming values in the realtime, when a debug session is active",
```
nkrkv, line 2: This should be updated. I’m thinking about:

> Collects values as table rows and tables as sheets. Useful to conduct measurement sessions and then analyze the data gathered in spreadsheet software.

The pins require docs too.

workspace/__lib__/xod/debug/table-log/patch.xodp
```diff
@@ -0,0 +1,46 @@
...
+      "position": {
+        "units": "slots",
+        "x": 3,
+        "y": 0
+      },
+      "type": "xod/patch-nodes/input-pulse"
+    },
+    {
+      "boundLiterals": {
+        "__out__": "On Boot"
```
nkrkv, line 25: This should be "Never" by default. Otherwise we always start with the second sheet

packages/xod-client/src/debugger/selectors.js
```diff
@@ -128,39 +134,62 @@ export const getCurrentChunksPath = createSelector(
...
+          nodeIdPath,
+          nodeMap
+        );
+
+        const tableLogValues = R.compose(
+          R.map(sheets => {
+            const lastSheetIndex = sheets.length > 0 ? sheets.length - 1 : 0;
+            return `Sheet ${sheets.length} has ${
+              sheets[lastSheetIndex].length
+            } rows`;
```
nkrkv, line 187: I’d suggest more coincise form: “1 sheets, 13 rows”

nkrkv (changes requested):
The source drop-down looks reversed (right to left) on my machine:

image

evgenykochetkov:
Not sure if this is intentional, but "Table Log" tab jumps to the rightmost position when drilling down.
https://user-images.githubusercontent.com/2527093/107238433-e2c1da00-6a38-11eb-91c2-ba1033c8b091.mov

packages/xod-client/src/editor/reducer.js
```diff
@@ -130,34 +131,51 @@ const setPropsToTab = R.curry((id, props, state) =>
...
+const addTableLogTab = R.curry((nodeId, activeSheetIndex, state) =>
+  setPropsToTab(
+    TABLE_LOG_TAB_ID,
+    {
+      id: TABLE_LOG_TAB_ID,
+      index: getTabNewIndex(state),
+      type: TAB_TYPES.TABLE_LOG,
+      nodeId,
+      activeSheetIndex,
+      // offset: { x: 0, y: 0, }, // TODO: Why do we need it?! :-(
```
evgenykochetkov, line 173: 👀

nkrkv (approved):
:fire:

evgenykochetkov (approved):

PR #2107, opened 2021-03-04
Introduce variadic-pass nodes

### LINKED ISSUE #1144: Implement variadic pins pass/forwarding
### Rationale

The current state of variadic node support is not :100: complete. A xoder can’t implement a variadic `average` node or `text-lcd` with an adjustable number of lines. See the [discussion on the forum](https://forum.xod.io/t/getting-data-for-variadic-nodes/542).

In other languages supporting variadic argument functions the very common pattern for creating a fold/reduce based routine is splitting it into private and public parts. The private part *is* the reducer and the only thing public part does is setting the initial accumulator value, call reduce *forwarding the list to it*, then trivially post-process the final accumulator value and return the result.

Currently, we can’t forward the list, i.e., the variadic pins. Let's fix it.

A xoder should be able to place a `variadic-pass-1`, `variadic-pass-2`, or `variadic-pass-3` marker node on his patch. Having done that N rightmost inputs become variadic, however, when transpiling the node is not expanded into a cascade. Instead, the input terminals propagate the arity level set down to the variadic nodes to which they link. The arity level gets *added* to the design-time arity level. Then the links incoming to the outer variadic-pass node are directly transferred/replicated to the inner variadic nodes. And after that’s done, the regular expansion stage takes place.

To be unambiguous and simpler to implement the following restrictions should be applied:

- There can’t be two different `variadic-*` markers on a patch node. The markers `variadic-2` and `variadic-pass-2` are mutually exclusive.
- A variadic pass input terminal must not link to a non-variadic (scalar) pin of an inner node.
- A node with a `variadic-pass-n` marker should have at least `n` inputs.
- All variadic pins of an inner variadic node must be linked to variadic pass input terminals.

The following case is OK regardless it might sound strange:

- A variadic pass input terminal links to nothing

### Acceptance criteria

- [ ] I can place a `variadic-pass-2` node onto a patch node `foo` with two input pins
- [ ] I can place another variadic marker and both will show a wavy-error and transpiler will fail¹
- [ ] Keeping only a single `variadic-pass-2` on `foo`, I can use `foo` as a node and adjust its arity level as it was a regular variadic
- [ ] Inside the `foo` I can link the input terminals to scalar pins and the marker will show a wavy-error and transpiler will fail²
- [ ] I can instead link single `foo` variadic pass input terminal to a `variadic-2` input. The marker will show a wavy-error and transpiler will fail³
- [ ] I can remove an input in `foo` and that will lead to an error⁴
- [ ] `xod/core/average` and `xod/core/arity-level` are implemented

¹
```markdown
**Conflicting variadic markers**
Patch @/main -> @/foo -> @/bar contains exclusive `variadic-pass-2` and `variadic-2`
marker nodes. | No more than a single `variadic-*` node is allowed. |
Delete one to continue.
```

²
```markdown
**Variadic input links to scalar**
Patch @/main -> @/foo -> @/bar contains a `variadic-pass-2` marker node which
| can’t make the input Y variadic as it links to a scalar pin. |
Delete the offending link or marker to continue.
```

³
```markdown
**Variadic linked to scalar**
Patch @/main -> @/foo -> @/bar has a variadic pass link from input `FOO` to the
`some-inner-node`. So, | all variadic inputs of `some-inner-node` should be
linked to variadic pass inputs. |
Pass all variadic pins to the input terminals or delete the offending link to continue.
```

⁴
```markdown
**Too few variadic inputs**
Patch @/main -> @/foo -> @/bar contains a `variadic-pass-2` marker node, so
| the patch node should have at least 2 inputs |.
Add inputs or delete the marker to continue.
```

The vertical bars denote tooltip-version message boundaries.

evgenykochetkov:
Closes #1144


packages/xod-project/src/expandVariadicPassNodes.js
```diff
@@ -0,0 +1,231 @@
...
+            R.compose(
+              expandPassPatch(desiredArityLevel),
+              Project.getPatchByPathUnsafe(patchPath)
+            )(project)
+          ),
+          R.reject(({ patchPath, desiredArityLevel }) =>
+            R.compose(
+              Maybe.isJust,
+              Project.getPatchByPath(
+                getExpandedVariadicPatchPath(desiredArityLevel, patchPath)
```
brusherru, line 189: Is it possible? 🤔

brusherru (approved):

PR #2108, opened 2021-03-10
Add sugar to table log feature

brusherru:
There is no issue.

It includes:
1. A few new utility nodes in `xod/debug` and restyled `table-log`, which is a variadic-pass (no more `join` on the user-side).
2. Tweaks in `xod-arduino` and `xod-client` to keep the behavior of nodes and table log tab as it was before: no excessive labels in the sources dropdown, show collected data on the `table-log`, open correct source by double-clicking on the `table-log` node.

### DIFF
--- a/packages/xod-client/src/debugger/selectors.js
+++ b/packages/xod-client/src/debugger/selectors.js
@@ -11,7 +11,7 @@ import { DEBUGGER_TAB_ID } from '../editor/constants';
 import { SESSION_TYPE } from './constants';
 
 import { createMemoizedSelector } from '../utils/selectorTools';
-import { getTableLogSourceLabels } from './utils';
+import { getTableLogSourceLabels, removeLastNodeIdInChain } from './utils';
 
 export const getDebuggerState = R.prop('debugger');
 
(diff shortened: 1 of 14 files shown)


workspace/__lib__/xod/debug/tsv-log/patch.cpp
```diff
@@ -0,0 +1,30 @@
+node {
+    void printPrefix(Context ctx) {
+        TimeMs tNow = transactionTime();
+        XOD_DEBUG_SERIAL.print(F("+XOD:"));
+        XOD_DEBUG_SERIAL.print(tNow);
+        XOD_DEBUG_SERIAL.print(':');
+        XOD_DEBUG_SERIAL.print(getNodeId(ctx));
+        XOD_DEBUG_SERIAL.print(':');
+    }
+    void printEOL() {
```
evgenykochetkov, line 10: I think this could be moved to `nodespace`

packages/xod-client/src/debugger/selectors.js
```diff
@@ -186,7 +186,16 @@ export const getInteractiveNodeValuesForCurrentPatch = createSelector(
               sheets[lastSheetIndex].length
             }`;
           }),
-          filterOutValuesForCurrentPatch(activeIndex, nodeIdPath)
+          filterOutValuesForCurrentPatch(activeIndex, nodeIdPath),
+          // `tsv-log`, which is actually stores the data, encapsulated
+          // inside `table-log`, which label should be replaced with
+          // this value. So we have to cut the latest nodeId part from
+          // keys of table log values map
```
evgenykochetkov, line 193: It could be done with something like `R.replace(/(~[^~]+)$/, '')`
But your version is probably readable 😀

evgenykochetkov (approved):
LGTM (with a couple very minor remarks)

PR #2110, opened 2021-03-17
Fix calling of GPIO functions only from global namespace

brusherru:
There is no issue.

Calling `digitalRead,` digitalWrite`, or `analogRead` functions only from the global namespace may not work on some architectures, like SAMD. So there is no need to call these functions exactly from the global namespace, much better to allow the compiler to find the function that matches the signature.


nkrkv (approved):
:+1:

